PR #7, opened 2014-10-02
_libc.py: add libc.dylib (osx dynamic library) to list passed to ffi.open()

blmorris:
I also grep'd through micropython-lib for other instances of ```ffi.open()``` and found 3 instances:
```
./_libc/_libc.py:            _h = ffi.open(n)
./re-pcre/re.py:pcre = ffi.open("libpcre.so.3")
./sqlite3/sqlite3.py:sq3 = ffi.open("libsqlite3.so.0")
```
My mac (OSX 10.7.5) has these libraries available in /usr/lib:
```
/usr/lib/libpcre.0.dylib
/usr/lib/libpcre.dylib
/usr/lib/libpcreposix.0.dylib
/usr/lib/libpcreposix.dylib
/usr/lib/libsqlite3.0.dylib
/usr/lib/libsqlite3.dylib
```
If you would like I could patch ```re.py``` and ```sqlite3.py``` so ffi can also open these libraries. Of course these files would also need a list iterating mechanism like that used in ```_libc.py``` or ```micropython/tests/unix/ffi_float.py```
I also don't know if I should include all of the *.dylib options or just one of each.

### DIFF
--- a/_libc/_libc.py
+++ b/_libc/_libc.py
@@ -4,7 +4,7 @@ import sys
 
 _h = None
 
-names = ('libc.so', 'libc.so.0', 'libc.so.6')
+names = ('libc.so', 'libc.so.0', 'libc.so.6', 'libc.dylib')
 
 def get():
     global _h
--- a/re-pcre/re.py
+++ b/re-pcre/re.py
@@ -1,8 +1,19 @@
 import ffi
 import array
 
+# pcre = ffi.open("libpcre.so.3")
 
-pcre = ffi.open("libpcre.so.3")
+def ffi_open(names):
+    err = None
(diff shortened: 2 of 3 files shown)


pfalcon:
Ok, since this patch, we addressed this issue, switching from _libc to ffilib module which should abstract away things like dynamic lib suffixes. @blmorris : Would you want to check where we stand with this issue now, or should we just close this issue for now?

blmorris:
@pfalcon - Yes, closing this since it is outdated; ffilib module seems to have resolved issues with dynamic lib suffixes on OSX. I will open another issue if it comes up again.

PR #21, opened 2015-02-08
binascii.unhexlify, base64, hashlib.sha224/256/384/512 

slush0:
My first contribution to micropython-lib.
SHA-X ported from PyPy
hmac ported from CPython 3.3
base64 ported from CPython 3.3
unhexlify is custom as PyPy's version was even slower.

### DIFF
--- a/base64/base64.py
+++ b/base64/base64.py
@@ -1,5 +1,339 @@
-def encodebytes():
-    raise NotImplementedError
+"""RFC 3548: Base16, Base32, Base64 Data Encodings"""
 
-def b64encode():
-    raise NotImplementedError
+# Modified 04-Oct-1995 by Jack Jansen to use binascii module
+# Modified 30-Dec-2003 by Barry Warsaw to add full RFC 3548 support
+# Modified 22-May-2007 by Guido van Rossum to use bytes everywhere
+# Modified 08-Feb-2014 by Marek Palatinus to run in Micropython
+import struct
+import binascii
+
+
+__all__ = [
+    # Legacy interface exports traditional RFC 1521 Base64 encodings
+    'encode', 'decode', 'encodebytes', 'decodebytes',
+    # Generalized interface for other encodings
+    'b64encode', 'b64decode', 'b32encode', 'b32decode',
+    'b16encode', 'b16decode',
+    # Standard Base64 encoding
+    'standard_b64encode', 'standard_b64decode',
+    # Some common Base64 alternatives.  As referenced by RFC 3458, see thread
+    # starting at:
+    #
+    # http://zgp.org/pipermail/p2p-hackers/2001-September/000316.html
+    'urlsafe_b64encode', 'urlsafe_b64decode',
+    ]
+
+
+bytes_types = (bytes, bytearray)  # Types acceptable as binary data
+
+def _bytes_from_decode_data(s):
+    if isinstance(s, str):
+        try:
+            return s.encode('ascii')
+        except:
+            raise ValueError('string argument should contain only ASCII characters')
+    elif isinstance(s, bytes_types):
+        return s
+    else:
+        raise TypeError("argument should be bytes or ASCII string, not %s" % s.__class__.__name__)
+
+# Base64 encoding/decoding uses binascii
+
+def b64encode(s, altchars=None):
+    """Encode a byte string using Base64.
+
+    s is the byte string to encode.  Optional altchars must be a byte
+    string of length 2 which specifies an alternative alphabet for the
+    '+' and '/' characters.  This allows an application to
+    e.g. generate url or filesystem safe Base64 strings.
+
+    The encoded byte string is returned.
+    """
+    if altchars is not None:
+        raise NotImplementedError("Altchars not implemented")
+    
+    if not isinstance(s, bytes_types):
+        raise TypeError("expected bytes, not %s" % s.__class__.__name__)
+
+    # Strip off the trailing newline
+    encoded = binascii.b2a_base64(s)[:-1]
+    return encoded
+
+
+def b64decode(s, altchars=None, validate=False):
+    """Decode a Base64 encoded byte string.
+
+    s is the byte string to decode.  Optional altchars must be a
+    string of length 2 which specifies the alternative alphabet used
+    instead of the '+' and '/' characters.
+
+    The decoded string is returned.  An Exception is raised if s is
+    incorrectly padded.
+
+    If validate is False (the default), non-base64-alphabet characters are
+    discarded prior to the padding check.  If validate is True,
+    non-base64-alphabet characters in the input result in an Exception.
+    """
+
+    if altchars is not None:
+        raise NotImplementedError("Altchars not implemented")
+   
+    if validate:
+        raise NotImplementedError("validate not implemented")
+ 
+    s = _bytes_from_decode_data(s)
+    return binascii.a2b_base64(s)
+
+
+def standard_b64encode(s):
+    """Encode a byte string using the standard Base64 alphabet.
+
+    s is the byte string to encode.  The encoded byte string is returned.
+    """
+    return b64encode(s)
+
+def standard_b64decode(s):
+    """Decode a byte string encoded with the standard Base64 alphabet.
+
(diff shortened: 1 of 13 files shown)


pfalcon:
Thanks for the contribution, much appreciated! However, there're few issues with the patches as provided. I'm currently travelling, so can't provide complete feedback at once, so going to add quick comment by comment, with the intention to give quick feedback rather than wait week(s).

So, issue no.1: general composition of patches (mostly commit messages) - it doesn't follow existing commits. I may imagine that it may be not immediately obvious which patterns are used, so I started this page to collect explicit requirements (WIP): https://github.com/micropython/micropython-lib/wiki/ContributorGuidelines

pfalcon:
2. uPy already has (C) implementation of sha256: http://docs.micropython.org/en/latest/library/uhashlib.html

pfalcon:
3) You mix patches for different modules in one pull request, which complicates review. (One PR for one module is good rule of thumb).

pfalcon:
4) You appear to miss adding new files (or see 3) for things which complicate review).

pfalcon:
5) having more hash algos is definitely good, but I'd like to hear your reasons for why it's useful ;-). The issue is that a specific application rarely needs to use more than 1. Then, you definitely have right direction of putting each algo in a separate module (sha512, sha224, etc.), but I don't see your taking special measures to benefit from that - hashlib module will import them all, and take a lot of memory.

pfalcon:
6) Implementing stuff in Python is nice idea, which I myself advocate. But at least for some functionality (e.g. base64) it's clear that it's important to have it implemented in C as micropython extmod. Do you really want to have Python implementation which is to be dismissed by C implementation soon (or not so soon, depends on when somebody contributes it)?

slush0:
ad 1 + 3 + 4) I reworked commits, please review now.

ad 2) I know there's sha256 in uhashlib, but sha224 need pure implementation of sha256 anyway, so I ported whole module and it's developer choice if he'll use uhashlib or full hashlib.

ad 5) I never worked with memory-restricted python before, so I did not realize this is a problem. shaX are completely separate modules now, so "from hashlib.sha256 import sha256" will work. Unfortunately it breaks CPython compatibility as "from hashlib import sha256" won't work.

ad 6) "Soon" (tm) :-) It is okay to replace pure-python implementation by native C in the future. For my project I need any implementation, and pure-python libraries are the way how I can contribute to Micropython project right now. Anyway, pure-python libraries are IMO more flexible for future development.

pfalcon:
Thanks for the refactors, it looks much better now. But there're still not clear enough things ;-).

So, let's go module by module, starting with hmac.

Why did you change it to accept bytearrays in addition to bytes?

pfalcon:
And can you please confirm the exact Python version you took it from. I used 3.3.3 as a source, so that's what I look at, and I for example don't see "name" property there.

Just to clarify: I very much appreciate your contribution, without people contributing the project won't progress much. But it's a peculiar project, where most code comes from upstream projects, and it must be done very carefully, or otherwise it will turn into unmaintainable mess. For example, consider that you took module from 3.3.1. Then someone wants to upgrade to 3.3.3, or to 3.4. It's one thing to port 5 changed lines, and another - half of the file. The first is relatively easy to do, if those 5 lines are very visible (uPy changes go in a separate commit), but changing half of lines here and there is big trouble. So, the idea is to apply porting effort and scrutiny on the micropython-lib input, once per module, to get as trouble-free maintenance as possible later. Otherwise, you can just assume that any your work will be overwritten later, because nobody will re-do work (number of upgrades) x (number of modules) times. So, that's it, these points are now added to https://github.com/micropython/micropython-lib/wiki/ContributorGuidelines also.

pfalcon:
Ok, I figured it out: contrary to commit message saying "hmac: Ported from CPython 3.3", it was actually taken from 3.4.2. And I see that it's not the bets module to start from, as it depends on hashlib, but that was the only module which had its source specified in commit message.

pfalcon:
> ad 5) I never worked with memory-restricted python before, so I did not realize this is a problem. shaX are completely separate modules now, so "from hashlib.sha256 import sha256" will work. Unfortunately it breaks CPython compatibility as "from hashlib import sha256" won't work.

You're right, sorry for sending you the wrong way :-(. Compatibility is apparently more important than efficiency here (for efficiency, we have uhashlib). But the problem comes up not the first time, I wonder if we need to implement descriptors in uPy to tackle it, after all.

pfalcon:
Ok, I reworked haslib and hmac per ideas above (so changes to upstream were minimized and clearly visible), and pushed to master. Please have a look how it was done. I'll look at base64 stuff as time permits.

pfalcon:
Ok, so "binascii: Added custom unhexlify implementation" contains not only unhexlify() implementation, but also base64 functions, then another implementation of base64 is coming with base64 module. I assume, implementation in binascii is superfluous?

pfalcon:
Aha, right, base64 module implements only base32 itself, and imports the rest from binascii.

slush0:
Dependency of base64 to binascii just copy the dependency model from CPython stdlib.

pfalcon:
Cleaned up and pushed binascii module.

slush0:
I just commited faster unhexlify. Can you pick this commit or should I do separate PR just for that? I think its my last commit to that area.

slush0:
Anyway, instead of using "from sha256 import *" in hashlib/hashlib/__init__.py, this would be better:

```
from .sha256 import sha224, sha256
from .sha512 import sha384, sha512
```

if I'm not mistaken, this avoids importing of supporting memory structures and significantly reduces memory footprint of the library until such methods are invoked.

pfalcon:
> I just commited faster unhexlify. Can you pick this commit or should I do separate PR just for that? I think its my last commit to that area.

Cherry-picked it. (Please use git pull --rebase to pull from master.)

pfalcon:
Of these, base64 turned to be the hardest. Not only because I had to revert a lot of your removals, but be cause it really uncovers couple of non-implemented features in uPy. I thus don't feel comfortable to merge it master, instead pushed my cleanups to https://github.com/micropython/micropython-lib/tree/base64-cpy so far.

pfalcon:
> Anyway, instead of using from sha256 import * in hashlib/hashlib/__init__.py, this would be better:

Fixed.

> this avoids importing of supporting memory structures and significantly reduces

A module is imported on its own completely in either case. What varies is how much of imported module's symbols get copied to current module's namespace.

pfalcon:
> I thus don't feel comfortable to merge it master, instead pushed my cleanups to https://github.com/micropython/micropython-lib/tree/base64-cpy so far.

To clarifu, one thing about I wasn't comfortable is patching array slice assignments. Cases like this should be good motivation to implement such missing features. So, I implemented it in https://github.com/micropython/micropython/commit/cefcbb22b2befac1baa9953c477fcee4031635d3 , but we're really arrived at case of diminishing returns - remaining missing features are so infrequently used, that relatively significant code size increase they bring is hard to justify - at least for all ports. So, I enabled it for unix port of course, but chickened out to do that for stmhal, etc. So, @slush0, can you confirm whether you wanted this code to be run on unix or stmhal? In the latter case, your working around array slice assignment is probably makes sense.

slush0:
Yes, I'm targeting to stmhal arch.

Anyway, is there any point in implementing some feature only for one architecture? I mean - code tested on unix should be running on stmhal without changes...

pfalcon:
Not implementing for one architecture, but selectively enabled/disabled per port. @dpgeorge: Do you have opinion on enabling patch linked above for stmhal?

dpgeorge:
I've wanted this feature for some time in stmhal but not got round to implementing it, so thanks!  Basically it allows you to "memmove/memcpy" which is very useful when trying to get away from any heap allocation.  I'll enable it in stmhal.

pfalcon:
Thanks @dpgeorge for confirming that array slice assignment is useful. But I decided to leave @slush0's patches to avoid it in base64 module just in case it ever will be used on a port with slice assignment disabled.

So, all modules are merged now, thanks @slush0! But please make future pull requests module by module ;-). Btw, feel free to share how porting elliptic curve crypto goes - probably, on the forum.

PR #26, opened 2015-06-03
uasyncio: Add StreamWriter.get_extra_info() to get peername.

dpgeorge:
This is the standard way of getting the name of the client that is
connected.

### DIFF
--- a/uasyncio/uasyncio/__init__.py
+++ b/uasyncio/uasyncio/__init__.py
@@ -95,8 +95,9 @@ class StreamReader:
 
 class StreamWriter:
 
-    def __init__(self, s):
+    def __init__(self, s, extra):
         self.s = s
+        self.extra = extra
 
     def awrite(self, buf):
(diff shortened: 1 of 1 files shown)


dpgeorge:
I was playing with picoweb and wanted to know the IP of the client.  With this patch you can now use the response object as such: `resp.get_extra_info('peername')`.

Note: in flask you do `request.remote_addr` but as far as I can tell that's just a nicety for what is proposed here, and anyway requires get_extra_info.

pfalcon:
Here comes the bloat ;-).

Please merge.

dpgeorge:
It's as minimal as I could make it while still being compatible with CPython API.  I think it's basic functionality to know who the connecting client is (eg for logging, white/black listing).

Alternative would be to support socket.getpeername(), then you could do `resp.s.getpeername()` (but it means you expose the socket object 's' in StreamWriter).

Issue #25, opened 2015-06-03
libpcre.so.3 is not on all systems


dpgeorge:
On my Arch Linux machine `import re` fails because libpcre.so.3 does not exist.  Instead it's libpcre.so.1 (also aliased as libpcre.so).

So should we just change it to load libpcre.so, or try different versions in succession?

dhylands:
Normally, you would just load libpcre.so and it would be symlinked to the appropriate version.

I've also seen stuff where x.so -> x.so.7 -> x.so.7.4 -> x.so.7.4.3

And most code would be symlinked to x.so and would only include the version if there was a very good reason to.

pfalcon:
> Normally, you would just load libpcre.so and it would be symlinked to the appropriate version.

No, that's not how it works. "3" in libpcre.so.3 is a "major version". It's deemed that if major version changes, the lib grew "completely incompatible" with older (or future) lib version. Apps usually link to such major-specified lib versions. Now pure "libpcre.so" is there for naive ld -lpcre to work (how it comes from user specifying -lpcre but "libpcre.so.3" being recorded in the final executable is under veil of darkness, consider that linker magic). In Debian and derivatives, "libpcre.so" is part of *-dev packages. So yes, the easiest way is to load "libpcre.so", except you can't rely it's there.

pfalcon:
So, we have a method to try few lib names in turn already (for libc), can use it here too.

@dpgeorge : What actual pcre version your system has? Here it's:

libpcre3:amd64                                              1:8.31-2ubuntu2

pfalcon:
> Apps usually link to such major-specified lib versions. 

e.g.:

~~~~
$ ldd `which mc`
[]
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc325fd7000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fc325dd3000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fc325acd000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fc3258b3000)
	libgcrypt.so.11 => /lib/x86_64-linux-gnu/libgcrypt.so.11 (0x00007fc325633000)
	libpcre.so.3 => /lib/x86_64-linux-gnu/libpcre.so.3 (0x00007fc3253f5000)
~~~~

dpgeorge:
[...]

PR #27, opened 2015-06-04
ffilib: Initial version of wrapper for ffi module.

dpgeorge:
This library allows to search for an module to load with ffi.  See issue #25 for background.

There is a bit of code duplication here but I couldn't find a way to reduce it.

I think I'd rather just allow a list of libs to be passed in and searched.  It's much simpler.

### DIFF
--- a/ffilib/ffilib.py
+++ b/ffilib/ffilib.py
@@ -1,3 +1,4 @@
+import sys
 import ffi
 
 _cache = {}
@@ -7,22 +8,18 @@ def open(name, maxver=10, extra=()):
         return _cache[name]
     except KeyError:
         pass
+    def libs():
+        if sys.platform == "linux":
+            yield '%s.so' % name
+            for i in range(maxver, -1, -1):
+                yield '%s.so.%u' % (name, i)
+        else:
+            for ext in ('dylib', 'dll'):
+                yield '%s.%s' % (name, ext)
+        for n in extra:
+            yield n
     err = None
-    for n in ("%s.so" % name, "%s.dylib" % name):
-        try:
-            l = ffi.open(n)
-            _cache[name] = l
-            return l
-        except OSError as e:
-            err = e
-    for i in range(maxver):
-        try:
-            l = ffi.open("%s.so.%u" % (name, i))
-            _cache[name] = l
-            return l
-        except OSError as e:
-            err = e
-    for n in extra:
+    for n in libs():
         try:
             l = ffi.open(n)
             _cache[name] = l
--- a/ffilib/ffilib.py
+++ b/ffilib/ffilib.py
@@ -1,0 +1,32 @@
+import ffi
+
+_cache = {}
+
+def open(name, maxver=10, extra=()):
+    try:
+        return _cache[name]
+    except KeyError:
+        pass
+    err = None
+    for n in ("%s.so" % name, "%s.dylib" % name):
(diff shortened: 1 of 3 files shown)


pfalcon:
You didn't run make_metadata.py script on this, did you? *Its* purpose is to generate setup.py from metatdata.txt.

pfalcon:
> There is a bit of code duplication here but I couldn't find a way to reduce it.

There's of course a way, possibly taking more heap space. But how it is should be good enough for v1.

pfalcon:
> I think I'd rather just allow a list of libs to be passed in and searched. It's much simpler.

Patching _libc.py with few lines was much simpler, but you decided to go a longer way ;-). There're various ways to tweak it, but as I said, should be good enough for v1. Possible tweaks: make "extra" a "wellknown", and start with them (that's to address your concern, I'm not sure it wouldn't be changed later again). Start with maxversion and down-count to 1. And well, you didn't even consider that on Windows, libpcre.so would likely be pcre.dll ;-)

pfalcon:
So, again, this is good enough to go, once metadata is fixed. Let's make sure we know the next steps. And that would be adding a convenience method to load libc - as I mentioned, this new module is to replace _libc.py. So, suppose we add "libc()" function for that, which saves other modules from repeating params, us patching them in multiple places, and actually, it's quite adhoc matter (guess how libc is called under Windows).

What's next? On normal unix, socket API is part of normal API, but on windows, it's a separate DLL. So, need another convenience function in this module, which will return the same libc on unix, but adhoc lib on Windows. The idea is clear I hope. Thinking about naming convention for such functions beforehand may help then.

dpgeorge:
> You didn't run make_metadata.py script on this, did you?

No, I just copied over from _libc/*.  I guess the script should be called make_setup.py :)

> There's of course a way, possibly taking more heap space.

Yeah, I didn't want to make an explicit list of all libs to try.  But I just realised there is a way using generators:
```
def name_list():
    for ext in ('so', 'dylib', 'dll'):
        yield '%s.%s' % (name, ext)
    for i in range(maxver):
        yield '%s.so.%u' % (name, i)
    for n in extra:
        yield n
```
And of course you can use sys.platform to see if you're on linux and need to check the .so's.  Generators are a bit pricey, but I think I like the above better than the way it currently is; what do you think?

> Possible tweaks: make "extra" a "wellknown", and start with them

"extra" was supposed to be for corner cases that didn't fit the standard scheme.

> Start with maxversion and down-count to 1.

That's probably better.

> this new module is to replace _libc.py.

So does in need "bitness"?

dpgeorge:
Updated:
- now using yield pattern to iterate filenames
- checks for linux platform
- setup.py generated by make_metadata.py

pfalcon:
> Generators are a bit pricey, but I think I like the above better than the way it currently is; what do you think?

Yes, I meant a generator. But as I told above, even how it is is good enough for starters.

> So does in need "bitness"?

Ah, so there's bitness... You see, _libc.py was already used as a catch-all. Currently, bitness is used only by select.py, but between moving it there and to ffilib, I think the latter is better. (Again, that can go as next action(s), to be done on demand).

dpgeorge:
Merged in 6dae7a234e92d40a2479fd7e386cf5372ca391f7.

pfalcon:
Pushed to pypi.

PR #28, opened 2015-06-05
argparse: Initial minimal implementation of argparse.

dpgeorge:
CPython's argparse needs around 160k heap just for the module, and has
heaps of dependencies.  This minimal version can run in 16k heap and has
only sys and namedtuple dependencies (which are builtin to MicroPython).

### DIFF
--- a/argparse/argparse.py
+++ b/argparse/argparse.py
@@ -18,39 +18,39 @@ class _Arg:
         self.default = default
         self.help = help
 
-    def parse(self, argv):
+    def parse(self, args):
         # parse args for this arg
         if self.action == "store":
             if self.nargs == None:
-                if argv:
-                    return argv.pop(0)
+                if args:
+                    return args.pop(0)
                 else:
                     raise _ArgError("expecting value for %s" % self.name)
             elif self.nargs == "?":
-                if argv:
-                    return argv.pop(0)
+                if args:
+                    return args.pop(0)
                 else:
                     return self.default
             else:
                 if self.nargs == "*":
                     n = -1
                 elif self.nargs == "+":
-                    if not argv:
+                    if not args:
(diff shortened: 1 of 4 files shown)


pfalcon:
Looks good, though ideally, there would be example_*.py file showing what it can do (and as a manual smoke test).

Note that we should make some policy how to go about extending such "minimal, but functional" implementations. E.g. I already have https://github.com/micropython/micropython-lib/tree/master/utarfile , changes like https://github.com/micropython/micropython-lib/commit/2c2f6104f48830c7284ac083ab7e4b81007f493d make me wonder if I should split older implementation as ulogging, etc.

dpgeorge:
> ideally, there would be example_*.py file

test_argparse.py now added (also serves as an example).

> Note that we should make some policy how to go about extending such "minimal, but functional" implementations.

Yes, I flipped between calling this uargparse and argparse.  If the aim is to be able to run existing Python programs unchanged then we need to have something called argparse.  This could be an extension of uargparse (to start with just from uargparse import *).

The main thing that CPython's argparse has that this one doesn't is pretty printing of the usage, including line wrapping and locale behaviour.  They add lots of bloat.

pfalcon:
>  This could be an extension of uargparse (to start with just from uargparse import *).

Yes. Other approach of course to just try to port CPython module. Anyway, those are thoughts for future.

Looks good, please merge!

dpgeorge:
> Other approach of course to just try to port CPython module.

I did manage to get CPython's argparse.py to run in uPy with minimal changes, but it took 160k heap, and thought that a micro-version would be a useful thing to have.

dpgeorge:
Merged in eba0c0ad8dd4eaf747da9d6de9ebe116a3e1d4fa.

PR #29, opened 2015-06-14
BytesIO

cloudformdesign:
add `BytesIO` object to the `io` module. Also pave the way for future buffered objects.

Note that I removed the first line in `io.py` that did something like `from _io import *`. I don't know what this was supposed to do.

Also I was thinking a description like "reduced stdlib io module from the micropython" would be appropriate with no long_description

### DIFF
--- a/io/io.py
+++ b/io/io.py
@@ -11,6 +11,8 @@ class BufferedIOBase:
     def write(self, data):
         self.flush()
         self._data[self._location: self._location + len(data)] = data
+        self._location += len(data)
+        return len(data)
 
     def seek(self, location):
         if location < 0:
@@ -19,8 +21,7 @@ class BufferedIOBase:
         if location >= len(self._data):
             raise ValueError("location {} outside of length".format(location))
         self._location = location
-
-        self._location += len(data)
+        return location
 
     def tell(self):
         return self._location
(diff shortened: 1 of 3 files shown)


pfalcon:
You must be kidding me, right? What word in the phrase:

> You'd take your favorite Python project and would work towards making it run on uPy (not by preparing long patches - by discussing what's missing to make it run, and how to approach it!)

(specifically in the part in parentheses) perplexed you so that you went to do this? Don't you think that I wrote it not of love of preaching, but because it was very easy on my side to predict you doing just that? (Even though we spent few mails discussing just that issues in slightly different context!)

> from _io import *. I don't know what this was supposed to do.

But don't you know from all your life experience that there're on 2 reliable ways to deal with this situation: 1) "don't touch it"; 2) get to learn and understand it. Funny thing that it's 2nd time in last 24hr I'm pointing that out wrt to uPy submissions (but the other guy went for choice 2 long before I made that comment).

Also, the phrase like above means that you don't even know Python, and despite it being very easily to learn, you don't even bother.

~~~~
$ micropython 
>>> import _io
>>> dir(_io)
['__name__', 'open', 'FileIO', 'TextIOWrapper', 'StringIO', 'BytesIO']
~~~~

This patch is a waste and rejected.

PR #31, opened 2015-06-20
select: add EPOLLERR, EPOLLHUP, EPOLLRDHUP constants

dxxb:


pfalcon:
What's the usecase for these?

dxxb:
epoll_wait() sets these bits when the relevant condition occurs (no need to set them when doing an EPOLL_CTL_ADD). I have used EPOLLHUP to distinguish between a closed pipe event and an EPOLLIN condition. I thought it would be useful to provide the other two constants for completeness and I am also guessing I may use EPOLLERR soon.

pfalcon:
Thanks, I just wanted to make sure these come from a real usecase vs just speculation that they're missing. Merged.

PR #32, opened 2015-06-20
multiprocessing: improve cpython compatibility

dxxb:
According to the docs Connection:recv() must raise EOFError on EOF


pfalcon:
Thanks, merged. But I reworded commit message a bit.

PR #33, opened 2015-06-20
logging: log to sys.stderr instead of sys.stdout

dxxb:
I am not sure if this will fly and how it affects non-unix ports but if we have to pick logging to stdout or stderr I would pick stderr.

### DIFF
--- a/logging/logging.py
+++ b/logging/logging.py
@@ -4,6 +4,8 @@
 # Basic, useful module, by CPython impl depends on module "string" which
 # uses metaclasses.
 
+import sys
+
 CRITICAL = 50
 ERROR    = 40
 WARNING  = 30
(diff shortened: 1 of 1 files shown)


pfalcon:
Well, one risk is indeed non-unix ports. But otherwise, we should do it like upstream ;-). Providing direct references to docs will improve turnaround time.

dpgeorge:
sys.stderr is relatively common in ports, so ok to use it like this.

CPython has concept of "handlers" and it seems the default one does indeed go to stderr: 
https://docs.python.org/3.4/library/logging.html#logging.basicConfig
https://docs.python.org/3.4/library/logging.handlers.html#logging.StreamHandler

dxxb:
CPython defaults to sys.stderr when no destination is selected. micropython-lib's logging module does not support selecting a configuration right now so I guess it makes sense to default to sys.stderr. See [direct reference to docs](https://docs.python.org/3/howto/logging.html#what-happens-if-no-configuration-is-provided) ;-)

pfalcon:
Thanks for references, guys. Merged.

@dxxb : Good start with the patches, feel free to submit/raise for discussion any other points you see. Thanks.

Issue #10, opened 2014-10-30
Makefile depends on GNU core utilities ('make install' fails on OSX)


blmorris:
The reason appears to be that the GNU and BSD versions of 'find' require different options:

```
bryans-computer:micropython-lib bryan$ make install
find: illegal option -- m
find: illegal option -- a
find: illegal option -- e
find: illegal option -- p
find: illegal option -- t
find: illegal option -- h
find: 1: No such file or directory
```

I haven't yet figured out how to modify the command to work on OSX, will continue to look into it.
Reference for BSD (OSX) find (the man page on OSX is basically the same):
https://www.freebsd.org/cgi/man.cgi?query=find(1)&sektion=
Reference for GNU find:
http://linux.die.net/man/1/find

jonnor:
No motion over the last 9 years. I am pretty sure that Mac OSX builds have worked for many years already.

PR #36, opened 2015-07-18
Install by python script

pramasoul:
Implement <code>make install</code> with a python script per #10. Tested on Ubuntu 14.04.2 LTS and OSX 10.9.5.


pfalcon:
> in python3 for uniform cross-platform behavior

Unfortunately, using python3 isn't *that* cross-platform. For example, we recently had a case of OpenWRT build failure, because python3 isn't among default host tools there.

@roger- : What would you say? ^^^

pfalcon:
Overall, the script looks over-engineered for me.

~~~~
from distutils.dir_util import mkpath, copy_tree
~~~~

brings in avoidable dependencies. General rule of thumb I tried to follow is to write self-hosted tools (i.e. Python scripts which can be run with uPy itself). That single line breaks it. argparse dependency is not so breaking, but still not unavoidably necessary.

On a quick look, the script itself is also longer than I'd otherwise imagine.

pfalcon:
Summing up: thanks, but I don't intend to merge this immediately. I'd rather wait couple of weeks in case:

1. You might rework it based on comments above.
2. Someone else might provide alternative implementation along the lines above.

roger-:
In this case, yes, Python 3 isn't available but maybe it'll be good to have this as an option.

pfalcon:
@pramasoul : So, any responses to comments above?

pramasoul:
- Verified that <code>install_modules.py</code> works in python2.7 as well as python3. It now uses the system default python.
- "Down-engineered" the coding style

What approach do you recommend for accomplishing the file tree copies, in place of <code>distutils.dir_util.copy_tree</code>?

pfalcon:
> What approach do you recommend for accomplishing the file tree copies, in place of distutils.dir_util.copy_tree?

There's os.walk() and glob already available, and if you really feel like that, you can implement shutil.copytree().

Thanks!

pfalcon:
Ping - this patch's branch now has unrelated changes and needs a rebase or something.

pfalcon:
Ping again.

pramasoul:
Roger that. In the queue.

pfalcon:
Thanks.

puhitaku:
This PR is inactive for more than 1 year, and I've found that a Makefile in `master` branch is inconvenient for my macOS ( #10 ) so I want this to be merged ASAP.

My idea is that you can do merge-with-squash even if this branch has unrelated changes.

`make install` still works with both Python 2 / 3 flawlessly. I'm really into it. Thanks.

puhitaku:
Ping @pfalcon

jimmo:
Closing due to inactivity and a significant restructure of the way package installation works in micropython-lib.

PR #38, opened 2015-07-24
unittest: make unittest.main() work

pramasoul:
Enables standalone test scripts to use the [common idiom](https://docs.python.org/3.4/library/unittest.html#basic-example) in unit test scripts:
```python
if __name__ == '__main__':
    unittest.main()
```


pfalcon:
Merged, thanks.

Issue #37, opened 2015-07-20
XML tokenizer


pfalcon:
Lately, I was looking into parsing XML. Doing my homework, I looked into bunch of small XML parsers, but they all were still too bloated and insisted on well-formedness checking, and had other funky things.

I was looking into really minimal way to parse/extract useful parts from XML, and that's just working with stream of tokens. So, I wrote a simple tokenizer in Python. There's no XML tokenizer module in CPython standard lib, but I guess it would be nice addition to micropython-lib, as other XML APIs are unlikly coming soon.

Thoughts?

I propose "xmltok" for module name.

dpgeorge:
What are existing (generic) tokeniser options in CPy?  A generic tokeniser might be more useful (but also more difficult to design so copying existing might be a way to go).

pfalcon:
In CPython stdlib? re. In general? PLY http://www.dabeaz.com/ply/ appears to be popular choice, but I never used it.

And yes, as usual, it's quite different tasks (and results) of implementing tokenizer for X via a generic tokenization framework, and just adhoc, optimized tokenizer for X. (But I ended up reusing generic hand-coded LL(1) parser approach here, which is my favorite.)

dpgeorge:
PLY looks neat.  I tried it with uPy but it won't run because it uses `sys._getframe` to do munging of locals/globals.  A shame.  But anyway it's overkill to implement a simple XML parser.

For parsing XML in CPython there are many (too many I'd say) ways of doing it; eg lxml, ElementTree, minidom.  But my guess is that these all parse the entire file into RAM, which isn't going to work well in uPy for large files, especially when you only need a few bits of the XML data (or can operate on the data in a sequential way).

So I'd say having an XML tokeniser with which you can write your own simple XML parser for a given application is a good idea.

In CPython the xml stuff lives in xml/.  So the tokeniser could be xml/tok.py.  But I think xmltok.py is simpler and probably a better option so we don't clash with CPython in the future.  (Or we could have uxml/ package...).

Is it worth thinking about an XML writer (eg xmlprint), so as to have a compatible counterpart to xmltok?

danicampora:
[...]

PR #40, opened 2015-08-22
Draft implementation of xmltok module

pfalcon:
Per https://github.com/micropython/micropython-lib/issues/37 .

One thing to decide whether attribute lists should be explicitly terminated or not. As you can see, I initially wanted to have terminators, and tokens for them were added, but skipped implementing them. With terminators, parsing will be slightly easier, but of course there's more garbage to collect. The reason I didn't implement them is because I thought that for many useful algos (see e.g. gfind() helper), you'd need to consume "next" token anyway. Then token stream would be represented by pair of (head, tail), where head is a token (or maybe None) and tail is generator.

That's all speculative though - I played with parsing UPNP responses a bit, but switched away from it.

### DIFF
--- a/xmltok/test.xml
+++ b/xmltok/test.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0"?>
+<s:Envelope
+  xmlns:s="http://schemas.xmlsoap.org/soap/envelope/"
+  s:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">
+<s:Body>
+<u:GetConnectionTypeInfo
+  xmlns:u="urn:schemas-upnp-org:service:WANIPConnection:1">
+  foo bar
+  baz
+  
+</u:GetConnectionTypeInfo>
+</s:Body>
+</s:Envelope>
--- a/xmltok/test_xmltok.py
+++ b/xmltok/test_xmltok.py
@@ -1,0 +1,21 @@
+import xmltok
+
+expected = [
+('PI', 'xml'),
+('ATTR', ('', 'version'), '1.0'),
+('START_TAG', ('s', 'Envelope')),
+('ATTR', ('xmlns', 's'), 'http://schemas.xmlsoap.org/soap/envelope/'),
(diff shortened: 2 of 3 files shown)


pfalcon:
Ah, also - should be get string.isalnum() ? Helps with writing parsers.

dpgeorge:
There is going to be a lot of garbage if you generate START_TAG_END etc for each tag.  A lot of the time there are no attributes so it's not needed.  Perhaps there could be 2 different start tags, one when there are no attrs, one when there is 1 or more attrs, and this one has an end token?  Eg:
```
START_TAG "TagNoAttr"
...
START_TAG_ATTR "TagWithAttr"
ATTR attr1
ATTR_VAL val1
START_TAG_ATTR_END
```

Also, how about combining ATTR and ATTR_VAL into one returned "token"?  Attr vals are always strings (right?).  You're already returning a tuple with namespace + id, so why not add attr val?

> Ah, also - should [we impl] string.isalnum() ?

Yeah, sounds fine.

pfalcon:
> Perhaps there could be 2 different start tags, one when there are no attrs, one when there is 1 or more attrs, and this one has an end token? Eg:

Nope, don't like that idea. It should be resource-efficient, but not at the expense of API clarity and ease of use - why burden users with matching 2 token types instead of 1? ;-)

> Also, how about combining ATTR and ATTR_VAL into one returned "token"? Attr vals are always strings (right?). You're already returning a tuple with namespace + id, so why not add attr val?

Well, namespace support is the latest addition. Before I really was thinking such things as: attribute name can be collected when value arrives ;-). But yes, with namespace support probably makes sense to combine them - beyond values lifetime issues, there doesn't seem to be more memory usage with such arrangement.

pfalcon:
Ok, update code based on the discussion above. I guess, it can go to master in such shape, further adjustments may be made based on usage experience.

pfalcon:
Ok, so I'm merging this now, keeping in mind that behavior may be tweaked later.

Issue #41, opened 2015-08-25
itertools.cycle fails for generators


stinos:
The cycle implementation does not make a copy of the elements iterated over and as such only works for proper sequences like lists, but not for all types of iterables:

```
from itertools import cycle

def reverse(l):
    while l:
        yield l.pop()

for x in cycle(reverse([1, 2, 3])):
    print(x)
```

output:

```
3
2
1

```

Proposed fix: just copy the equivalent python implementation?

```
def cycle(iterable):
    saved = []
    for element in iterable:
        yield element
        saved.append(element)
    while saved:
        for element in saved:
              yield element
```

pfalcon:
> Proposed fix: just copy the equivalent python implementation?

Do you mean copy from CPython stdlib? Simple copy is never simple copy, there're also licensing issues to account for.

Also, the above implementation clearly bloats memory in case simple sequence is passed, so I'd suggest to add check for that and avoid copying in that case (I know there're bunch of corner cases, but let's resolve them on being hit). Feel free to submit a patch.

stinos:
> Do you mean copy from CPython stdlib?

No I just meant copying python implementation as shown from https://docs.python.org/3/library/itertools.html#itertools.cycle, not the actual C implementation.

dpgeorge:
That definition almost mandates the given implementation.  And, as it says, it's expected to use lots of memory.

stinos:
Fixed with #42

PR #42, opened 2015-08-25
Itertools cycle

stinos:
See #41

### DIFF
--- a/itertools/itertools.py
+++ b/itertools/itertools.py
@@ -4,7 +4,7 @@ def count(start=0, step=1):
         start += step
 
 def cycle(p):
-    while True:
+    while p:
         yield from p
 
 def repeat(el, n=None):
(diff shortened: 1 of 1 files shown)


pfalcon:
Well, neat. But extra recursion in yield from doesn't help memory usage again. So, better to rewrite to avoid it (feel free to leave that to me).

stinos:
Ah ok, fixed

pfalcon:
Thanks, pushed, with a bit more optimization: https://github.com/micropython/micropython-lib/commit/91563f06db292d322428e43ce0b8c43dd0676702 . Also, did similar change for islice() - to be able to write tests for all this. (But decided against optimizing sequence case for islice(), after all, there's normal slice() for them).

stinos:
Yes looks good, thanks.

PR #43, opened 2015-08-31
unittest: Add assertAlmostEqual and assertNotAlmostEqual

pramasoul:
To support more reuse of existing tests. Useful in their own right as well.


unittest/unittest.py
```diff
@@ -74,6 +114,7 @@ def assertRaises(self, exc, func=None, *args, **kwargs):
             raise
 
 
+
```
pfalcon: This empty line is stray ;-)

pfalcon:
Merged, thanks!

PR #44, opened 2015-09-01
random: A version of Random hacked-down from Cpython

pramasoul:
Uses 128-bit xorshift generator, with fallback to linear congruential, instead of the massive Mersenne Twister of Cpython implementation.

Where does this work belong? Does this belong in micropython-lib? How should it be improved?

### DIFF
--- a/random/_random.py
+++ b/random/_random.py
@@ -1,0 +1,22 @@
+#rng = 'lcg'
+rng = 'xorshift'
+if rng == 'xorshift':
+    try:
+        import sys
+        if sys.platform == 'pyboard':
+            from xorshift_pyb import Random
+        else:
+            from sys import maxsize
+            if maxsize > 1<<60:
+                from xorshift_py_64 import Random
+            else:
+                from warnings import warn as _warn
+                _warn("Can't use xorshift, falling back to lcg")
+                from lcg import Random
+    except Exception as e:
+        from warnings import warn as _warn
+        _warn(e)
+        _warn("Can't use xorshift, falling back to lcg")
+        from lcg import Random
+else:
+    from lcg import Random
--- a/random/_random_pyb.py
+++ b/random/_random_pyb.py
@@ -1,0 +1,4 @@
+# for pyboard
+
+from xorshift_pyb import XorShift128 as Random
+
--- a/random/lcg.py
+++ b/random/lcg.py
@@ -1,0 +1,28 @@
+from under_random import UnderRandom
+
+class Random(UnderRandom):
+    def __init__(self):
+        self.state = 7603603
+
+    def _rng(self):
+        # For these choice of numbers, see P L'Ecuyer,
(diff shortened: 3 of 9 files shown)


pfalcon:
> Where does this work belong? Does this belong in micropython-lib?

Well, my idea has always been to provide small, but adequate, PRNG via C codebase (perhaps urandom module), consistent for all ports. And my idea was to develop full-fledged random module from scratch, to make sure it's minimal and well suited to be used on bare-metal ports. I didn't make any progress on that because surprisingly, I didn't face a need for random module yet.

So, seeing this code drop with 700+ lines of CPython code, I can't say I'm too excited. Of course, if there's argumentation "CPython's doesn't really contain much unneeded stuff" (I didn't look closely myself), then maybe, but then it still should follow https://github.com/micropython/micropython-lib/wiki/ContributorGuidelines : 

> Start with committing 3rd-party source as-is, clearly stating in commit message from where it was taken (exact version, preferrably with URL to tarball, or VCS revision, preferrably with URL to it).

> Follow up with your changes as a next commit.

dpgeorge:
The code here is rather large.  Probably 50% is comments and/or doc strings, which are arguably not needed.

pramasoul:
This is helpful. I hit a need for repeatable PRNG for testing on the pyboard, and one thing led to another.

Cpython ```Random``` is well-tested, and is a standard interface. But it's got methods that that are just bloat, until the day you really need ```vonmisesvariate()```. Such methods should only be present in RAM when needed, perhaps by being late-loaded and dynamically-bound. Can I import a function and dynamically bind it to a class in micropython? Micropython ```cannot create 'bound_method' instances``` so this Cpython idiom doesn't work:
```python
from types import MethodType

class Foo():

    def fun(self, *args, **kwargs):
        from somewhere import fun
        self.fun = MethodType(fun, self)
        return self.fun(*args, **kwargs)
```
Would a ```bound_meth_make_new()``` be possible? Or is there a better way to keep the API but run lean?

pfalcon:
> so this Cpython idiom doesn't work:

Feel free to report this as a bug to micropython, so it wasn't forgotten. Preferrably, with references, because I for example first time see such code.

> Or is there a better way to keep the API but run lean?

Yes, I think starting with the minimal implementation, and then extending it only on the real need, carefully logging what needed more stuff, and why, is a better approach.

pfalcon:
MicroPython now has urandom.getrandbits() builtin function, from which other functions can be implemented. But it would be nice to do this functions by function, as actually needed, to make "random" module work on MCU systems.

jimmo:
We now have `getrandbits` (and `seed` which solves the repeatable part), and subsequently a micropython-lib random.py has been added which adds additional functions (e.g. random.shuffle). This would be a good base to start adding more functions from Python (e.g. random.choice for boards that don't include it).

PR #46, opened 2015-09-16
argparse: reproduce CPython's behaviour more closely

dxxb:
Make argparse compile with CPython.
Make test work with CPython.
Make description a kwarg for ArgumentParser().
Make add_argument() accept multiple option names (i.e. long and short options).

### DIFF
--- a/argparse/argparse.py
+++ b/argparse/argparse.py
@@ -126,7 +126,8 @@ class ArgumentParser:
 
         # print full information
         print()
-        print(self.description)
+        if self.description:
+            print(self.description)
         if self.pos:
             print("\npositional args:")
             for pos in self.pos:
--- a/argparse/argparse.py
+++ b/argparse/argparse.py
@@ -9,8 +9,8 @@ class _ArgError(BaseException):
     pass
 
 class _Arg:
-    def __init__(self, name, dest, action, nargs, const, default, help):
-        self.name = name
+    def __init__(self, names, dest, action, nargs, const, default, help):
+        self.names = names
         self.dest = dest
         self.action = action
         self.nargs = nargs
@@ -18,14 +18,14 @@ class _Arg:
         self.default = default
         self.help = help
 
-    def parse(self, args):
+    def parse(self, optname, args):
         # parse args for this arg
         if self.action == "store":
             if self.nargs == None:
                 if args:
                     return args.pop(0)
                 else:
-                    raise _ArgError("expecting value for %s" % self.name)
+                    raise _ArgError("expecting value for %s" % optname)
             elif self.nargs == "?":
                 if args:
                     return args.pop(0)
@@ -36,7 +36,7 @@ class _Arg:
                     n = -1
                 elif self.nargs == "+":
                     if not args:
-                        raise _ArgError("expecting value for %s" % self.name)
+                        raise _ArgError("expecting value for %s" % optname)
                     n = -1
                 else:
                     n = int(self.nargs)
@@ -53,12 +53,21 @@ class _Arg:
                         ret.append(args.pop(0))
                         n -= 1
                 if n > 0:
-                    raise _ArgError("expecting value for %s" % self.name)
+                    raise _ArgError("expecting value for %s" % optname)
                 return ret
         elif self.action == "store_const":
             return self.const
         else:
             assert False
+
+def _dest_from_optnames(opt_names):
+    dest = opt_names[0]
+    for name in opt_names:
+        if name.startswith("--"):
+            dest = name
+            break
+    return dest.lstrip("-").replace("-", "_")
+
 
 class ArgumentParser:
     def __init__(self, *, description=""):
@@ -66,7 +75,7 @@ class ArgumentParser:
         self.opt = []
         self.pos = []
 
-    def add_argument(self, name, **kwargs):
+    def add_argument(self, *args, **kwargs):
         action = kwargs.get("action", "store")
         if action == "store_true":
             action = "store_const"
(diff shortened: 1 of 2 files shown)


argparse/argparse.py
```diff
@@ -61,12 +64,12 @@ def parse(self, args):
             assert False
 
 class ArgumentParser:
-    def __init__(self, *, description):
+    def __init__(self, *, description = None):
```
pfalcon: Violates PEP8.

pfalcon:
@dpgeorge : This will depend on your word, as the original module author.

argparse/argparse.py
dxxb: Thanks. Fixed.

dpgeorge:
Sorry, but review of this will have to wait.

pfalcon:
@dpgeorge: Sure, no hurry, just to make sure you saw it and as a feedback to submitter.

pfalcon:
@dpgeorge : Friendly ping.

argparse/argparse.py
```diff
@@ -3,7 +3,10 @@
 """
 
 import sys
-from _collections import namedtuple
+try:
+    from _collections import namedtuple
+except ImportError:
+    from collections import namedtuple
```
dpgeorge: Does it need to run under CPython?

argparse/argparse.py
```diff
@@ -140,11 +151,9 @@ def parse_args(self, args=None):
 
     def _parse_args(self, args):
         # add optional args with defaults
-        arg_dest = []
```
dpgeorge: As far as I can tell, the changes in this file from this point onwards are not needed.  They don't change the behaviour, it's just a different way of accumulating the argument values.  Is that so?

argparse/argparse.py
dxxb: It doesn't need to run under CPython in production but it makes testing a lot easier by comparing the output of running the same test under CPython and uPy. We could have just
```
from collections import namedtuple
```
if this module depended on collections.

argparse/argparse.py
dxxb: I chose to keep _Arg() the same and store one instance of it for each valid option string passed to add_argument(). Without the changes below this module does not behave like CPython's argparse when parsing options having both short and long option.

argparse/argparse.py
dpgeorge: Most of the other modules in micropython-lib won't run under CPython (eg due to ffi module).

Don't you anyway want to compare with CPython's argparse, not this one?  Eg:
```
$ mkdir cpy
$ cp test_argparse.py cpy
$ cd cpy
$ python3 test_argparse.py
```
(This way CPython will import its own argparse, not ours.)

argparse/argparse.py
dpgeorge: I see.

I think it would be cleaner to have _Arg() take a list of alternative names for each option instead of creating a new _Arg() for each one.

The way you have it now, the usage() string prints multiple lines (and multiple descriptions) for options with short and long versions.

pfalcon:
So, based on the discussion, I see following issues with this patch:

1. Too many changes are done in one commit.
2. These changes are not described well enough (example: commit message says "Make description a kwarg for ArgumentParser().", but that's not what being done here - the description *was* a kwarg already, the patch just supplies a default).
3. They are not being tested well enough either. Specifically, if the aim is to achieve CPython compatibility, then CPython's test should be taken, and micropython-lib's implementation should be tested against it. Of course, majority of testcases will fail, so they need to be carefully disabled. As a specific example, considering following change in the test:

~~~
-assert args.a == False and args.b == 456
+assert args.a == False and args.b == "456"
~~~

I have no idea if it's a move in the right direction or not. (If @dpgeorge's original implementation captured option value with wrong type, this should have been fixed and documented on its own).


Summing up: I guess it's fair to ask for this patch to be reworked.

dpgeorge:
> If @dpgeorge's original implementation captured option value with wrong type, this should have been fixed and documented on its own

Yes, my original test is broken, in this and another place.

argparse/argparse.py
dxxb: ack. I agree with your preference. When I started making changes to this module I thought having one _Arg() instance per option would result in less changes and make the PR easier to accept. I will look into changing _Arg() the way you suggest.

argparse/argparse.py
dxxb: I thought it would be valuable to make a module work with both uPy and CPython when possible but it's not a requirement for me. Making it work with CPython was a side-effect of trying to figure if some unexpected behaviour I was seeing was caused by uPy.

dxxb:
@pfalcon, description was an arg not a kwarg. I am happy to rework and split this PR in smaller commits as per @dpgeorge's and your suggestions.

argparse/argparse.py
pfalcon: Fairly speaking, I have a similar change for gzip.py - to make it easier to test it with CPython (i.e. against CPython behavior). The fact that I don't merge it for months says that I'm not really sure if it's good or bad...

pfalcon:
> description was an arg not a kwarg

In Python, any arg can be used as kwarg. In original @dpgeorge's implementation, it even was kwarg-only (was specified after lone *).

> I am happy to rework and split this PR in smaller commits as per @dpgeorge's and your suggestions.

Thanks! https://github.com/micropython/micropython-lib/commit/fca6f0cc72265741ef04cc47b9504e81e8412620 and https://github.com/micropython/micropython-lib/commit/2c5e45b1b84c9aa7bcc81196eb319dc9b655b087 makes a head start with this and shows good way to proceed - separate simple patches, where it's easy to see what each one does and decided if it's right or wrong.

dxxb:
@pfalcon, you went and took all the low hanging fruits from me! ;-) I have pushed v2, it's just two commits with the second one being really trivial. I do not see a reason to split 48ee26d48bb052a1d3c119e50aa7cdf77e3eb835 further. If v2 is accepted I am going to add an extra commit to bring the module in line with PEP8 formatting.

dpgeorge:
Looks much better, thanks!

pfalcon:
@dxxb : Indeed, much better, thanks! The only thing I'd suggest is not to remove anything from the test - that code represents valid testcases (short options only handling) and should pass after your changes, right? Just add new complete testcases (i.e. starting with ArgumentParser creation) which check for both short&long options case.

dxxb:
@pfalcon, v3 pushed with amended tests (I basically reverted to how I did it in v1).
@dpgeorge, added a 3rd commit (dcf9c5c1972f442e28844ae3be57543c5e344f10) to test the case of positional arguments with dest but no name.

dpgeorge:
@dxxb, all is good except that the 3rd commit now makes the test fail :(

dxxb:
@dpgeorge, sorry about that I have amended and pushed a new version of the 3rd commit.

dpgeorge:
Merged in dd34477222c40f148abf1a48e58c48d4574c2c1e, e0aab10144fa819166e67746a964b6fe4a0ec406 and 5e9dcae281b531b71f9b486dcc4c2ba95bdb1284.

Thanks!

dxxb:
Thanks! I wanted to add an extra PEP8 cleanup commit to this series. Shall I create a new PR or there is no interest in it?

dpgeorge:
What PEP8 things are there to do?

dxxb:
Please see PR #48

Issue #49, opened 2015-10-19
contextlib: (re) add ExitStack?


dxxb:
ExitStack was added in 5557382b5e6673bee60b0c1658ff8bd752856ae4 from CPython 3.4.2 and subsequently removed in 92ef77c3d2b4dbd1e4ac257860327ebdaa967c33 maybe because it wasn't compatible with uPy out of the box.

I am using a slightly modified ExitStack taken from CPython lib for some unittests which need to run on both my development host and on the target device and I am wondering if there is any interest in having ExitStack added to contextlib. Thanks

dpgeorge:
I'd say it was removed simply because it didn't run out-of-the-box, and there was no need at the time to get it working.

pfalcon:
What seems as "removal" is actually following guidelines of https://github.com/micropython/micropython-lib/wiki/ContributorGuidelines ("When porting a module from CPython or other 3rd-party source" section). So, it wasn't removed, it just "wasn't added" to micropython-lib by the original submitter.

pfalcon:
Some random collection of comments regarding ExitStack:

1. This is the first time I hear about it.
2. Does anybody use it? Yeah, I figure you do, but does anybody use it in the sense of https://github.com/micropython/micropython/wiki/ContributorGuidelines#django-criteria , i.e. is there well established codebase which is useful to run on uPy, which uses ExitStack, and where ExitStack is few of missing features required to run it?
3. contextlib is one of those modules which is useful for PyBoard, so I'd like to keep it small, and from patches you quote, ExitStack is huge.


Actually, I'd like to keep it minimal, and current contextlib.py is already too big, and full of comments and stuff. Solve-it-all solution would be to create ucontexlib.py (in the own module of course), and let it be truly minimal (e.g. start with just contextmanager). Then contextlib.py can be fully CPython compatible (either importing ucontexlib, or just being completely separate, depending on what's easier to maintain).

This still leaves questions of why you use "slightly modified" version, instead of verbatim CPython version, and how to be about tests (as I mentioned, if something tries to be ~fully CPython compatible, it should also take tests from CPython and pass them).

So, if you'd be willing to resolve all these issues, it would be very welcome!

dxxb:
Hi,

Starting from @pfalcon's last question: ExitStack code from CPython does not run verbatim see https://github.com/dxxb/micropython-lib/commit/f16304f5890ada028328b9c80d1ee862c0dc76c6 for the (very limited) changes I made.

[...]

PR #50, opened 2015-10-21
Dxxb/ucontextlib

dxxb:
Minimal module spin off from existing contextlib as discussed in Issue #49.
Test were broken because of changes to unittest.run_class() so I moved to invoking unittest.main().
Not sure what to change the description in metadata and doctext to.


pfalcon:
I'll merge this as is, but preference should be on making as little as possible changes to CPython tests (by byte size of the resulting diff). By this logic, ContextManagerTestCase should not have been removed (regardless whether it's duplicated in ucontextlib module).

pfalcon:
You also apparently don't use make_metadata.py utility, but edit setup.py manually, which leads to mistakes. Please follow https://github.com/micropython/micropython-lib/wiki/ContributorGuidelines

dxxb:
Hi, I thought duplicating the test case would lead to PR rejection. I can fix that and use make_metadata.py to generate setup.py

pfalcon:
Merged with number of metadata changes.

dxxb:
Thanks

pfalcon:
> Hi, I thought duplicating the test case would lead to PR rejection.

This matter also covered in https://github.com/micropython/micropython-lib/wiki/ContributorGuidelines . And those guidelines are not arbitrary, but rooted in the fact that maintaining a project with dozen of individual components isn't easy, moreover if majority of components come from other projects.

> I can fix that and use make_metadata.py to generate setup.py

Well, I don't want to block on this, but please follow those guidelines in the future ;-)

dxxb:
Thank you. Point taken! :)

PR #53, opened 2015-11-14
OSX seems to return errno.EISDIR when mkdir() on an existing directory

njouanin:
Fixes issue #1624 in micropython project

### DIFF
--- a/upip/metadata.txt
+++ b/upip/metadata.txt
@@ -1,6 +1,6 @@
 srctype = micropython-lib
 type = module
-version = 0.6.1
+version = 0.6.2
 author = Paul Sokolovsky
 extra_modules = upip_errno, upip_gzip, upip_utarfile
 desc = Simple package manager for MicroPython.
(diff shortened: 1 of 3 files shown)


njouanin:
Sorry, I went too fast.

pfalcon:
Still not good enough: merges are not allowed, commits should be squashed (or represent 100% independent changes), there should be good description why a change is applied (especially if it's a workaround for bug, especially in a proprietary system). I've polished it myself: https://github.com/micropython/micropython-lib/commit/414c475e2e7394e7b5603786eef3d5fd7609c5a3

Note that this change needs to be propagated to "os" module too.

And thanks for contributing!

pfalcon:
Ok, I managed to miss that mkdir("/") is the culprit, so had to rebase to update the commit message: https://github.com/micropython/micropython-lib/commit/cd1fea8798065decf3907642dde6d5d4c515cd4a

PR #57, opened 2015-12-05
__future__: Add dummies for all available feature descriptions

mbuesch:
This will make statements such as
from __future__ import division
not fail.


pfalcon:
Merged, thanks.

PR #58, opened 2015-12-05
os: Import os.path

mbuesch:
On CPython os.path does not need to be imported explicitly.
This fixes the following issue:

```
>>> import os
>>> os.path.exists("")
  AttributeError: 'module' object has no attribute 'path'
````

### DIFF
--- a/os/metadata.txt
+++ b/os/metadata.txt
@@ -1,5 +1,5 @@
 srctype = micropython-lib
 type = package
-version = 0.3
+version = 0.4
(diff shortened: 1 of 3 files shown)


pfalcon:
Nope, I don't have plan to merge this - this pollutes heap with os.path even if user doesn't need it. Just import it explicitly, it's compatible with CPython.

mbuesch:
Ok, fair enough.

PR #59, opened 2015-12-05
functools: Add very simple implementation of reduce()

mbuesch:
The implementation was taken from
https://docs.python.org/3/library/functools.html


pfalcon:
Merged, thanks.

PR #60, opened 2015-12-05
socket: Add socket.error

mbuesch:
CPython raises socker.error (or classes derived from it) as exceptions on errors.
This change fixes code that tries to catch socket errors such as:

```
try:
  socket...
except socker.error as e:
  ...
```

usocket uses OSError. So we assign that to socket.error

### DIFF
--- a/socket/metadata.txt
+++ b/socket/metadata.txt
@@ -1,4 +1,4 @@
 srctype = micropython-lib
 type = module
-version = 0.3.2
+version = 0.3.3
 author = Paul Sokolovsky
--- a/socket/setup.py
+++ b/socket/setup.py
@@ -6,7 +6,7 @@ from setuptools import setup
 
 
 setup(name='micropython-socket',
-      version='0.3.2',
+      version='0.3.3',
       description='socket module for MicroPython',
       long_description="This is a module reimplemented specifically for MicroPython standard library,\nwith efficient and lean design in mind. Note that this module is likely work\nin progress and likely supports just a subset of CPython's corresponding\nmodule. Please help with the development if you are interested in this\nmodule.",
       url='https://github.com/micropython/micropython/issues/405',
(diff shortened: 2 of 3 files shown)


dpgeorge:
Apparently socket.error is deprecated: https://docs.python.org/3/library/socket.html#socket.error

pfalcon:
Yes, that's what I wanted to comment. Above in the same doc, it says:

> starting from Python 3.3, errors related to socket or address semantics raise OSError or one of its subclasses (they used to raise socket.error).

So, formally, this change is not needed. @mbuesch, if you think it's needed, please argue why (specific examples of software which need it are recommended).

mbuesch:
Ok, well. In that case I'll simply add a compatibility fix to my code.
(Specific examples can be found via google on interest)

pfalcon:
Well, anything can be found on google, but just anything should not be added to micropython-lib. Bottom line, is that there should be clearer documentation why a particular change was made. And clearer not just to @dpgeorge or me now, but to anyone at any later point of time. Yes, it's one-line change, but there can be a thousand of such changes,so without clear direction, the result can be messy.

mbuesch:
I found this issue, because my real-world application http://bues.ch/h/awlsim uses (used) socket.error for catching socket errors. And I don't think my application is special here. Before 3.3 that was the way to go. So there are a lot more applications doing that.
But as socket.error (and friends) are deprecated, I have no problem at all to add a workaround to my code. I will have to do so anyway, if socket.error finally is removed.

pfalcon:
@mbuesch : Thanks, that's exactly the kind of info I was looking for. It's really helpful to know how and for what people use micropython/micropython-lib.

So, while socket.error is deprecated in modern python3, it's supported for compatibility with older versions. Knowing that people want to run such compatible software with uPy helps, let's have this merged.

pfalcon:
Merged with somewhat modified commit message, to capture the discussion. Thanks!

PR #63, opened 2015-12-09
os.path: Add sep

mbuesch:
My application uses this to transform paths to an os-indepentent form and back (basically string.replace(os.path.sep, "/")).
We only support unix compatible '/'


pfalcon:
Merged, thanks.

PR #65, opened 2016-02-14
Pystone lowmem

pfalcon:


pfalcon:
@dpgeorge : Please check for soundness.

dpgeorge:
Yes, it looks good!  pyboard gives exact same pystone value for original script and this lowmem version.

One thing: it would be more accurate (and one could run the test with a smaller number of iterations) to use a timing function that wasn't just seconds as integers.  For pyboard I have a modified pystone.py that uses pyb.millis() [which should be changed to time.ticks_ms() and time.ticks_diff()].

pfalcon:
> pyboard gives exact same pystone value for original script and this lowmem version.

Well, on unix, lowmem is consistently and visibly slower (something like 50K vs 60K).

> should be changed to time.ticks_ms() and time.ticks_diff()

Ok, making this change.

pfalcon:
Merged.

PR #68, opened 2016-02-26
email.utils: fix urllib.parse dependency

balloob:
Fixes a typo in the `email.utils` dependencies. `setup.py` refers to `micropython-urlib.parse` instead of `micropython-urllib.parse` (notice the 2nd L). [Link to correct dependency.](https://pypi.python.org/pypi/micropython-urllib.parse)

### DIFF
--- a/email.utils/setup.py
+++ b/email.utils/setup.py
@@ -16,4 +16,4 @@ setup(name='micropython-email.utils',
       maintainer_email='micro-python@googlegroups.com',
       license='Python',
       packages=['email'],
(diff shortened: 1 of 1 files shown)


pfalcon:
Nice catch! But FYI, as https://github.com/micropython/micropython-lib/wiki/ContributorGuidelines explains, setup.py is actually autogenerated from metadata.txt. I've added fix to it too and merged as a479a5eb197940ba0a724cccb8d231e2a382a7f6

PR #70, opened 2016-03-07
Add LED.set() function

Ivoah:


pyb/pyb.py
```diff
@@ -9,13 +9,12 @@ def on(self):
     def off(self):
         self.f.write(b"0")
 
+    def set(self, v):
+        self.f.write(b"{}".format(0 if v < 0 else 255 if v > 255 else v))
```
SpotlightKid: I find this much clearer to read:

    self.f.write(b"{}".format(max(0, min(255, v)))

pfalcon: Maybe, but it involves 2 function calls. And the original is almost plain English anyway ;-).

Ivoah: Should I revert it to what I had before?

pfalcon:
@Ivoah : Sorry, I'm losing track of things ;-). I thought @SpotlightKid's comment was about a file already in a master, not about a pull request.

The problem with this patch is that this module is intended to emulate pyboard's "pyb" module. And pyb.LED class of pyboard doesn't have .set() method.

Ivoah:
Well, I could always make a pull request to the main pyb module too ;)

PR #72, opened 2016-05-07
Fix os.walk

palaviv:
I tried running walkdir (https://github.com/ncoghlan/walkdir) using micropython and noticed that os.walk is currently broken. I fixed the current function and added the ``followlinks`` and ``onerror`` arguments support. I didn't added tests for this yet. I will try to implement ``tempfile.mkdtemp`` on a different PR and then I will add the tests.

### DIFF
--- a/os/os/__init__.py
+++ b/os/os/__init__.py
@@ -143,12 +143,15 @@ def walk(top, topdown=True, onerror=None, followlinks=False):
     files = []
     dirs = []
     try:
-        for dirent in listdir(top):
-            if stat_.S_ISDIR(stat(top + sep + dirent)[0]):
-                if dirent != b"." and dirent != b"..":
-                    dirs.append(fsdecode(dirent))
+        for dirent in ilistdir(top):
+            mode = dirent[2] << 12
+            fname = fsdecode(dirent[0])
+            if fname in (".", ".."):
+                continue
+            if stat_.S_ISDIR(mode):
+                    dirs.append(fname)
             else:
-                files.append(fsdecode(dirent))
+                files.append(fname)
     except OSError as err:
         if onerror is not None:
             onerror(err)
--- a/os/os/__init__.py
+++ b/os/os/__init__.py
@@ -139,21 +139,25 @@ def listdir(path="."):
             res.append(fname)
     return res
 
-def walk(top, topdown=True):
+def walk(top, topdown=True, onerror=None, followlinks=False):
     files = []
     dirs = []
-    for dirent in ilistdir_ex(top):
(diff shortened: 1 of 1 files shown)


pfalcon:
Thanks for patch. As a quick reply, I would be reluctant to apply this right away, without thinking/let it stay in queue for a while. os.walk() is useful function, and it would be nice to keep it lean to be able to run on devices like micro:bit with 16K of RAM. Any feature added not strictly needed makes this less and less likely. At the very least, you should not drop usage of ilistdir_ex() which is there for a reason. Again, these are just first-impression comments, I'll need to think about it for a while (and/or get comments from other folks).

palaviv:
``ilistdir_ex`` was changed to ``ilistdir`` in 8884dbbee58cf4d43c15efa4b3686b3719c95f3c. Also it seems that ``ilistdir`` output has changed since walk was written. I changed the function to use ilistdir.
I am new to micropython so i would like any advice as to what will improve my implementation to fit  devices with small RAM. The problem with not allowing ``followlink`` and ``onerror`` is that will change the API from the cpython implementation thus limiting external libraries compatibility.

pfalcon:
> I am new to micropython so i would like any advice as to what will improve my implementation to fit devices with small RAM.

Please make sure first that you're familiar with "Zen on MicroPython" as presented https://github.com/micropython/micropython/wiki/ContributorGuidelines . 

> The problem with not allowing followlink and onerror is that will change the API from the cpython implementation thus limiting external libraries compatibility.

That's true for every other function/method in uPy. Changing one won't make much difference, changing all will make uPy lose its identity. The situation is not impasse, the situation requires careful consideration of each case. Consider for example https://github.com/pfalcon/pyelftools/commits/micropython-compat . I implemented few things which the lib missed, and patched the lib itself for few other. Next step is trying to post the changes back to author and see what they say (can't and won't find time for that). (Note that one patch there is already supefluous and can be removed - there was enough critical mass of interested parties to make a decision; most cases wait for such critical mass).

palaviv:
Thanks for the advice @pfalcon. I think that after reading the "Zen on MicroPython" I understand better what are the problems with my patch. Would removing the ``onerror`` support make this patch suitable? If not I would close this PR and open a different one that just fixes ``walk`` without adding new functionality.

pfalcon:
Yes, fixing ilistdir() definitely should be done, and I guess adding the followlink param in the simplest possible way (no stat()) makes sense, as all pieces are already there. Please make these 2 commits. Thanks!

palaviv:
Closing this PR. please see #73.

PR #73, opened 2016-05-11
Fix os.walk 2

palaviv:
This is a new PR instead of #72.
I changed os.walk to use ilistdir. I can't see how to add followlink's without a call to stat @pfalcon.

### DIFF
--- a/os/os/__init__.py
+++ b/os/os/__init__.py
@@ -142,18 +142,18 @@ def listdir(path="."):
 def walk(top, topdown=True):
     files = []
     dirs = []
-    for dirent in ilistdir_ex(top):
-        mode = dirent[3] << 12
-        fname = dirent[4].split(b'\0', 1)[0]
+    for dirent in ilistdir(top):
+        mode = dirent[1] << 12
+        fname = fsdecode(dirent[0])
         if stat_.S_ISDIR(mode):
-            if fname != b"." and fname != b"..":
-                dirs.append(fsdecode(fname))
+            if fname not in (".", ".."):
+                dirs.append(fname)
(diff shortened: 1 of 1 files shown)


os/os/__init__.py
```diff
@@ -142,18 +142,18 @@ def listdir(path="."):
...
     files = []
     dirs = []
-    for dirent in ilistdir_ex(top):
-        mode = dirent[3] << 12
-        fname = dirent[4].split(b'\0', 1)[0]
+    for dirent in ilistdir(top):
+        mode = dirent[1] << 12
+        fname = fsdecode(dirent[0])
         if stat_.S_ISDIR(mode):
-            if fname != b"." and fname != b"..":
```
pfalcon: You changed semantics of this line, even though MicroPython's ContributorGuidelines explicitly warns against it. Your change does superfluous allocation which in turn leads to cases like https://github.com/micropython/micropython/issues/2057 .

os/os/__init__.py
```diff
@@ -142,18 +142,18 @@ def listdir(path="."):
...
+            if fname not in (".", ".."):
+                dirs.append(fname)
         else:
-            files.append(fsdecode(fname))
+            files.append(fname)
     if topdown:
         yield top, dirs, files
     for d in dirs:
-        yield from walk(top + "/" + d, topdown)
+        yield from walk(top + sep + d, topdown)
```
pfalcon: ditto

pfalcon:
Fixed and merged, thanks. Following guidelines exactly will allow much quicker turnaround for patches.

PR #74, opened 2016-05-19
urequests: support ports other than default

garybake:
Sourced from https://github.com/kennethreitz/requests/blob/190cddd4ae7d649ea6d85d062fd6e4a69d0cf417/requests/packages/urllib3/util/url.py#L179

### DIFF
--- a/urequests/urequests.py
+++ b/urequests/urequests.py
@@ -32,23 +32,21 @@ class Response:
 
 def request(method, url, data=None, json=None, headers={}, stream=None):
     try:
-        proto, dummy, hostname, path = url.split("/", 3)
+        proto, dummy, host, path = url.split("/", 3)
     except ValueError:
-        proto, dummy, hostname = url.split("/", 2)
+        proto, dummy, host = url.split("/", 2)
         path = ""
     if proto != "http:":
         raise ValueError("Unsupported protocol: " + proto)
-    
-    if ':' in hostname:
-        host, port = hostname.split(':', 1)
-        if not port.isdigit():
-            raise ValueError('Unable to parse port')
+
+    try:
+        _host, port = host.split(':', 1)
+        host = _host
         port = int(port)
-    else:
-        host = hostname
+    except ValueError:
         port = 80
 
-    ai = usocket.getaddrinfo(host, 80)
+    ai = usocket.getaddrinfo(host, port)
     addr = ai[0][4]
     s = usocket.socket()
     s.connect(addr)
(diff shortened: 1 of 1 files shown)


urequests/urequests.py
```diff
@@ -32,12 +32,21 @@ def json(self):
...
         path = ""
     if proto != "http:":
         raise ValueError("Unsupported protocol: " + proto)
+    
+    if ':' in hostname:
+        host, port = hostname.split(':', 1)
+        if not port.isdigit():
+            raise ValueError('Unable to parse port')
+        port = int(port)
+    else:
```
pfalcon: Please avoid this if, and having both "host" and "hostname" vars.

urequests/urequests.py
```diff
@@ -32,12 +32,21 @@ def json(self):
...
     except ValueError:
-        proto, dummy, host = url.split("/", 2)
+        proto, dummy, hostname = url.split("/", 2)
         path = ""
     if proto != "http:":
         raise ValueError("Unsupported protocol: " + proto)
+    
+    if ':' in hostname:
+        host, port = hostname.split(':', 1)
+        if not port.isdigit():
```
pfalcon: Please avoid this if, int("foo") already throws ValueError.

garybake:
Thank you for the comments. 
For my own curiosity - the removal of the` if ':' in hostname:` was this due to efficiency with micropython or just a more pythonic way of doing it?

pfalcon:
> Please avoid this if, and having both "host" and "hostname" vars.

I'm sorry for formulating it in such confusing way. I meant "avoid else". I.e. instead of:

~~~
if ...:
    var = expr
else:
    var = const
~~~

preferrable to use:

~~~
var = const
if ...:
    var = expr
~~~

This will make code a tiny bit shorter and tiny bit more efficient.

garybake:
Gotcha - I've updated the file

urequests/urequests.py
```diff
@@ -39,7 +39,13 @@ def request(method, url, data=None, json=None, headers={}, stream=None):
     if proto != "http:":
         raise ValueError("Unsupported protocol: " + proto)
 
-    ai = usocket.getaddrinfo(host, 80)
+    port = 80
+    if ':' in host:
+        _host, port = host.split(':', 1)
+        host = _host
```
pfalcon: This is still superfluous. MicroPython allocates a memory slot for each local variable, more variables == more memory used. So, if it's possible to reuse a variable, that should be done - on the level of micropython-lib for sure. I'm fixing this myself.

pfalcon:
Merged with fixes, thanks.

PR #75, opened 2016-05-27
upysh: Minimalistic file shell use native Python syntax.

pfalcon:
E.g. cat("file") is equivalent of Unix shell's 'cat file'. With some tricks
though makes pwd and ls work without parens.

### DIFF
--- a/upysh/upysh.py
+++ b/upysh/upysh.py
@@ -1,0 +1,46 @@
+import sys
+import os
+
+class LS:
+
+    def __repr__(self):
+        l = os.listdir()
+        l.sort()
+        return str("\n".join(l))
+
+    def __call__(self, path):
+        l = os.listdir(path)
+        l.sort()
+        for f in l:
+            print(f)
+
+class PWD:
+
+    def __repr__(self):
+        return os.getcwd()
+
+pwd = PWD()
+ls = LS()
+
+def cd(path):
+    os.chdir(path)
+
+def head(f, n=10):
+    with open(f) as f:
+        for i in range(n):
+            l = f.readline()
+            if not l: break
+            sys.stdout.write(l)
+
+def cat(f):
+    head(f, 1 << 30)
+
+def help():
+    print("""
+This is 'upysh' help, for builtin Python help run:
+import builtins
+builtins.help()
+
+upysh commands:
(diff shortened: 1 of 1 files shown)


pfalcon:
@dpgeorge : I'd like to include something like this into esp8266 image, and as it's small and generic, propose for inclusion in micropython-lib.

@dhylands : FYI. I looked whether your https://github.com/dhylands/upy-shell could be something like this, but it's too big and too pyboard-specific. The idea here is to be truly minimal by even avoiding a special parser and truly generic.

dpgeorge:
Yeah this looks ok.  With fresh eyes, I would expect to write `ls()` and `pwd()` and would be surprised by only needing to do `ls`, `pwd`.  But I guess it's convenient.

upysh/upysh.py
```diff
@@ -0,0 +1,46 @@
+import sys
+import os
+
+class LS:
+
+    def __repr__(self):
+        l = os.listdir()
+        l.sort()
+        return str("\n".join(l))
```
dpgeorge: Why do you use str here?

pfalcon: Probably was str(l) before ;-).

pfalcon:
> With fresh eyes, I would expect to write ls() and pwd() and would be surprised by only needing to do ls, pwd.

Well, that can be made to work too ;-).

Anyway, I guess the main question - whether this should be in micropython-lib (vs separate project) is positive.

dpgeorge:
> Anyway, I guess the main question - whether this should be in micropython-lib (vs separate project) is positive.

I'd say it's ok in micropython-lib.  Would be a sensible addition to CPython (which I guess has nothing similar already?).

pfalcon:
> Would be a sensible addition to CPython (which I guess has nothing similar already?).

I doubt they would/should be interested - they don't have a usecase for it ("there's no 'native' shell, there's no memory to run anything fancy"). My concern is mostly due to the fact that other community members run other shell-like projects, and I don't want there be a concern then I'm trying to duplicate effort and get ahead of them by pushing directly to micropython-lib instead of running it as independent project. Anyway, I don't think there's any conflict, as other folks optimize for utility, while I for strict minimality and genericity.

Merged with minor tweaks.

PR #76, opened 2016-06-04
upysh: Added rm, rmdir, clear and simplified help

robert-hh:
- commands added: rm, rmdir and clear
- comands simplified: help does not need brackets any more
- help text rearranged
- pwd replaces empty string by "/"
- renamed help to man, to avoid conflicts with the built-in help()
- show file date & time with ls

### DIFF
--- a/upysh/upysh.py
+++ b/upysh/upysh.py
@@ -1,5 +1,6 @@
 import sys
 import os
+from time import localtime
 
 class LS:
 
@@ -12,10 +13,11 @@ class LS:
         l.sort()
         for f in l:
             st = os.stat("%s/%s" % (path, f))
+            tm = localtime(st[7])
             if st[0] & 0x4000:  # stat.S_IFDIR
-                print("   <dir> %s" % f)
+                print("%04d-%02d-%02d %02d:%02d    <dir> %s" % (tm[0], tm[1], tm[2], tm[3], tm[4], f))
             else:
-                print("% 8d %s" % (st[6], f))
+                print("%04d-%02d-%02d %02d:%02d %8d %s" % (tm[0], tm[1], tm[2], tm[3], tm[4], st[6], f))
 
 class PWD:
 
@@ -36,19 +38,19 @@ class CLEAR:
     def __call__(self):
         return self.__repr__()
         
-class HELP:
+class MAN:
 
     def __repr__(self):
         return ("""
-This is 'upysh' help. upysh commands:
+This is 'upysh' command list:
 
 cat(file)
 clear 
-cd("new_dir") 
+cd(new_dir) 
 ls 
 ls(object) 
 head(file [, #_of_lines]) 
-help
+man
 mkdir(newdir) 
 mv(from, to)
 newfile(file)
@@ -64,7 +66,7 @@ rmdir(empty_dir)
 pwd = PWD()
 ls = LS()
 clear = CLEAR()
-help = HELP()
+man = MAN()
 cd = os.chdir
 mkdir = os.mkdir
 mv = os.rename
--- a/upysh/upysh.py
+++ b/upysh/upysh.py
@@ -42,7 +42,7 @@ class MAN:
 
     def __repr__(self):
         return ("""
-This is 'upysh' command list:
+This is the 'upysh' command list:
 
 cat(file)
 clear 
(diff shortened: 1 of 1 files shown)


robert-hh:
Hello Paul. I really love that little script of yours and immediately added it to the Micropython board images. Even if urgently needed on PyBoard and Wipy, I find myself often issuing sh command at the REPL prompt. I polished and extended it a little bit, hopefully in your intent. 
Best Regards, Robert

pfalcon:
@robert-hh : Thanks. This contains some good ideas, but I'm not sure they all suitable for upysh as-is and right away. I've added rm, rmdir commands for now. I like an idea to use "man" for help, but people should know it's man, so I've made help to print on module load for now.

But: upysh is intended to be a filesystem shell, so things like "clear" are outside of its scope. It's also intended to be a very minimal shell, so printing date/time are also outside of its scope, at least for now (I find a file size to be a valuable info, and myself almost always pay attention to file size; contrary, I rather rarely consult file's time).

robert-hh:
Hello Paul,
Thanks for the comment. I was unsure about man- Help() collided with the
buitl-in command. But printing it a module start is surely a good
compromise. The 'clear' was requested by someone else. I rarely use it.
But the file date is more a matter of personal preference. I a m using
it as often as the file size, especially when working with data files,
which often have the same size. Then the date gets important. And
together with the PR2160 of Micropython, it can have a meaningful value.
Best Regards, Robert

dpgeorge:
I use "clear" a lot and would be +1 on including it.  It is related to "file" interaction, eg clearing the screen before cat'ing or ls'ing so you can see things clearly.

pfalcon:
@dpgeorge : Well, if *you* use it, what can we do? ;-). But mind my concern: I started upysh because I considered @dhylands' upy-shell too big. I don't want it to grow to be just a fork of upy-shell. Of course, any single function, we can add, but it's main purpose is to run on some 2K-heap (where it of course already won't run).

Anyway, are you sure it should be "clear" and not "cls" for example?

robert-hh:
'clear' is the unix/linux command, 'cls' is windows/dos. Since upysh mimics unix, it would be 'clear'. The aspect of RAM usage could be get less important for esp8266, if Damien's PR2067 would get mainstream.

pfalcon:
"cls" is what random human walking down the street would remember about clearing screen, because s/he read a BASIC book somehere in childhood (or for whatever) reason. It's also shorter.

> The aspect of RAM usage could be get less important for esp8266, if Damien's PR2067 would get mainstream.

upysh has nothing to do with esp8266, it's just a most minimal to still be useful file shell. Nor any hacks would help it - it itself a hack, made to influence as little as possible anything ride side by side with it. And memory will always be a scarce resource.

pfalcon:
> Anyway, are you sure it should be "clear" and not "cls" for example?

@dpgeorge : ping

pfalcon:
I've implemented "man" command now, thanks for the idea, @robert-hh !

dpgeorge:
I'd say it should be "clear", since, as @robert-hh says, all other commands are unixy.  But if its addition is controversial then don't worry about it.

BTW, if minimal is the goal then making extra clases for, eg, ls, just so you don't need to type parenthesis, seems a bit excessive.  But then, balancing minimal with usability is always hard (and I'd say the addition of "clear" is a usability thing).

robert-hh:
When placing upysh into frozen bytecode, the variant with classes uses about 1300 bytes of heap space, whereas a version without classes uses 640 bytes. That's my private copy, which is a little bit larger than Paul's, including clear and date & time on ls. From my previous fights for space I recall, that every class defintion take about 200 bytes of heap. That matches the difference roughly.

mad474:
1. Thank you all for upysh!
2. Today on a freshly flashed ESP8266: 

```
>>> from upysh import *
  File "upysh.py", line 35, in <module>
AttributeError: 'module' object has no attribute 'rmdir'

```
... because there's no uos.rmdir() on ESP8266. 
Although upysh.py bug fix 
```
- rmdir = os.rmdir
+ rmdir = os.remove
```
works (on ESP8266), I'm not sure if it would be the correct solution for all platforms? (I therefore canceled https://github.com/micropython/micropython-lib/pull/81).

pfalcon:
os.rmdir() was implemented in esp8266 port. "clear" command was added, based on implementation in this patch. Thanks!

PR #78, opened 2016-06-12
Makefile: Don't add _install_tmp to itself on make install.

ticpu:
When OpenWrt builder uses make install, some conditions cause make install to
include _install_tmp as a submodule and causes cp to error with
cp: _install_tmp/<file>.py is the same file as <full-path>/<file>.py

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ install:
 	else \
 	    for d in $$(find -maxdepth 1 -type d ! -name ".*"); do \
 	        echo $$d; \
+	        [ "$$d" = "./_install_tmp" ] && continue; \
 	        (cd $$d; sh -c $(CMD)); \
 	    done \
 	fi
--- a/os/metadata.txt
+++ b/os/metadata.txt
@@ -1,5 +1,5 @@
 srctype = micropython-lib
 type = package
-version = 0.4.2
+version = 0.4.3
 author = Paul Sokolovsky
 depends = ffilib, errno, stat
--- a/os/os/__init__.py
+++ b/os/os/__init__.py
@@ -87,6 +87,7 @@ def rename(old, new):
 def unlink(name):
     e = unlink_(name)
     check_error(e)
+remove = unlink
 
 def rmdir(name):
     e = rmdir_(name)
--- a/os/setup.py
+++ b/os/setup.py
@@ -6,7 +6,7 @@ from setuptools import setup
 
 
 setup(name='micropython-os',
-      version='0.4.2',
+      version='0.4.3',
       description='os module for MicroPython',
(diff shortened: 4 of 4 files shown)


pfalcon:
Ok, so do I understand right that you propose micropython-lib to work around openwrt issues? As whom do you make this request - as a maintainer or as a random user?

ticpu:
I assumed _install_tmp was created by micropython-lib.  All my apologies.

I'll take a look at why OpenWrt sets PREFIX in micropython-lib build directory.

pfalcon:
I grepped thru micropython-lib codebase and don't see anything like that.

ticpu:
That is exactly what I should have done before issuing a pull request, «git grep _install_tmp».

Here is the patch I posted to OpenWrt-Devel for reference.

```patch
Since make install path was hardcoded inside the build dir, it can sometime
fail with cp indicating that source and destination are the same since make
install will copy all py file in build dir to install dir. In this case,
install dir is in build dir causing a loop.
---
 lang/micropython-lib/Makefile | 5 +++--
 1 file changed, 3 insertions(+), 2 deletions(-)

diff --git a/lang/micropython-lib/Makefile b/lang/micropython-lib/Makefile
index b8af737..aefb8f4 100644
--- a/lang/micropython-lib/Makefile
+++ b/lang/micropython-lib/Makefile
@@ -22,6 +22,7 @@ PKG_SOURCE_VERSION:=bfbbf85a181d84e2494ea6f15be311734666bf67
 PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)-$(PKG_SOURCE_VERSION)
 PKG_SOURCE:=$(PKG_SOURCE_SUBDIR).tar.gz
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_SOURCE_SUBDIR)
+PKG_INSTALL_DIR:=$(BUILD_DIR)/_install_tmp
 PKG_BUILD_PARALLEL:=1
 
 include $(INCLUDE_DIR)/package.mk
@@ -43,12 +44,12 @@ endef
 
 MAKE_FLAGS:=\
 	-C $(PKG_BUILD_DIR) \
-	PREFIX=$(PKG_BUILD_DIR)/_install_tmp \
+	PREFIX=$(PKG_INSTALL_DIR) \
 	install
 
 define Package/micropython-lib/install
 	$(INSTALL_DIR) $(1)/usr/lib/micropython
-	$(CP) $(PKG_BUILD_DIR)/_install_tmp/* $(1)/usr/lib/micropython
+	$(CP) $(PKG_INSTALL_DIR)/* $(1)/usr/lib/micropython
 endef
 
 $(eval $(call BuildPackage,micropython-lib))
```

PR #83, opened 2016-07-26
umqtt.robust: publish() adapt to umqtt.simple

puuu:
`umqtt.simple` uses `publish(self, topic, msg, retain=False, qos=0)`, so
make `umqtt.robust` use the same. Otherwise, `retain` and `qos` will not work.


pfalcon:
Merged, nice catch.

PR #88, opened 2016-08-08
umqtt.simple: do not close the socket in ping()

puuu:
IMHO there is no need to close the socket in `ping()` and it  break the MQTT connection.


pfalcon:
Thanks, dunno how that close got in there.

PR #91, opened 2016-08-10
umqtt.simple: add MQTT connect authentication.

ngraziano:
To connect to a secured MQTT server it is necessary to be able to send username and password.

Limitation the total length of client id, user name and password must be under 111 characters.

### DIFF
--- a/umqtt.simple/umqtt/simple.py
+++ b/umqtt.simple/umqtt/simple.py
@@ -45,13 +45,13 @@ class MQTTClient:
         msg = bytearray(b"\x10\0\0\x04MQTT\x04\x02\0\0")
         msg[1] = 10 + 2 + len(self.client_id)
         msg[9] = clean_session << 1
-        if self.user and self.pswd:
+        if self.user is not None:
             msg[1] += 2 + len(self.user) + 2 + len(self.pswd)
             msg[9] |= 0xC0
         self.sock.write(msg)
         #print(hex(len(msg)), hexlify(msg, ":"))
         self._send_str(self.client_id)
-        if self.user and self.pswd:
+        if self.user is not None:
             self._send_str(self.user)
             self._send_str(self.pswd)
         resp = self.sock.read(4)
--- a/umqtt.simple/umqtt/simple.py
+++ b/umqtt.simple/umqtt/simple.py
@@ -7,7 +7,7 @@ class MQTTException(Exception):
 
 class MQTTClient:
 
-    def __init__(self, client_id, server, port=0, ssl=False):
+    def __init__(self, client_id, server, port=0, ssl=False, user=None, password=None):
         if port == 0:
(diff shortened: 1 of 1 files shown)


pfalcon:
@ngraziano : Thanks, please describe how you tested this change.

ngraziano:
Develop using documenation : http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718028

Test with : 
- mqtt server : mosquitto 1.4.9 configured with  allow_anonymous false and password_file.
- client ESP8266 firmware esp8266-20160808-v1.8.2-118-g3e5534c with code :
```
c = MQTTClient("umq", "192.168.0.20", user=b"nico",password=b"test")
c.connect()
c.publish(b"test",b"test")
```

umqtt.simple/umqtt/simple.py
```diff
@@ -43,9 +45,15 @@ def connect(self, clean_session=True):
         msg = bytearray(b"\x10\0\0\x04MQTT\x04\x02\0\0")
         msg[1] = 10 + 2 + len(self.client_id)
         msg[9] = clean_session << 1
+        if self.user and self.pswd:
```
dpgeorge: What if user and/or password is (intentionally) the empty string?  Is that allowed by the spec?  If so, this check here should be `if self.user is not None and self.pswd is not None:` (or it could be "optimised" to `if self.user is not None:`, or even better to `if self.user:` if the spec doesn't allow empty user names).

ngraziano: You are right, if have well read the documentation, the user name and password can be empty, so I must check with ```if self.user is not None and self.pswd is not None```

But I don't understand the "optimised" version, because I can't "_send_str" None.

I have 3 choice:
  1. test ```if self.user is not None and self.pswd is not None```
  2. init with empty bytes: ```def __init__(self, client_id, server, port=0, ssl=False, user=None, password=b''):``` and test ```if self.user is not None```
  3. make separate action for user name and password
``` python
if self.user is not None:
    msg[1] += 2 + len(self.user)
    msg[9] |= 0x80
if self.pswd is not None:
    msg[1] +=  2 + len(self.pswd)
    msg[9] |= 0x80
```
...
``` python
if self.user is not None:
    self._send_str(self.user)
if self.pswd is not None:
    self._send_str(self.user)
```

Which choice is the best for you ?

dpgeorge: I'd say it's an error to specify user without pswd, and vice versa.  If you specify one you should specify both.  Thus, it should be enough to check that user is not None (ie `if self.user is not None:`).

pfalcon:
@ngraziano : Cool, thanks for contribution! I squashed it, applied minor changes, and merged in fcec7f7c1d3485474596a0db38a83e7b9ce5870b.

PR #96, opened 2016-08-27
time: Added the functions mktime localtime and gmtime

rguillon:
A simple implementation of the functions mktime localtime and gmtime.

I didn't add a test/example file, I could if you think it's needed.

### DIFF
--- a/time/example_time_tuple.py
+++ b/time/example_time_tuple.py
@@ -1,0 +1,10 @@
+import time
+
+ts = 0
+for i in range(0, 31):
+    ts += 1 << i
+    ltt = time.localtime(ts)
+    gtt = time.gmtime(ts)
+    print("ts: ", ts, " local: ", tuple(ltt), " utc: ", tuple(gtt))
+    lts = time.mktime(ltt)
+    assert ts == lts
--- a/time/time.py
+++ b/time/time.py
@@ -1,4 +1,6 @@
 from utime import *
+import ustruct
+import uctypes
 import ffi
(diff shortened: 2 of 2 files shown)


pfalcon:
Thanks. Sorry, I missed to respond to your forum post, and actually was even ready to ack adding it on C side. I'm glad you tried ffi though, that's how it was originally intended to be done (i.e. if it's possible to do it on Py side, do it there). One question though - did you test this on both 32-bit and 64-bit platforms?

rguillon:
Nope I only ran my tests on a 64bits one. Would setting up a 32bit VM on a 64bits platform cover this case?

pfalcon:
Just build uPy with "make MICROPY_FORCE_32BIT=1" (after clean). Assuming you're on x86 of course.

rguillon:
Ok nice, it's working on both platforms. I've added a little example usable like micropython tests (it prints the same results from micropython and cpython3.5)

time/example_time_tuple.py
```diff
@@ -0,0 +1,10 @@
+import time
+
+ts=0
```
pfalcon: Sorry for backlogging with merging this. Unfortunately, there're multiple PEP8 inconsistencies in this example, so I have to request to fix them.

rguillon: Fixed, sorry about that.

pfalcon:
Merged, thanks! Update pushed to PyPI.

PR #97, opened 2016-08-29
umqtt.simple: Added last will capability

candale:
This PR introduces last will capabilities.


pfalcon:
I think the design with a separate method to set last will is good. Sorry for the delay with processing, merging, thanks.

pfalcon:
Well, one problem still: publish() uses retain=False, qos=0 qrg order, so going to  you the same for set_last_will() too.

pfalcon:
Also took a chance to use shorter var names. Merged.

PR #101, opened 2016-09-12
upip: Initialize socket with protocol family as gotten from resolver.…

michaelschrijver:
Fixes issue #2296 (tracked in micropython repo).

### DIFF
--- a/upip/upip.py
+++ b/upip/upip.py
@@ -120,7 +120,7 @@ try:
         #print("Address infos:", ai)
         addr = ai[0][4]
 
-        s = usocket.socket()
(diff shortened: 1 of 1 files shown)


dpgeorge:
That's an easy fix!  Were you able to test that it worked with IPv6 addresses?

michaelschrijver:
Yep it's working for me. I've only tested the unix port though.

dpgeorge:
Other ports should return the correct structure fromthe  getaddrinfo call, so they should also work.

pfalcon:
Merged, thanks. I'll push new version of upip a bit later.

PR #106, opened 2016-10-08
Simplification of delay during reconnect

kfricke:
The delay method was wasting an argument, static class member and only got called from one location.

### DIFF
--- a/umqtt.robust/umqtt/robust.py
+++ b/umqtt.robust/umqtt/robust.py
@@ -3,11 +3,7 @@ from . import simple
 
 class MQTTClient(simple.MQTTClient):
 
-    DELAY = 2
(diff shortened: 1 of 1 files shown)


pfalcon:
umqtt.robust.MQTTClient is intended to be configurable, extensible, and subclassable by a user.

kfricke:
But a base class should at least show reasonable behaviour and wasteless implementation.

pfalcon:
And that's exactly what it does.

kfricke:
The argument i passed to the method delay is never used. I did assume that your intent was to pass a dynamic delay, which does increase over time when the connection keeps failing.
Dropping the method was just another brave brevity.

PR #107, opened 2016-10-09
Remaining changes to support upip low-heap operation

pfalcon:
These are remaining changes to make upip be runnable on esp8266. @dpgeorge, if you have any comments, please share them, otherwise I plan to merge this soon and proceed with re-uploading 4K-optimized packages to PyPI. Note that micropython-pystone_lowmem already uploaded in 4K version for testing.

### DIFF
--- a/optimize_upip.py
+++ b/optimize_upip.py
@@ -1,0 +1,106 @@
+#
+# This script optimizes a Python source distribution tarball as produced by
+# "python3 setup.py sdist" command for MicroPython's native package manager,
+# upip. Optimization includes:
+#  * Removing metadata files not used by upip (this includes setup.py)
+#  * Recompressing gzip archive with 4K dictionary size so it can be
+#    installed even on low-heap targets.
+#
+import sys
+import os
+import zlib
+from subprocess import Popen, PIPE
+import glob
+import tarfile
+import re
+import io
+
+
+def gzip_4k(inf, fname):
+    comp = zlib.compressobj(level=9, wbits=16 + 12)
+    with open(fname + ".out", "wb") as outf:
+        while 1:
+            data = inf.read(1024)
+            if not data:
+                break
+            outf.write(comp.compress(data))
+        outf.write(comp.flush())
+    os.rename(fname, fname + ".org")
+    os.rename(fname + ".out", fname)
+
+
+def recompress(fname):
+    with Popen(["gzip", "-d", "-c", fname], stdout=PIPE).stdout as inf:
+        gzip_4k(inf, fname)
+
+def find_latest(dir):
+    res = []
+    for fname in glob.glob(dir + "/*.gz"):
+        st = os.stat(fname)
+        res.append((st.st_mtime, fname))
+    res.sort()
+    latest = res[-1][1]
+    return latest
+
+
+def recompress_latest(dir):
+    latest = find_latest(dir)
+    print(latest)
+    recompress(latest)
+
+
+EXCLUDE = [r".+/setup.py"]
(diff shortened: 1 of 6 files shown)


upip/upip.py
```diff
@@ -157,6 +161,9 @@ def install_pkg(pkg_spec, install_path):
     f3 = tarfile.TarFile(fileobj=f2)
```
dpgeorge: Why not reuse local f2 variable here, intsead of making a new f3?  Then there's less Python stack needed, and less risk that there are dangling pointers lef to f2/f3 on the Python value stack.  It also means you don't need to explicitly del f3 and f2, so saves on some bytecode.

upip/upip_utarfile.py
```diff
@@ -39,7 +39,13 @@ def readinto(self, buf):
...
 
     def skip(self):
-        self.f.read(self.content_len + self.align)
+        sz = self.content_len + self.align
+        if sz:
+            buf = bytearray(16)
+            while sz:
+                s = min(sz, 16)
+                self.f.readinto(buf, s)
+                sz -= s
```
dpgeorge: I guess you didn't want to rely on f having a seek method.  But would it be worth demanding this?  DecompIO.seek would have other uses.

optimize_upip.py
```diff
@@ -0,0 +1,106 @@
...
+def gzip_4k(inf, fname):
+    comp = zlib.compressobj(level=9, wbits=16 + 12)
+    with open(fname + ".out", "wb") as outf:
+        while 1:
+            data = inf.read(1024)
+            if not data:
+                break
+            outf.write(comp.compress(data))
+        outf.write(comp.flush())
+    os.rename(fname, fname + ".org")
```
dpgeorge: I would find "orig" clearer than "org".

optimize_upip.py
```diff
@@ -0,0 +1,106 @@
...
+    os.rename(fname + ".out", fname)
+
+
+def recompress(fname):
+    with Popen(["gzip", "-d", "-c", fname], stdout=PIPE).stdout as inf:
```
dpgeorge: Why not just use zlib.decompress?

upip/upip.py
pfalcon: To support debug logging, and potentially each of them may (should) have its own .close().

upip/upip_utarfile.py
pfalcon: My first motion was to implement seek(SET_CUR). But then I looked that no .ioctl was implemented in uzlib.DecompIO at all, plus it's not possible to provide correct return value for it. And I thought: why bother for a single usecase? If you think that implementing that (with invalid return value, like -1) is useful, I can queue it up.

pfalcon:
Ok, so I assume this stuff is "good enough", though there's as usual something to improve, but there're more important things to improve... So, going to merge and proceed with reuploading to PyPI.

upip/upip_utarfile.py
dpgeorge: DecompIO could only support SEEK_CUR with a non-negative argument, and the return value would have to be 0 I'd say (basically you're always at the start of the stream).  That's essentially a skip(n) method.  I think we can leave it for now but remember that if a use for skipping comes up again then we can implement it.

optimize_upip.py
dpgeorge: You could use zlib.decompressobj().

dpgeorge:
> Ok, so I assume this stuff is "good enough"

Yep.

pfalcon:
Ok, these changes were merged, and all packages republished with 4k gzip dict.

PR #117, opened 2016-11-08
There were no check_msg(self) function in umqtt.robust/umqtt/robust.py

Actpohomoc:


pfalcon:
Please follow https://github.com/micropython/micropython-lib/wiki/ContributorGuidelines next time.

pfalcon:
Also, I'm not exactly sure my intention with check_msg() was like that. The semantics of check_msg() is that it should complete immediately. Adding self.reconnect() to it makes it not such. I'm afraid, I have to reject this patch for now - if you use ckeck_msg(), please move catching exception and .reconnect() login into your app. And then you can handle it any way you like, and you let to do everyone else to do the same, in particular guarantee that check_msg() return immediately and can be used in main loops doing something else (otherwise, check_msg() may block your main loop indefinitely).

Actpohomoc:
Thank you.

Issue #116, opened 2016-11-07
SyntaxError in asyncio_slow


Anniywell:
```python
def async(coro):
    if isinstance(coro, Future):
        return coro
return Task(coro)
```

Change the function name async to ensure_future？

pfalcon:
Fixed in cad77291d5e17daae545168d2f3dca8508cb8122

PR #121, opened 2016-11-15
asyncio_slow: Rename 'async' to remove syntax error

betabrain:
I got syntax errors at `t = async(coro)` (line 50).

Now that `async` is a keyword it cannot double as a function name. I renamed all occurrences to `_async`.

This package has neither a `metadata.txt` nor a `setup.py`, please advise.

### DIFF
--- a/asyncio_slow/asyncio_slow.py
+++ b/asyncio_slow/asyncio_slow.py
@@ -47,7 +47,7 @@ class EventLoop:
         self.call_soon(_cb)
 
     def run_until_complete(self, coro):
-        t = async(coro)
+        t = ensure_future(coro)
         t.add_done_callback(lambda a: self.stop())
         self.run_forever()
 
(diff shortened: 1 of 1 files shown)


pfalcon:
No, please use `async_` , that's more or less common Python convention in such case. Not that asyncio_slow is not intended to be used, it's experimental code which is a bit sad to drop.

pfalcon:
Actually, the party interested to fix this well, should look how similar issue was fixed in the main uasyncio, that's it.

pfalcon:
https://github.com/micropython/micropython-lib/issues/116

betabrain:
In this case it should be `ensure_future` and this function should actually return a `Future` not a `Task`.

betabrain:
No. Task is correct, `Future` takes no `coro` parameter. Naming is less than intuitive on first sight, but I guess it's alright given that Future is the superclass.

Updated pull request.

pfalcon:
e6ca37faa81cb5c612d7a56fbb129ea1411e3ea7 (shows that it's useful to look in commit history, and then important to have good commit messages!)

pfalcon:
Fixed in cad77291d5e17daae545168d2f3dca8508cb8122

PR #123, opened 2016-11-22
time.gmtime() and time.localtime() return instance of struct_time.

nakagami:
For more compatibility

### DIFF
--- a/time/time.py
+++ b/time/time.py
@@ -16,14 +16,45 @@ localtime_ = libc.func("P", "localtime", "P")
 strftime_ = libc.func("i", "strftime", "sisP")
 mktime_ = libc.func("i", "mktime", "P")
 
+class struct_time:
+    def __init__(self, tm_year, tm_mon, tm_mday, tm_hour, tm_min, tm_sec, tm_wday, tm_yday, tm_isdst, tm_gmtoff=0, tm_zone='UTC'):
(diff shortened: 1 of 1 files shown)


dpgeorge:
A much simpler way to do this is to use a namedtuple.

nakagami:
Thanks, I did not know ucollections.namedtuple has implemented.

But tm_gmtoff and tm_zone is not memmbers of tuple.
So struct_time is not namedtuple, I think.

pfalcon:
I implemented namedtuple solution in d69aaba546576f053c5845809e5ba731b9e7e6df. While tm_gmtoff and tm_zone fields are indeed not supported, those are novelty fields themselves, not available in "classical Python". And micropython-lib lacked attribute-style access for 3 years, hopefully we can sustain couple more years without those fields.

Usecases for adding them are certainly welcome. The triggering usercase for adding namedtuple solution was desire to minimize patching of CPython's datetime module test, and lack of attribute access accounted for quite a lot of patching there (there's still a bunch of patching left beyond that). And of course, just a background that many CPython apps use such attribute access (but it's still discouraged for native MicroPython apps of course).

jimmo:
Closing due to inactivity. As mentioned above, this could be re-implemented with namedtuple. Also this has moved to the unix-ffi directory.

PR #143, opened 2017-01-08
uasyncio.core.py: Only import logging for debug

torwag:
As long as DEBUG is set to False, there is no need to import logging. Logging itself requires to be installed, thus it is an unneccary dependency for the normal user case.

### DIFF
--- a/uasyncio.core/uasyncio/core.py
+++ b/uasyncio.core/uasyncio/core.py
@@ -3,12 +3,12 @@ try:
 except ImportError:
     import time
 import utimeq
-import logging
-
 
 DEBUG = 0
 
-log = logging.getLogger("asyncio")
+if __debug__ and DEBUG:
+    import logging
+    log = logging.getLogger("asyncio")
 
 type_gen = type((lambda: (yield))())
 


pfalcon:
Fairly speaking, DEBUG can be changed programmatically...

torwag:
Hmmm. Okay, but does it make sens to call for the logging package, which is not a standard package, just in case some people might eventually, set DEBUG?
I wouldn't mind to have it like this on my PC, but afaik, we try to keep the footprint for MP as small as possible isn't it?

Would it make sense to encapsulate the entire debug stuff in an own package specifically tailored for MP? Within this package one could take care to import modules if needed and also catch exceptions if dependencies are missing, informing the user what exactly to do to activate debugging features. Another benefit would be a uniform debug interface among MP and MP libs. Furthermore, this debug package could address specific needs for MP like sending debug infos via serial in case REPL is not activated. Send debug via a socket to a host. Store logging into a permanent circular buffer (user selectable size) to catch seldom events during field tests or in production environments, etc.

Just an idea ;)

pfalcon:
> the logging package, which is not a standard package

What do you mean? https://docs.python.org/3/library/logging.html

> just in case some people might eventually, set DEBUG?

People *are* intended to enable debug logging in case they experience any issue.

> we try to keep the footprint for MP as small as possible isn't it?

Not at the expense of diagnostibility, debuggability and insight-ability of the code. People aren't interested likely use C or machine code already.

> own package specifically tailored for MP?

"we try to keep the footprint for MP as small as possible" and also avoid inventing our own wheels. logging is the standard package for logging, and so so far, issues in uasyncio are being discovered one after another. So, so far it's early to see how optimize logging completely away, though at later time, perhaps.

pfalcon:
Implemented in a way to allow user to enable logging after importing uasyncio in b2f69eb23b75a6915a06054132395f9bb97ec2bf . Thanks for the original report.

Issue #114, opened 2016-11-03
os.listdir() causes _Comparison between bytes and str_ warnings


dxxb:
This is caused by ```if fname not in (b".", b"..", ".", ".."):``` at https://github.com/micropython/micropython-lib/blob/master/os/os/__init__.py#L137

AFAICS uos.ilistdir() accepts a string or byte argument but always returns an entry's name as string. os.ilistdir() replacement instead always returns an entry's name as bytes. Would it be acceptable to make os.ilistdir() always return an entry's name as string and simplify os.listdir()? It would also make the warning go away.

pfalcon:
This was fixed.

PR #149, opened 2017-02-06
os: avoid str/bytes comparison warnings for "listdir"

sumpfralle:
Closes: #114

The "listdir" function caused the following warning for every single
directory item:
  Warning: Comparison between bytes and str


pfalcon:
Fixed a bit differently in 43ce994fa6a00f4503c84aeb4236dc4a9ca6a30b. Thanks for the submission still!

sumpfralle:
That looks much better - thank you!

PR #158, opened 2017-03-02
urequests: A Location-header with a 2XX status is not a redirect

MrShark:
When using the POST method you get the Location of the new
object in a Location header but it's not a redirect.
For details see for example RFC2616 10.2.2 and 10.2.3

### DIFF
--- a/urequests/urequests.py
+++ b/urequests/urequests.py
@@ -85,7 +85,7 @@ def request(method, url, data=None, json=None, headers={}, stream=None):
         if l.startswith(b"Transfer-Encoding:"):
             if b"chunked" in l:
                 raise ValueError("Unsupported " + l)
-        elif l.startswith(b"Location:"):
+        elif l.startswith(b"Location:") and not 200 <= status <= 299:
             raise NotImplementedError("Redirects not yet supported")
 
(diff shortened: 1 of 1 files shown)


pfalcon:
Thanks for the patch! I'd say it makes sense, but I'd like to check the standard reference that you quote, and also how upstream requests and/or urillib.request handle that case, so there were no unexpected surprises later. That may take some time.

MrShark:
Very well (I'm already using the patched version to get my application working :-) 

For reference, request seems even more strict requiring the status code to be one of a few 30X to consider it a redirect.
https://github.com/kennethreitz/requests/blob/master/requests/models.py#L700

urllib seems requires an even smaler number of status codes (it doesn't handle 308: permanent redirect)
https://github.com/python/cpython/blob/master/Lib/urllib/request.py#L651

Perhaps the logic should even be something like:

```python
 elif l.startswith(b"Location:") and status in (301, 302, 303, 307, 308):
    raise NotImplementedError("Redirects not yet supported")
```

pfalcon:
Sorry for the delay with this. Excellent supporting material for the change. Merged, thanks!

Issue #155, opened 2017-02-21
umqtt simple, __init__ needs wifi to work.


timpur:
Ive noticed that to create a new MQTTClient Object you need wifi since in the __init__  this line "self.addr = socket.getaddrinfo(server, port)[0][-1]" needs wifi to resolve the IP address. if there is no wifi the initialization of the new object fails and raises an error.

This can be really annoying when your working with a unstable wifi connection and your trying to code this situation. Wouldn't it be better is this address resolving occurs in the connect function? Every time you connect you also resolve the IP if a host name is used, if an IP was given then there is no need to resolve the IP.

Also little side point when using an host name like 'mqtt.example.com' this get resolved to an IP and stored in memory on the creation of the MQTT object which is the line above. But if the IP of this host name is dynamic then when the IP of the host name changes the MQTT object will never get the new IP and will never be able to reconnect unitl the object is recreated. This make long term connection for people using dynamic DNS for their WAN solution problematic. This can also be fixed by the above fix. moving this line to the connect function.

I think its silly that to create an MQTT object you need wifi to do so. 
Any thoughts on this ?

dmascord:
Hi Timpur,

I agree with you.  I have tested the following patch, and it works for me:

```
diff --git a/umqtt.simple/umqtt/simple.py b/umqtt.simple/umqtt/simple.py
index e3b5835..ab5572f 100644
--- a/umqtt.simple/umqtt/simple.py
+++ b/umqtt.simple/umqtt/simple.py
@@ -13,7 +13,8 @@ class MQTTClient:
             port = 8883 if ssl else 1883
         self.client_id = client_id
         self.sock = None
-        self.addr = socket.getaddrinfo(server, port)[0][-1]
+        self.server = server
+        self.port = port
         self.ssl = ssl
         self.ssl_params = ssl_params
         self.pid = 0
@@ -53,6 +54,7 @@ class MQTTClient:

     def connect(self, clean_session=True):
         self.sock = socket.socket()
+        self.addr = socket.getaddrinfo(self.server, self.port)[0][-1]
         self.sock.connect(self.addr)
         if self.ssl:
             import ussl

```

Cheers,

Damien

pfalcon:
@dmascord : That would be a duplicate of https://github.com/micropython/micropython-lib/pull/161. Please cooperate with @jcea who submits a proper patch, and let's have it merged.

timpur:
I left a comment over there. 
Either solution solves my problem.

[...]

PR #161, opened 2017-03-17
umqtt.simple: Do the DNS resolution every time you connect

jcea:
Do the DNS resolucion everytime you connect.

This is specially relevant for client "umqtt.robust" because you could not have network access when instantiate the object, so subsequent connections tries (after network is available) will fail because the "cached" address is wrong.

Resolving DNS in each connection is not a problem because reconnections should be sporadic and your DNS server (typically your local router) most probably cached the DNS reply anyway.

### DIFF
--- a/umqtt.simple/umqtt/simple.py
+++ b/umqtt.simple/umqtt/simple.py
@@ -13,7 +13,6 @@ class MQTTClient:
             port = 8883 if ssl else 1883
         self.client_id = client_id
         self.sock = None
-        self.addr = socket.getaddrinfo(server, port)[0][-1]
         self.ssl = ssl
         self.ssl_params = ssl_params
         self.pid = 0
@@ -53,7 +52,8 @@ class MQTTClient:
 
     def connect(self, clean_session=True):
         self.sock = socket.socket()
-        self.sock.connect(self.addr)
+        addr = socket.getaddrinfo(server, port)[0][-1]
+        self.sock.connect(addr)
         if self.ssl:
             import ussl
             self.sock = ussl.wrap_socket(self.sock, **self.ssl_params)


pfalcon:
> Resolving DNS in each connection is not a problem

But what if your DNS server went down, while everything else works? Then with you change, you (everyone!) is hosed, whereas with the code like it is now, it will likely continue to work (host addresses change quite rarely in general).

Bottom line: I wish someone else +1 this.

pfalcon:
And while we're waiting for it, could you please format commit message consistently with the existing commits (just do "git log" and you'll see how they should be). Thanks.

jcea:
DNS server failure should be both critical and temporal (if your DNS is failing, you sure are going to notice soon). In any case, with my proposal "connect()" will recover gracefully when your DNS is up again.

If you care about DNS dying, current code is even worse, because that "failure" will be cached forever unless you drop your MQTT object and create a new instance.

PS: I will be careful with future commits messages. Thanks for pointing out.

pfalcon:
Suppose that sounds good. But you didn't test your change ;-). (Hint: see what params you pass at the new location of getaddrinfo())

dpgeorge:
To make it even more robust you could catch any exceptions from the getaddrinfo and not update the cached address if the lookup fails.

timpur:
Hi, I submitted #155. 

This request solves my issue that i stated. But @dmascord had a slightly different solution but practically the same. Only difference is, to cache the address? Also as dpgeorge said you could use the old address if getting a new one fails? Something to think about? Either way I'm happy with these changes.

MrSurly:
Found this PR because I was going to submit something similar.  DNS resolution can be done in `__init__` , and re-done as needed, but shouldn't prevent `__init__` from succeeding.

puuu:
I am also interested in this patch.

IMHO doing DNS lookup during initialization is bad. If the class has a connect() method, I expect, that everything before calling this method can be done without network connection.

I have a typical setup, the ESP8266 starts up, read a sensor, publish the sensor data via mqtt and go in deep sleep for a few minutes. Short before running the connect() method, the program checks and waits for network connection with the hope that the ESP already has connected during initialization and sensor read out. This should reduce the awake time. Unfortunately, I did not expected that the umqtt already need network during initialization for DNS lookup. This was hard to debug, because this only happens at the final sensor location (bad network condition, long connection time), where I do not have a serial debug connection ...

So, to summarize: +1 

@jcea, you are willing to fix rework this patch as suggested by @pfalcon ?
> could you please format commit message consistently with the existing commits (just do "git log" and you'll see how they should be)

>  with it connect() accesses undefined local variables server, port (which then may be randomly picked from the outer context).

pfalcon:
@puuu , In case, feel free to submit an alternative PR, I'll merge it.

puuu:
> @puuu , In case, feel free to submit an alternative PR, I'll merge it.

@pfalcon please see #191 .

pfalcon:
Closing in favor of #191.

PR #173, opened 2017-05-03
umqtt.simple: Adds support for MQTT v3.1

dracode:
MQTT v3.1.1 uses a different connection string ("MQTT\x04") than MQTT v3.1 ("MQIsdp\x03").  This commit allows the library to support either version.  Older version of Mosquitto do not support v3.1.1, which was preventing MicroPython clients from connecting to those brokers.

### DIFF
--- a/umqtt.simple/umqtt/simple.py
+++ b/umqtt.simple/umqtt/simple.py
@@ -6,14 +6,17 @@ class MQTTException(Exception):
     pass
 
 class MQTTClient:
+    MQTTv31=3
+    MQTTv311=4
 
     def __init__(self, client_id, server, port=0, user=None, password=None, keepalive=0,
-                 ssl=False, ssl_params={}):
+                 proto=MQTTv311, ssl=False, ssl_params={}):
         if port == 0:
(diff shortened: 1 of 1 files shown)


agatti:
Thanks for the PR!

Unfortunately this has been overlooked for quite a while, but it looks like that the `umqtt.simple` module is still using v3.1.1.

According to https://en.wikipedia.org/wiki/Comparison_of_MQTT_implementations#Protocol_support there are very few MQTT servers that do not handle v3.1.1 as of today.  For the ones that do, if they have versions that do not handle v3.1.1, they are probably long obsolete (v3.1.1 was ratified twelve years ago, after all).

If you think this is still needed, the PR can be easily modified to the current version of the module: files have been moved around but the contents is more or less what the PR expects, with the only requested change to maybe use top-level `const()` values rather than having class-level constants.

dpgeorge:
As mentioned, MQTT v3.1.1 should be well supported now.  So I don't think we need this added complexity.

I'll close the PR, but feel free to reopen with a discussion about why it would be useful.

PR #183, opened 2017-05-24
xmltok: support comment markup in xml

ThunderEX:
basic support for comment which begin with `<!--` and end with `-->`

### DIFF
--- a/xmltok/test.xml
+++ b/xmltok/test.xml
@@ -9,5 +9,6 @@
   baz
   
 </u:GetConnectionTypeInfo>
+<!-- Comments, you won't see me after tokenize. -->
 </s:Body>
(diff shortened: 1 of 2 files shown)


pfalcon:
Thanks for the patch. I'd say it's worth to think a bit more about it though. E.g., whether comment token should be passed to client or no. Also, IIRC, comment syntax is <![ws]--[anything]--[ws]>.

So, let the patch stay here for awhile and collect more feedback/ideas.

pfalcon:
Ok, I guess we rather merge this than not. So, in fdf984d07de7f5cdc8e81b97c8c28a2c39e07b34, with a warning that comment tokenization may change later.

Thanks!

PR #184, opened 2017-05-27
Missed socket close. Result - buffer overflow on esp8266

ezdev128:


ezdev128:
Response already have close

PR #194, opened 2017-07-06
urequests: Added parsing response headers

manningt:
Uses response header content-length to determine how many bytes to read when reading the body.

Tested on 3 servers (xively, aws, other) using unix and esp micropython.  urequest never finished when doing a GET from AWS before this change.

removed redundant print from example_xively.

### DIFF
--- a/urequests/example_xively.py
+++ b/urequests/example_xively.py
@@ -7,7 +7,6 @@ r = requests.get("http://api.xively.com/")
 print(r)
 print(r.content)
 print(r.text)
-print(r.content)
 print(r.json())
 
 # It's mandatory to close response objects as soon as you finished
--- a/urequests/urequests.py
+++ b/urequests/urequests.py
@@ -6,6 +6,7 @@ class Response:
         self.raw = f
         self.encoding = "utf-8"
         self._cached = None
+        self.response_headers = {}
 
     def close(self):
         if self.raw:
@@ -16,7 +17,10 @@ class Response:
     @property
     def content(self):
         if self._cached is None:
-            self._cached = self.raw.read()
+            if "content-length" in self.response_headers:
+                self._cached = self.raw.read(int(self.response_headers["content-length"]))
+            else:
+                self._cached = self.raw.read()
             self.raw.close()
             self.raw = None
         return self._cached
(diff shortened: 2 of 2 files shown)


pfalcon:
> removed redundant print from example_xively.

Please don't add any unrelated changes to existing commits, if there's a separate issue, it should be a separate commit (likely, in a separate pull request). There's also a PEP8 error in your change.

But the main question - does your change affect memory usage? Did you try to profile it?

 > urequest never finished when doing a GET from AWS before this change.

That might be an HTTP protocol violation from AWS. This module makes an HTTP 1.0 request and expects a connection closure from server after it finished sending data.

manningt:
The urequest not completing with AWS does appear to be a protocol violation.  I tried using HTTP1.1 and explicitly adding Connection: close to the header, which did not resolve the issue with AWS.

As for memory, this modification consumes 3200 more bytes.  I wouldn't have expected that much, since I think the only new object is the response_headers dictionary.  Because of the increased memory size, its probably best to abandon this pull request; I will continue to use this version for my application.  Do you have any other suggestions?  Thanks for your time.

Below are the memory results for the master urequests and this pull request:
**Master**
```
MicroPython v1.9.1-125-g80b31dc0-dirty on 2017-07-07; ESP module with ESP8266
>>> import urequests
>>> import gc
>>> r = urequests.get('https://github.com/timeline.json')
>>> print(r.json())
{'message': 'Hello there, wayfaring stranger. If you\u2019re reading this then you probably didn\u2019t see our blog post a couple of years back announcing that this API would go away: http://git.io/17AROg Fear not, you should be able to get what you need from the shiny new Events API instead.', 'documentation_url': 'https://developer.github.com/v3/activity/events/#list-public-events'}
>>> print(gc.mem_free())
23728
>>> r.close()
>>> gc.collect()
>>> print(gc.mem_free())
26800
```
**Branch  manningt:urequest-with-content-length**
```
MicroPython v1.9.1-125-g80b31dc0-dirty on 2017-07-07; ESP module with ESP8266
>>> import trequests as urequests
>>> import gc
>>> r = urequests.get('https://github.com/timeline.json')
>>> print(r.json())
{'message': 'Hello there, wayfaring stranger. If you\u2019re reading this then you probably didn\u2019t see our blog post a couple of years back announcing that this API would go away: http://git.io/17AROg Fear not, you should be able to get what you need from the shiny new Events API instead.', 'documentation_url': 'https://developer.github.com/v3/activity/events/#list-public-events'}
>>> print(gc.mem_free())
20720
>>> r.close()
>>> gc.collect()
>>> print(gc.mem_free())
23664
```

PR #199, opened 2017-07-19
umqtt.simple: Added call to main() function in example scripts

spynappels:
Added call to main() function at end of script after the function definition. Matches other example scripts in umqtt.simple lib and is tested as functional on ESP8266 with MicroPython 1.9.1

Stefan

### DIFF
--- a/micropython/umqtt.simple/example_pub_button.py
+++ b/micropython/umqtt.simple/example_pub_button.py
@@ -28,3 +28,7 @@ def main(server=SERVER):
         time.sleep_ms(200)
 
     c.disconnect()
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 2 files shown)


dpgeorge:
The example_sub.py and example_pub.py scripts do have this call to main(), but not the example_sub_led.py and example_pub_button.py scripts.  So the latter pair either stay as they are, or both should have main() added to the end like is done here.

spynappels:
@dpgeorge No problem, I have added an additional commit adding the call to example_pub_button.py too.

pfalcon:
Please follow https://github.com/micropython/micropython-lib/blob/master/CONTRIBUTING.md

dpgeorge (approved):
Thanks for updating, and sorry this took forever to get back to.

This change looks fine, it makes all the examples consistent in calling `main()`.

PR #201, opened 2017-07-30
uasyncio: Fix for I/O mis-scheduling

peterhinch:
This has minimal performance impact in applications where the IORead/IOWrite mechanism is not in use.

### DIFF
--- a/uasyncio.core/uasyncio/core.py
+++ b/uasyncio.core/uasyncio/core.py
@@ -19,6 +19,7 @@ class EventLoop:
 
     def __init__(self, len=42):
         self.q = utimeq.utimeq(len)
+        self.reg_objects = 0
 
     def time(self):
         return time.ticks_ms()
@@ -53,6 +54,8 @@ class EventLoop:
         cur_task = [0, 0, 0]
         while True:
             if self.q:
+                if self.reg_objects and time.ticks_diff(self.q.peektime(), self.time()) <= 0:
+                    self.wait(0)  # Ensure pending IO gets scheduled.
                 # wait() may finish prematurely due to I/O completion,
                 # and schedule new, earlier than before tasks to run.
                 while 1:
--- a/uasyncio/test_stream_read.py
+++ b/uasyncio/test_stream_read.py
@@ -1,0 +1,35 @@
+try:
+    import uasyncio as asyncio
+except:
+    import asyncio
+
+try:
+    import utime as time
+except:
+    import time
+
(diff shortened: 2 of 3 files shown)


dbc:
@peterhinch  I'm wondering why it is necessary to guard the wait(0) with if self.reg_objects?  Is there something that breaks if we simply wait(0) on every trip through the loop?  The extra wait(0) would have some performance impact, of course, but then again so does the guard expression.  If the system has no StreamReader/Writer's, then I would argue that it is likely to be lightly enough loaded that the extra wait(0) is inconsequential.  If it is a busy system with Stream I/O, then then guard is wasted cycles to get 'True'.  There is a code simplication benefit of not having to reference count StreamReader/Writer's. Am I missing something that the guard is protecting against?

peterhinch:
I measured the performance with multiple coros waiting on a zero delay and with no stream readers or writers. Without the guard there was a significant performance impact. One of the strengths of the uasyncio implementation is its efficiency of scheduling coros (around 210μs latency on the Pyboard), and the overhead of the guard was too small to measure. Some hardware-related applications don't use the IORead/IOWrite mechanism but might be performance critical so I considered its small cost in code size worthwhile.

dbc:
@peterhinch There is nothing like actual measurement to bring clarity. Seems like the guard is worth having.

pfalcon:
Sorry for the delay with processing this, it's definitely an important issue.

@peterhinch , I liked the test, thanks for expressing it so precise. Tweaked a bit and merged with a description of the issue in 65605e3de8564ee6cdc766a527088465b6b76dc5.

Now, in [TDD](https://en.wikipedia.org/wiki/Test-driven_development) fashion, we can look for a solution.

pfalcon:
And following another principle, [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), the simplest solution is usually the best, at least for iteration no.1. The simplest change fixing the testcase is thus applied: eef054d98a6bc1e54d6b1b81f8ddb3dc448f7341.

We now can move to the optimization phase: interested parties are welcome to present multi-factor analysis that the solution above can be improved (for most cases, without deteriorating other cases).

pfalcon:
Ok, as there were no comments re: simple fix described above, it's now released to PyPI as uasyncio.core 1.5. Still keeping open for now.

pfalcon:
Ok, closing then, as the reported issue was fixed.

PR #202, opened 2017-08-12
base64: Add missing dependencies.

ccooper21:
The `base64` module includes imports for the `re`, and `binascii` modules, but its associated `upip` metadata did not specify as dependencies the `micropython-binascii`, and `micropython-re-pcre` packages that provide these modules.


pfalcon:
Merged, adding spaces to deps list. Thanks!

PR #203, opened 2017-08-12
binascii: Add required argument to <int>.to_bytes(...) call.

ccooper21:
The `a2b_base64(...)` function in the `binascii` module was failing due to an invalid call being made to the `<int>.to_bytes(...)` function.  The `<int>.to_bytes(...)` function requires two arguments.  The first specifies the number of bytes to return, and the second specifies the endianness of those bytes.  By definition, Base64 encoding is big endian.

This change also fixed the test cases for the `base64` module.

### DIFF
--- a/binascii/binascii.py
+++ b/binascii/binascii.py
@@ -68,7 +68,7 @@ def a2b_base64(ascii):
             #
             if leftbits >= 8:
                 leftbits -= 8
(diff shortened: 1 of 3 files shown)


pfalcon:
Merged, thanks. Also added testcase which exposed the issue.

PR #204, opened 2017-08-13
hmac: Calculate correct digest when non-trivial key is used.

ccooper21:
As per section 3 of [RFC 2104](https://www.ietf.org/rfc/rfc2104.txt), the HMAC algorithm must manipulate the specified key such that its length is the same as the block size of the specified digest algorithm.  If the key is shorter than the block size, then it will be right padded with 0x00 bytes to extend its length.  If the key is longer than the block size, it will be passed through the digest algorithm to produce a derivative key.  As per section 2, XOR operations must then be performed on the resulting key to produce two intermediate keys for internal use.  Every key produced by the algorithm steps described will have a length equal to the block size.

Previously, the HMAC implementation had a defect resulting in keys being used with the wrong value and length when:

- The specified key had a length less than or equal to the block size, and contained a byte with its most significant bit set (i.e. its value was between 128 and 255).

- The specified key had a length greater than the block size.  In this situation, the derivative key would almost certainly have at least one byte with its most significant bit set.

This incorrect behavior was a result of how the function that performs the XOR operation was implemented.  The function first used the `chr(...)` function to convert each resulting XORed byte, represented as an integer, to a string of length one.  Then, the `<str>.encode(...)` function was used to convert the string to a byte string with an intended length of one.  The code attempted to use the `ascii` encoding to accomplish this, but there are two problems in doing so.  

- The MicroPython core implementation ignores the specified encoding and always uses `utf-8` encoding.  This causes single bytes with their most significant bit set to be converted to a sequence of bytes.

- The `ascii` encoding, as defined in CPython, only supports input values in the range 0 to 127. 

This problem does not exist in CPython's HMAC implementation, because it uses the `<bytes>.translate(...)` function for this translation, which is not implemented in MicroPython.

### DIFF
--- a/hmac/hmac.py
+++ b/hmac/hmac.py
@@ -13,7 +13,7 @@ trans_5C = bytes((x ^ 0x5C) for x in range(256))
 trans_36 = bytes((x ^ 0x36) for x in range(256))
 
 def translate(d, t):
-    return b''.join([ chr(t[x]).encode('ascii') for x in d ])
+    return bytes(t[x] for x in d)
 
 # The size of the digests returned by HMAC depends on the underlying
 # hashing module used.  Use digest_size from the instance of HMAC instead.
--- a/hmac/metadata.txt
+++ b/hmac/metadata.txt
@@ -1,4 +1,4 @@
 srctype = cpython
 type = module
-version = 3.4.2-2
+version = 3.4.2-3
 depends = warnings, hashlib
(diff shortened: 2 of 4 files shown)


pfalcon:
Thanks for the fix! This bug turns out to have been contributed. However, a part of your detailed explanation of the issue should have been a part of the commit message. I condensed that description and included in the commit message.

ccooper21:
Thanks for merging this so quickly!  I'll keep your request in mind regarding commit messages for future pull requests.  I looked at some of the commit history, and it appeared that the preference was to keep commit messages minimal.

PR #207, opened 2017-08-25
Time api fix

sschwartzcpu:
time: parameter modification made to the strftime method. The modification allows the strftime method to accept tuples as its second argument so that it matches the behavior of the core python strftime method. Prior to this modification, the uPython strftime method would display
    "TypeError: can't convert tuple to int"
when the second argument of strftime was a tuple. The modification now allows conversion of gmtime as well. Before, it only allowed conversion of localtime.
strftime now accepts:
     time.strftime('%Y%m%d%H%M%S',time.gmtime())
     time.strftime('%Y%m%d%H%M%S',time.localtime())
     time.strftime('%Y%m%d%H%M%S',(2017, 1,  1,  23, 40, 39, 6, 222, 1))
     time.strftime('%Y%m%d%H%M%S') # defaults to localtime()

Also added a self test module: test_strftime.py
time: this test module has been added to test a parameter modification made to the strftime method. The modification allows the strftime method to accept tuples as its second argument so that it matches the behavior of the core python strftime method. Prior to this modification, the uPython strftime method would display "TypeError: can't convert tuple to int" when the second argument of strftime was a tuple.

### DIFF
--- a/time/time.py
+++ b/time/time.py
@@ -27,15 +27,13 @@ def _c_tm_to_tuple(tm):
 
 
 def strftime(format, t=None):
-    if t is None:
-        t = time()
+    if (t is None):
+        t = localtime() 
 
-    t = int(t)
-    a = array.array('i', [t])
-    tm_p = localtime_(a)
     buf = bytearray(32)
-    l = strftime_(buf, 32, format, tm_p)
+    l = strftime_(buf, 32, format, _tuple_to_c_tm(t))
     return str(buf[:l], "utf-8")
+
 
 
 def localtime(t=None):
--- a/time/test_strftime.py
+++ b/time/test_strftime.py
@@ -1,0 +1,101 @@
+import sys
+import getopt
+import time
+
+
+def arg_check(argv):
+    try:
+        opts, args = getopt.getopt(argv[1:], 'hv', ['help', 'verbose'])
+    except getopt.GetoptError:
+        print(argv[0] + ' -h -v')
+        sys.exit(2)
+    for opt, arg in opts:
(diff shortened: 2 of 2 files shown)


time/time.py
```diff
@@ -27,17 +27,15 @@ def _c_tm_to_tuple(tm):
 
 
 def strftime(format, t=None):
-    if t is None:
-        t = time()
+    if (t is None):
```
pfalcon: Parens are superfluous. As well as the change of unrelated code.

time/time.py
```diff
@@ -27,17 +27,15 @@ def _c_tm_to_tuple(tm):
...
-    t = int(t)
-    a = array.array('i', [t])
-    tm_p = localtime_(a)
     buf = bytearray(32)
-    l = strftime_(buf, 32, format, tm_p)
+    l = strftime_(buf, 32, format, _tuple_to_c_tm(t))
     return str(buf[:l], "utf-8")
 
 
+
```
pfalcon: Stray change.

pfalcon:
Thanks for the patch! However, please avoid making changes not directly related to a fix being made. Cleaned up and merged as ac66b1ca1aeb98938b11a0608b98b50942646096 . The test is useful, but needs more review and likely changes. For example, tests shouldn't require command line params and depend on libs (like getopt) to parse them.

pfalcon:
Yep, sorry, a test should be in general very simple and clear. I've extracted your test data, but rewrote the test script as 4cae9a3fc0c626c0235ddb233b52c43af45c1f69.

Thanks again for contribution.

PR #208, opened 2017-09-18
random: randrange() implementation

pfalcon:


pfalcon:
@dpgeorge, please review this impl for correctness.

dpgeorge:
> please review this impl for correctness.

It's not uniform when using integers.  The last line of the function is mapping a range of [0..pwr2-1) to [0..upper-1) using floor division and this will bias certain integers.  Eg:
```
randrange(3)
-> upper = 3
-> pwr2 = 4
-> bits = 2
-> r = [0, 1, 2, 3] # one of these
-> return [0, 0, 1, 2] # one of these
```
So for `randrange(3)` it returns 0 with 0.5 probability, and 1 and 2 with 0.25 probability (assuming getrandbits is uniform, which it should be).

As far as I can tell there are only 2 ways to do it:
1. Keep calling `getrandbits(pwr2)` until the result is less than `upper` (this is how the C version does it in extmod/modurandom.c).
2. Use the formula `getrandbits(pwr2 * upper)) // pwr2`.  But that's likely going to go outside the range that `getrandbits` can handle (you could improve it by using lcm/gcd but that's getting even more complex).

pfalcon:
@dpgeorge, thanks for review.

> using floor division and this will bias certain integers.

Yep, I expected that, but the idea was "with sufficiently large numbers, that will be almost unnoticeable". Your example shows that the problem exists.

> Keep calling getrandbits(pwr2) until the result is less than upper (this is how the C version does it in extmod/modurandom.c).

Intuitively, that would lead to bias, but maybe I need to both refresh myself on random number theory and think about it for a while.

> Use the formula getrandbits(pwr2 * upper) // pwr2

Hm, stubborn me intuitively think about oversampling by just a fixes number of bits...
 

Anyway, will need to look at all this with a fresh mind (which I won't have for a while).

dpgeorge:
> Use the formula getrandbits(pwr2 * upper) // pwr2.

Actually, this won't work either.  There's no way any power of 2 can contain an arbitrary integer (not 2) as a factor.

pfalcon:
> Keep calling getrandbits(pwr2) until the result is less than upper (this is how the C version does it in extmod/modurandom.c).

Ok, I see, this wouldn't work with any non-uniform distribution, but as long as it's uniform and unbiased, it works. Will redo like that.

pfalcon:
Fixed version merged.

PR #209, opened 2017-09-21
Fixed SyntaxError in subscribe()

Himura2la:
```python
>>> 1.to_bytes(1)
  File "<stdin>", line 1
SyntaxError: invalid syntax for number
```


pfalcon:
So, what's exactly being fixed here?

Himura2la:
I had an exception on this string, because `int()` has no `to_bytes()` method. I call `subscribe()` without QoS and it makes QoS become `1`.

pfalcon:
> because int() has no to_bytes() method. 

Of course it has:

~~~
MicroPython v1.9.2-96-g3c7f2184b-dirty on 2017-09-19; linux version
>>> (1).to_bytes(1, "little")
b'\x01'
~~~

Himura2la:
I'll check it once again with the second argument, but the version cloned yesterday (with one argument) did not work for me, I posted the output from ESP8266 in the first comment

Himura2la:
So do you see the issue now?

pfalcon:
I don't, sorry. As far as I can tell, the issue you had was because you used a wrong branch of the project. That branch was rebased since, but why did you use it in the first place?

Himura2la:
I don't see the correct syntax neither in **master** not in **mqtt** branch. If I don't pass a tuple to `qos`, it becomes int https://github.com/Himura2la/micropython-lib/blob/master/umqtt.simple/umqtt/simple.py#L144 and failing here https://github.com/Himura2la/micropython-lib/blob/master/umqtt.simple/umqtt/simple.py#L152

Just tested on my ESP8266, ints have no `.to_bytes()`

```python
MicroPython v1.9.1-8-g7213e78d on 2017-06-12; ESP module with ESP8266
>>> 1.to_bytes(1, "little")
  File "<stdin>", line 1
SyntaxError: invalid syntax for number
>>>
```

Himura2la:
I don't see the correct syntax neither in **master** not in **mqtt** branch. If I don't pass a tuple to `qos`, it becomes int https://github.com/micropython/micropython-lib/blob/master/umqtt.simple/umqtt/simple.py#L144 and failing here https://github.com/micropython/micropython-lib/blob/master/umqtt.simple/umqtt/simple.py#L152

Just tested on my ESP8266, ints have no `.to_bytes()`

```python
MicroPython v1.9.1-8-g7213e78d on 2017-06-12; ESP module with ESP8266
>>> 1.to_bytes(1, "little")
  File "<stdin>", line 1
SyntaxError: invalid syntax for number
>>>
```

dpgeorge:
@Himura2la the problem was simply that `1.to_bytes(...)` looks like a decimal number to Python and so that's why it gives a syntax error.  It can be fixed by putting a space before the period: `1 .to_bytes(...)`, or as shown above putting the number in parenthesis: `(1).to_bytes(...`).  Note that it's not a tuple with one element, that would be `(1,).to_bytes(...)` which doesn't work (tuple doesn't have a to_bytes method).  And since `qos` is not an integer literal that works fine without a space.  This is all standard Python behaviour.

Himura2la:
Thanks for explanation, there are some quite confusing dark corners of Python...

PR #211, opened 2017-10-03
pwd: add basic implementation of pwd

xrmx:
The only function implemented is getpwnam

### DIFF
--- a/pwd/metadata.txt
+++ b/pwd/metadata.txt
@@ -1,0 +1,5 @@
+srctype = micropython-lib
+type = module
+version = 0.1.0
+author = Riccardo Magliocchetti
+
--- a/pwd/pwd.py
+++ b/pwd/pwd.py
@@ -1,0 +1,24 @@
+import ffilib
+import uctypes
+import ustruct
+
+from ucollections import namedtuple
+
+
+libc = ffilib.libc()
+
+getpwnam_ = libc.func("P", "getpwnam", "s")
+
+
+struct_passwd = namedtuple("struct_passwd",
+    ["pw_name", "pw_passwd", "pw_uid", "pw_gid", "pw_gecos", "pw_dir", "pw_shell"])
+
+
+def getpwnam(user):
+    passwd = getpwnam_(user)
+    if not passwd:
+        raise KeyError("getpwnam(): name not found: {}".format(user))
+    passwd_fmt = "SSIISSS"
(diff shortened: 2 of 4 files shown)


pfalcon:
Thanks, would be a nice addition! But did you actually test your code? When you say "s" for struct.unpack(), you say "there's a string in this place of the structure", and as you don't specify a length, it's length 1. But struct passwd contains a *pointer* to a string there, not a string. See top comment in py/modstruct.c on how to resolve it.

pwd/metadata.txt
```diff
@@ -0,0 +1,5 @@
+srctype = micropython-lib
+type = module
+version = 0.0.1
```
pfalcon: Version should be 0.1, 0.0.x is for completely dummy modules.

pwd/pwd.py
```diff
@@ -0,0 +1,25 @@
+import ffilib
+import uctypes
+import ustruct as struct
```
pfalcon: Maybe use ustruct directly?

pwd/pwd.py
```diff
@@ -0,0 +1,25 @@
...
+
+from ucollections import namedtuple
+
+
+libc = ffilib.libc()
+
+if libc:
```
pfalcon: Other ffilib-using modules don't have this if, maybe it's not needed here either?

xrmx: I've used os as reference, but will remove

pwd/pwd.py
xrmx: fine for me

xrmx:
lol, i've overlooked quite badly the actual output :)

pwd/metadata.txt
xrmx: ok

xrmx:
Updated branch, now the output looks fine, thanks for the review!

pwd/pwd.py
pfalcon: I see. Well, I believe there were reports that libc can't be imported e.g. on Windows, maybe on MacOS. So, that "if" was added to not make the module fail on import and just offer constants and things from "uos" module. We don't need such handling in most other modules, so thanks for removing it.

pfalcon:
Merged, thanks! Note that setup.py is autogenerated from metadata.txt: https://github.com/micropython/micropython-lib/wiki/ContributorGuidelines , so that part was tweaked.

xrmx:
Thanks for merging! Completely missed the contribution guidelines :)

PR #213, opened 2017-10-06
select: Add POLLPRI mask for use with uselect module.

reidwagner:
Add POLLPRI as an attribute of select module as suggested in micropython/micropython#3345.


reidwagner:
Closed because this really should just be included with a full implementation of poll.

pfalcon:
Sorry, I don't see any problem with this patch, except that it should be accompanied by f5bc4d8ea3a0b9fd54c7754867829b5c2516dd0c. Merged, thanks.

reidwagner:
That makes sense. Appreciated.

PR #216, opened 2017-10-08
select: Recompute timeout after EINTR

reidwagner:
As detailed in [PEP 475](https://www.python.org/dev/peps/pep-0475/), timeout is recomputed before retrying an interrupted epoll_wait. Previously epoll_wait was restarted with the original timeout, meaning repeated interruptions could result in an extended or indefinite timeout.

I've copied the strategy used in CPython [here](https://github.com/python/cpython/blob/4d3f084c035ad3dfd9f8479886c41b1b1823ace2/Modules/selectmodule.c#L1489), where a deadline is computed at the beginning of poll().

Regarding Python versions, I noted that 3.3 is the recommended reference, and that 475 is [not yet implemented in micropython](https://github.com/micropython/micropython/issues/1329), but because it is here and with an explicit TODO about timeout I figured this was an appropriate change.

### DIFF
--- a/select/select.py
+++ b/select/select.py
@@ -3,6 +3,7 @@ import ustruct as struct
 import os
 import errno
 import ffilib
+import utime
 from uselect import *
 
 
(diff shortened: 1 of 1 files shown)


pfalcon:
Unfortunately, epoll.poll() timeout is in seconds (yeah, inconsistently with poll.poll()): https://docs.python.org/3/library/select.html#select.epoll.poll

reidwagner:
I noticed this but assumed that it was intentional that the micropython-libs implementation of epoll.poll uses ms (epoll_wait takes ms and timeout is passed unmodified).

Should epoll.poll be changed to use seconds?

pfalcon:
Well, there's definitely no intention to be incompatible with CPython in such a way as for the same method to take different units. I applied c8f9cf1dcac7894b9632aa51616e7b83c79a5caf, feel free to rebase/rework on top of that.

reidwagner:
I rebased and it turned out fine.

I also looked into timeout as a float, as it is in CPython, and rounding it up to the nearest millisecond as described in CPython issues [20311](https://bugs.python.org/issue20311) and [20452](https://bugs.python.org/issue20452). But I ran into float issues that made it not immediately apparent to me how to proceed, and I figured it's better left for another PR.

pfalcon:
Merged, thanks!

PR #222, opened 2017-10-23
select: Convert float timeout to int with math.ceil.

reidwagner:
In CPython, timeout is a float and the value rounded up to the nearest millisecond. The rounding up is discussed in issues [20311](https://bugs.python.org/issue20311) and [20452](https://bugs.python.org/issue20452).

This will allow users to specify poll timeouts with greater precision than 1 second. But, this solution is susceptible to float round errors and is only accurate to +/- 1ms. E.g. epoll.poll(.001) actually results in a timeout of 2ms.

I don't know if its worth it to introduce an inaccuracy. Until 2008 epoll_wait actually had a resolution between 1 and 10ms: https://lwn.net/Articles/296578/, https://lwn.net/Articles/296398/.

I think the decimal module would help here if it existed.

### DIFF
--- a/select/select.py
+++ b/select/select.py
@@ -4,6 +4,7 @@ import os
 import errno
 import ffilib
 import utime
+import math
 from uselect import *
 
 
@@ -92,7 +93,7 @@ class Epoll:
         return res
 
     def poll(self, timeout=-1):
-        return self.poll_ms(-1 if timeout == -1 else timeout * 1000)
+        return self.poll_ms(-1 if timeout == -1 else math.ceil(timeout * 1000))
 
     def close(self):
         os.close(self.epfd)


pfalcon:
That looks complicated and misplaced. Why not:

~~~
-return self.poll_ms(-1 if timeout == -1 else timeout * 1000)
+return self.poll_ms(-1 if timeout == -1 else int(timeout * 1000))
~~~

reidwagner:
I originally wrote
```
-return self.poll_ms(-1 if timeout == -1 else timeout * 1000)
+return self.poll_ms(-1 if timeout == -1 else math.ceil(timeout * 1000))
```
But I decided to encapsulate the cast in poll_ms in case there was any use for that function outside of poll. If not, then I agree above is cleaner. 

I use math.ceil() instead of int() because int() truncates towards zero and the value is explicitly rounded up in CPython.

pfalcon:
> poll_ms was better so both functions would accept floats.

Well, the whole idea of poll_ms() is that it accepts non-ambiguous numbers.

> I use math.ceil() instead of int() because int() truncates towards zero and the value is explicitly rounded up in CPython.

That doesn't mean anything except that the guy who wrote it for CPython decided to use it. Maybe next guy will submit a bug "select.poll() overdues the delay requested, caused my robot fall on the floor" and switch it to something else.

Anyway, let me merge this.

pfalcon:
Merged thanks! (Ah, and the content which is in the description (which I managed to miss, I'd even say that it originally wasn't there, but was edited later, but github doesn't mark it as such, so apparently it's my mistake) could go into the commit message as the delayed motivation - thanks for that).

PR #224, opened 2017-10-29
uasyncio submodule for synchronization primitives, Lock implemented.

pfalcon:
@dpgeorge , @peterhinch : Please reviews. Points to consider:

* Name of the submodule. I wished to use uasyncio.sync for qstr reuse, but that may be not clear enough.
* Correctness of implementation. Note that this is compatible with CPython Lock API, which asyncio liking to confusingly introduce non-coroutines methods here and there, .release() being such. So, that was worked around by putting extra yield into acquire(), and even works without scheduling artifacts with the provided example, though in general, it might lead to such.

### DIFF
--- a/uasyncio.synchro/example_lock.py
+++ b/uasyncio.synchro/example_lock.py
@@ -1,0 +1,27 @@
+try:
+    import uasyncio.core as asyncio
+    from uasyncio.synchro import Lock
+except ImportError:
+    import asyncio
+    from asyncio import Lock
+
+
+def task(i, lock):
+    print(lock)
+    while 1:
+        yield from lock.acquire()
+        print("Acquired lock in task", i)
+        yield from asyncio.sleep(0.5)
+#        yield lock.release()
+        lock.release()
+
+
+loop = asyncio.get_event_loop()
+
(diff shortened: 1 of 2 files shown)


peterhinch:
Looks good to me.

peterhinch:
Why not support use in an asynchronous context manager as per [this attempt](https://github.com/peterhinch/micropython-async/blob/fc7aca62e31adfea75541eaed57cf5ce1bbb4f23/asyn.py#L49-L75)?

pfalcon:
Because context manager is a syntactic sugar requiring more code, and uasyncio's goal is to provide required functionality with the minimum of code.

pfalcon:
Ok, I'm still not sure about "synchro" part, but there're more things in queue, so merging this. As I figure noone really tested this, I'm leaving debug statements in the code.

pfalcon:
To clarify, this was developed to have an example on non-I/O-waiting coroutine which is fully suspended from the main scheduling queue (to be able to test cancellation algorithm and make sure that the proposed solution, which required patching the language core, is fully general).

PR #225, opened 2017-10-31
Fix `urequests.request` failing with `OSError: 97`

SpotlightKid:
…  when `getaddrinfo` returns addresses in unexpected order.

`urequests.request` incorrectly assumes that `usocket.getaddrinfo` always returns the addrinfo for IPv4/TCP as the first item and passes the address from this to `socket.connect`. This will fail with `OSError: 97` if the the first item returned by `getaddrinfo` is actually a IPv6 address.

Example:

```
>>> import usocket
>>> import urequests
>>> [ai[:3] for ai in usocket.getaddrinfo('jsonplaceholder.typicode.com', 80)]
[
    (10, 1, 6),
    (10, 2, 17),
    (10, 3, 0),
    (10, 1, 6),
    (10, 2, 17),
    (10, 3, 0),
    (2, 1, 6),
    (2, 2, 17),
    (2, 3, 0),
    (2, 1, 6),
    (2, 2, 17),
    (2, 3, 0)
]
>>> urequests.get('http://jsonplaceholder.typicode.com/posts/1')
  File "/usr/lib/micropython/urequests.py", line 108, in get
  File "/usr/lib/micropython/urequests.py", line 96, in request
  File "/usr/lib/micropython/urequests.py", line 58, in request
OSError: 97
```

This PR fixes that by iterating over the tuple of addrinfos returned by `getaddrinfo` and picking the addr from the first one, where the first three items match `(usocket.AF_INT, usocket.SOCK_STREAM, usocket.IPPROTO_TCP)`. If no matching address is found, `OSError` is raised (to be be consistent with what `getaddrinfo` raises).

Since `usocket.IPPROTO_TCP` is not available on every port, the module defines its own constant `IPPROTO_TCP`. It might be sufficient to check only the first two items of each addrinfo, though, and leave this out.

The error number and message of the `OSError` exception are only a (sensible) suggestion.

Signed-off-by: Christopher Arndt <chris@chrisarndt.de>

### DIFF
--- a/urequests/urequests.py
+++ b/urequests/urequests.py
@@ -1,5 +1,13 @@
 import usocket
 
+
+try:
+    IPPROTO_TCP = usocket.IPPROTO_TCP
+except AttributeError:
+    from micropython import const
+    IPPROTO_TCP = const(6)
+
+
 class Response:
 
     def __init__(self, f):
@@ -50,8 +58,12 @@ def request(method, url, data=None, json=None, headers={}, stream=None):
         host, port = host.split(":", 1)
         port = int(port)
 
-    ai = usocket.getaddrinfo(host, port)
-    addr = ai[0][-1]
+    for ai in usocket.getaddrinfo(host, port):
+        if ai[:3] == (usocket.AF_INET, usocket.SOCK_STREAM, IPPROTO_TCP):
+            addr = ai[-1]
+            break
+    else:
+        raise OSError(-2, "No supported address found.")
 
     s = usocket.socket()
     try:


pfalcon:
Yes, this issue needs to be solved, but solved by actually implementing more args to getaddrinfo() for all ports...

SpotlightKid:
Until that time, it remains that `urequests.request` a) will fail unpredictably for reasons not under control of the user where it shouldn't and b) there's no proper way for the user to recover from the error. And thus makes it rather useless in practice.

Do you have a suggestion how to fix this in better way until `usocket.getaddrinfo` is fixed everywhere?

Some ideas:

* Require `micropython-socket` and provide a better implementation of `getaddrinfo` there.
* Try to pass the address family (and maybe socket type) in `urequests` but fall back to two arguments, if a `TYpeError` is thrown.
* Add some code `urequests`, which checks whether `usocket.getaddr` supports a third (fourth, fifth) argument 

Options 2) and 3) would still leave `urequests` in a state of questionably usefulness on ports where `geaddrinfo` accepts only two arguments.

pfalcon:
> will fail unpredictably for reasons not under control of the user where it shouldn't

I'm not sure that's the proven case in the general, before it's fully understood why it happens to some users. This is definitely not the 1st report of that (though maybe the previous was from you too?). And you don't specify what system it happens on and any background info.


And anyway, why look for workarounds instead of working towards the proper fix?

urequests/urequests.py
```diff
@@ -50,8 +58,12 @@ def request(method, url, data=None, json=None, headers={}, stream=None):
         host, port = host.split(":", 1)
         port = int(port)
 
-    ai = usocket.getaddrinfo(host, port)
-    addr = ai[0][-1]
+    for ai in usocket.getaddrinfo(host, port):
+        if ai[:3] == (usocket.AF_INET, usocket.SOCK_STREAM, IPPROTO_TCP):
```
pfalcon: This allocates memory, and that should be avoided.

pfalcon:
> OSError: 97

That's EAFNOSUPPORT

> (10, 1, 6),

And that's IPv6, SOCK_STREAM, IPPROTO_TCP.

So, the resolver returns IPv6 address as the first choice, then when it comes to connecting to it, the system says IPv6 is not supported. Seems like system (or resolver) misconfiguration to me.

urequest won't force IPv4 on its users - there's nothing wrong with connecting over IPv6, and indeed, IPv4 may be not even available.

SpotlightKid:
Ok, now I get it: the actual error lies in the way the socket is created on line 56 of `urequests.py`. The socket should be configured for the correct address family.

I'll close this PR and hopefully come up with a better one.

PR #226, opened 2017-10-31
urequests: Fix OSError when connecting to IPv6 adress.

SpotlightKid:
If the first address returned by `getaddrinfo()` is a IPv6 address, `usocket.socket()` must be called with the correct address family and socket type, otherwise `connect()` on the socket will fail with `OSError: 97` (`EAFNOSUPPORT`).


SpotlightKid:
Question is: are the `af` and `type` arguments to `usocket.socket()` supported on every port?

SpotlightKid:
Fixed by another commit: a2647e316eee02d87e53adee60b21797b3272992,

PR #227, opened 2017-11-03
uasyncio: StreamReader: Separate "poll socket" vs "I/O socket".

pfalcon:
Poll socket is what's passed to uselect.poll(), while I/O socket is what's
used for .read(). This is a workaround of the issue that MicroPython doesn't
support proxying poll functionality for stream wrappers (like SSL, websocket,
etc.)
This issue is tracked as https://github.com/micropython/micropython/issues/3394

It may be that it's more efficient to apply such a workaround on uasyncio
level rather than implementing full solution of uPy side.


pfalcon:
Merged.

Issue #148, opened 2017-02-05
Merging uaiohttpclient


pfalcon:
I looked thru my repos and saw https://github.com/pfalcon/micropython-uaiohttpclient which I almost forgot about. So, I propose to merge it into micropython-lib. It's also the easiest way to achieve its installability on esp8266 (gzip-4k packing).

@dpgeorge ^

dpgeorge:
I saw there is a asyncio version of requests https://github.com/rdbhost/yieldfromRequests, and it would be nice to support/promote a consistent set of libraries, eg both requests and it's yield-from version.  OTOH, aiohttp looks better maintained than yieldfromRequests, and has more github stars, so perhaps aiohttp is the one we should use.

pfalcon:
Imported.

PR #228, opened 2017-11-03
WIP: uasyncio.websocket: Websocket implementation for uasyncio.

pfalcon:
Q #1: Should this be in uasyncio package at all? Upstream doesn't have
this. Pro: will be easier for people do discover (see e.g.
#148)

Q #2: This provides implements 2 ways to create a WS connections:
1) using start_ws_server(); 2) using wrapping existing StreamReader
and StreamWriter. History: initial prototype of course used 2). But
the idea was "it should be like the official start_server()!!1". But
then I though how to integrate it e.g. with Picoweb, and became clear
that 2) is the most flixble way. So, 1) is intended to be removed.

Q #3: Uses native websocket module for read path, but has own
write path due to micropython/micropython#3396

Q #4: Requires #227
due to micropython/micropython#3394 .

### DIFF
--- a/uasyncio.websocket/example_websock.py
+++ b/uasyncio.websocket/example_websock.py
@@ -1,0 +1,28 @@
+import uasyncio
+from uasyncio_.websocket import WSReader, WSWriter
+
+
+def echo(reader, writer):
+    # Consume GET line
+    yield from reader.readline()
+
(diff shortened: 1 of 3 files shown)


pfalcon:
@dpgeorge , @peterhinch : FYI

pfalcon:
Merged.

PR #229, opened 2017-11-05
gettext: implement gettext and ngettext

xrmx:
By wrapping the C counterparts

### DIFF
--- a/gettext/gettext.py
+++ b/gettext/gettext.py
@@ -1,0 +1,14 @@
+import ffilib
+
+libc = ffilib.libc()
+
+gettext_ = libc.func("s", "gettext", "s")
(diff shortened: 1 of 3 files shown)


pfalcon:
What's the motivation here?

xrmx:
Being able to run software that use gettext :)

pfalcon:
But not on a baremetal target which doesn't have libc and gettext? I'd like to be sure that contributors understand this, and that another contributor may propose to switch this implementation to a different one. With that in mind, it's ok.

xrmx:
Sure, on the unix port of course, not on micro controllers.

pfalcon:
Merged, thanks!

Issue #235, opened 2017-11-22
Hashlib questions (use of uhashlib, native code otherwise present, ripemd160)


doc-hex:
I have some questions and comments about the `hashlib` module.

1. Some boards/ports have `sha256`, `sha1` and others, as part of the binary already (particularly when SSL is being used). Some of these are made available in `uhashlib`, and so IMHO `hashlib` should use those versions since they are both faster and smaller than anything we can do in pure python.  It looks to me like `sha256` is guaranteed to be there (if `uhashlib` exists) and SHA1 may exist if `MICROPY_PY_UHASHLIB_SHA1` is set.

2. It pains me that `lib/axtls/crypto` contains lots of hash functions that could be part of `uhashlib`, plus AES and other primitives. That belongs as an issue against [axtls](https://github.com/pfalcon/axtls) though.

3. `hashlib.new()` seems like a useful addition to implement, since that would allow pruning of algorithms that aren't needed on today's project, and if this module is rewritten in C, it becomes a great place to extend with native versions, when available.

4. I need to have `ripemd160` but I don't think it's common enough to bother weighing-down this module for everyone.

I'm going to follow-up with pull requests for 1 and 3, and maybe 4  if there is interest.

Any thoughts?

PR #236, opened 2017-11-23
Add hashlib.new(), detect and use uhashlib native code versions when possible

peter-conalgo:
See discussion in #235 

- adds `hashlib.new()` which looks up any hash function by name
- uses `uhashlib` version of each hash function if it is defined there
- exports `sha1` if it's in uhashlib (no pure implementation)
- any future hash function can be added to `uhashlib` and will be available via `hashlib.new()`

### DIFF
--- a/hashlib/hashlib/__init__.py
+++ b/hashlib/hashlib/__init__.py
@@ -1,2 +1,41 @@
-from .sha256 import sha224, sha256
-from .sha512 import sha384, sha512
+try:
+    import uhashlibxx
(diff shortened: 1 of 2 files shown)


peter-conalgo:
I didn't change the version number because it looked kinda complex... sorry.

pfalcon:
Redone somewhat differently in 175634b3cdf3a9269f5f6f9f8e76dbd524e24c44. Thanks for giving it a try.

PR #237, opened 2017-11-23
Supported socket timeout

Himura2la:
I really need timeout, please merge.

I'm not sure that the behavior does not change, but maybe it's a good idea to set the default timeout to a kind of 30 seconds

### DIFF
--- a/umqtt.simple/umqtt/simple.py
+++ b/umqtt.simple/umqtt/simple.py
@@ -52,8 +52,9 @@ class MQTTClient:
         self.lw_qos = qos
         self.lw_retain = retain
 
-    def connect(self, clean_session=True):
+    def connect(self, clean_session=True, timeout=None):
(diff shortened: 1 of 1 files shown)


Himura2la:
@pfalcon ping

dpgeorge:
Not all MicroPython ports support socket.settimeout (notably the unix port).  So this won't work.

Himura2la:
Maybe we can reflect this in docs/comments? I'll change the code so that the .settimeout invokes only if the value is passed. It's better to allow this at least on some platforms rather than force the lib's modifications, isn't it?

Himura2la:
@dpgeorge  Hey, what's the status

NJannasch:
push

dpgeorge:
Timeout was added in d6faaf847243766c1bca92d8bca603025d3f19ba

Issue #238, opened 2017-11-30
urequests does not set Content-type to application/json when json parameter is not None


nucflash:
`urequests` does not set `Content-Type: application/json` when the `json` parameter is set. This makes a receiving flask server unable to parse the request and return `None` for `request.get_json()`

nucflash:
Duplicate of #239. Closing.

PR #239, opened 2017-11-30
urequests: set Content-Type to application/json when the json parameter is set

nucflash:
Sets the Content-Type to application/json when the json parameter is set. Otherwise, a receiving flask server is unable to parse the json payload. Relates to issue #238


pfalcon:
Updated commit message and merged in https://github.com/micropython/micropython-lib/commit/6c141990ac7a817ed6ca29ffb6e308ff664dbf0c . Thanks!

Issue #198, opened 2017-07-13
Request for package: micropython-ssl


alex-robbins:
Would you be open to adding a `micropython-ssl` package that wraps `ussl`, similarly to how `micropython-socket` wraps `usocket`? I can think of two use cases for this, as discussed below. I'd be happy to contribute the code for this, but I want to make sure it's something you'd be receptive to first, especially considering that it would be a new package.

Use case 1: Drop default arguments
--------------
According to the docs for most (or maybe all) ports, the signature for `wrap_socket` is ` ssl.wrap_socket(sock, server_side=False, keyfile=None, certfile=None, cert_reqs=CERT_NONE, ca_certs=None)`, with the caveat that
> Depending on the underlying module implementation for a particular board, some or all keyword arguments above may be not supported.

Indeed, most ports (it seems) do not support nearly so many of these. On Unix, for example, even the `keyfile` and `certfile` arguments are not accepted.

Note though, that even `ussl.wrap_socket(s, keyfile=None)` does not work, even though it should be equivalent to `ussl.wrap_socket(s)`. The proposed `micropython-ssl` package could provide an `ssl.wrap_socket` function that drops keyword arguments whose values are the default (i.e. `ssl.wrap_socket(s, keyfile=None)` results in a call to `ussl.wrap_socket(s)`).

This would be useful for user-defined libraries that accept a `keyfile` parameter from their callers and pass this to `wrap_socket`. (It would provide better compatibility across MicroPython ports and with CPython, without growing MicroPython proper.)

Use case 2: Allow SSL-wrapping of `micropython-socket` sockets
------------
Due to micropython/micropython#2749, and the fact that `socket.socket` is a subclass of `usocket.socket`, `micropython-socket` sockets cannot be passed to `ussl.wrap_socket` (attempts to do so cause a crash, for now). This makes the `micropython-socket` package almost entirely unusable with SSL.

If `micropython-socket` were modified, though, to wrap instances of `usocket.socket`, rather than subclassing it, then the proposed `micropython-ssl` package could provide a `wrap_socket` function that retrieves the underlying `usocket.socket` instance from any passed-in `socket.socket`s before forwarding to `ussl`.

[...]

PR #246, opened 2018-01-04
ssl: Drop default arguments to wrap_ssl

alex-robbins:
This is to take care of what I described as "Use case 1" in reporting #198. (Use case 2 needs a separate PR, so I'll leave #198 open for now).

I know `wrap_socket` looks a little repetitive (because it is), but I tried two other implementations, and (especially [without proper support] for `locals()` in MicroPython) it was always significantly more convoluted and harder to read.

[without proper support]: http://docs.micropython.org/en/latest/pyboard/genrst/core_language.html#local-variables-aren-t-included-in-locals-result

### DIFF
--- a/ssl/ssl.py
+++ b/ssl/ssl.py
@@ -1,1 +1,6 @@
 from ussl import *
+
+# Constants
+for sym in "CERT_NONE", "CERT_OPTIONAL", "CERT_REQUIRED":
+    if sym not in globals():
+        globals()[sym] = object()
--- a/ssl/ssl.py
+++ b/ssl/ssl.py
@@ -1,6 +1,28 @@
 from ussl import *
+import ussl as _ussl
 
 # Constants
 for sym in "CERT_NONE", "CERT_OPTIONAL", "CERT_REQUIRED":
     if sym not in globals():
         globals()[sym] = object()
+
+
+def wrap_socket(sock, keyfile=None, certfile=None, server_side=False,
+                cert_reqs=CERT_NONE, *, ca_certs=None, server_hostname=None):
+    # TODO: More arguments accepted by CPython could also be handled here.
+    # That would allow us to accept ca_certs as a positional argument, which
+    # we should.
+    kw = {}
+    if keyfile is not None:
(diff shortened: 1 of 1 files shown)


ssl/ssl.py
```diff
@@ -1,6 +1,28 @@
...
+
+
+def wrap_socket(sock, keyfile=None, certfile=None, server_side=False,
+                cert_reqs=CERT_NONE, *, ca_certs=None, server_hostname=None):
+    # TODO: More arguments accepted by CPython could also be handled here.
+    # That would allow us to accept ca_certs as a positional argument, which
+    # we should.
+    kw = {}
+    if keyfile is not None:
+        kw['keyfile'] = keyfile
```
pfalcon: FYI, micropython-lib's preferences is to use "strings".

pfalcon:
Can you please also update the PR title to correspond to its content?

alex-robbins:
> Can you please also update the PR title to correspond to its content?

Done, sorry.

> use "strings".

Are you saying to use double quotes instead of single quotes (`"keyfile"` instead of `'keyfile'`)? 

> over-dynamicity like this

So should I just unroll the loop and use a series of if statements? Like
```python
if 'CERT_NONE' not in globals():
    CERT_NONE = object()


if 'CERT_OPTIONAL' ...
```

pfalcon:
> Are you saying to use double quotes instead of single quotes ("keyfile" instead of 'keyfile')?

Yes, I'm saying that majority of the code written from scratch for micropython-lib uses double-quotes, so for example someone may rely that searching for `"foo"` finds all instances of it.

> So should I just unroll the loop and use a series of if statements?

D'oh, sorry, I missed that this magic is to avoid overriding anything which could have been imported from ussl.

Ok, then let me know if you're going to change strings and this is good to go.

alex-robbins:
I usually use single quotes in python, but as PEP 8 says, "consistency within a project is more important".

Force push to change quote characters (also rebased onto master).

pfalcon:
Merged, thanks.

PR #255, opened 2018-01-23
argparse: Implement parse_known_args

stinos:
This is convenient when components need only to parse a subset of an
application's arguments, and can be implemented with minor changes to
_parse_args: basically just add unknown arguments to a list instead of
raising an exception.


pfalcon:
I assume this corresponds to the CPython behavior. Merged, thanks.

PR #265, opened 2018-03-23
add configparser module

Mika64:
Configparser module
TODO : add comments support

### DIFF
--- a/configparser/ConfigParser.py
+++ b/configparser/ConfigParser.py
@@ -1,0 +1,127 @@
+"""
+Minimal and functional version of CPython's ConfigParser module.
+"""
+
+class ConfigParser:
+    def __init__(self):
+        self.config_dict = {}
+
+    def sections(self):
+        """Return a list of section names, excluding [DEFAULT]"""
+        to_return = [section for section in self.config_dict.keys() if not section in "DEFAULT"]
+        return to_return
+
+    def add_section(self, section):
+        """Create a new section in the configuration."""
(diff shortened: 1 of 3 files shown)


thijstriemstra:
+1!

Zitt:
+1

Zitt:
I have a uefi script which uses this ... can we get this plus +2ed and merged into the master? @dpgeorge

thijstriemstra:
i used this PR 3 years ago and since moved away from micropython; prs are unfortunately not reviewed and merged.

Zitt:
> i used this PR 3 years ago and since moved away from micropython; prs are unfortunately not reviewed and merged.
@thijstriemstra 
Is there a better micropython-lib implementation which is actually mantained?

dpgeorge:
Thanks for the PR and sorry this never got any attention.

In its current form this PR cannot be merged, for at least the following reasons:
- the module file is named incorrectly (you need to be able to do `from configparser import ConfigParser`)
- methods have different signature to CPython (eg `read` should take an iterable of filenames, `write` should not take a filename)
- errors use `print(msg); raise` which is very non-standard

Also, it would need to be updated to the latest repo conventions.

So I'll close this for now.  It's a nice contribution and we would merge it if the above were addressed, so feel free to reopen with changes, or someone else submit an improved version).

thijstriemstra:
glad to see this getting some attention after 8 years.......

PR #266, opened 2018-04-02
uasyncio: close server socket when the generator exits

larsks:
uasyncio.start_server creates a socket but does not close it upon
generator exit.  Since the socket isn't otherwise exposed to the
calling code, this means it's not possible to restart the server
without resetting the board.

This commit catches the GeneratorExit exception and closes the socket.

### DIFF
--- a/uasyncio/uasyncio/__init__.py
+++ b/uasyncio/uasyncio/__init__.py
@@ -240,18 +240,22 @@ def start_server(client_coro, host, port, backlog=10):
     s.setsockopt(_socket.SOL_SOCKET, _socket.SO_REUSEADDR, 1)
     s.bind(ai[-1])
     s.listen(backlog)
-    while True:
-        if DEBUG and __debug__:
-            log.debug("start_server: Before accept")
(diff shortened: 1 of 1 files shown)


larsks:
I don't actually know the answer to that question. I'm happy to make the change.

peterhinch:
I'm merely a user of uasyncio, not a maintainer.

The canonical way to stop a coroutine is by means of a **CancelledError** (which is also used in the timeout mechanism). I'm unsure of the implications of sending **close()** to a running coroutine. Further [this reference](https://stackoverflow.com/questions/30862196/generatorexit-in-python-generator) suggests that trapping **GeneratorExit** may be inadvisable.

Perhaps the best approach is a try-finally block? Something along the lines of
```python
    # code
    try:
        while True:
            # code
    finally:
        s.close()
        raise GeneratorExit  # Ensure coro is descheduled
````
A solution needs to be tested against task cancellation and timeouts in my opinion, and it may be advisable to use cancellation rather than **close()** in your application.

PR #270, opened 2018-04-04
uasyncio.core: run_until_complete should return coroutine's result

ciniml:
According to [CPython reference](https://docs.python.org/3.6/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.run_until_complete), run_until_complete method of EventLoop returns Future's result or raise exception occured inside coroutine, but current run_until_complete method in uasyncio does not return anything.

Since there are no Future object in uasyncio, it's difficult to propagate exception occured inside coroutne. But it's not difficult to return result from coroutine.

This modification fixes the run_until_complete method to return coroutine's result.

### DIFF
--- a/uasyncio.core/test_coro_result.py
+++ b/uasyncio.core/test_coro_result.py
@@ -1,0 +1,26 @@
+try:
+    import uasyncio.core as asyncio
+except ImportError:
+    import asyncio
+import logging
+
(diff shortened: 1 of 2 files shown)


jimmo:
This was fixed in asyncio v3, and confirmed that the included test in this PR passes.

PR #271, opened 2018-04-09
Add the ability to connect if IP is used.

goyzhang:
When I use the mqtt module I found an OSerror(no entry) will be thrown if I use my mqtt broker's IP for creating a new mqtt instance.
In some limited cases, like in a factory using LAN instead of internet connection, an local lightweight mqtt broker should be allowed to use IP address. I added a simple regex function to check if it is a valid IP. If not, then everything is the same as usual. See if this help.

### DIFF
--- a/umqtt.simple/umqtt/simple.py
+++ b/umqtt.simple/umqtt/simple.py
@@ -54,7 +54,11 @@ class MQTTClient:
 
     def connect(self, clean_session=True):
         self.sock = socket.socket()
-        addr = socket.getaddrinfo(self.server, self.port)[0][-1]
+        addr=None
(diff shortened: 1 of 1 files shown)


dpgeorge:
> an local lightweight mqtt broker should be allowed to use IP address.

It should already be able to do this.  You should be able to pass an IP address as a string like "192.168.0.1" and it will work.  Some of the examples for umqtt do use this approach.

> I found an OSerror(no entry) will be thrown if I use my mqtt broker's IP

Then something else is wrong, maybe with the MicroPython port/board you use and the implementation of `socket.getaddrinfo()`.

PR #275, opened 2018-04-17
Updates for new PyPI

### LINKED ISSUE #274: pypi redirects to https://pypi.org/pypi/%s/json and brakes upip
Changing URL to "https://pypi.org/pypi/%s/json" appears to fix but I can't test now yet because I don't have the build process setup yet.

A put modified upip.py inside the flash memory but my ESP8266 board do not have enough ram to run upip.install

```
>>> upip.install('micropython-urequests')
Installing to: /lib/
Warning: pypi.python.org SSL certificate is not validated
Error installing 'micropython-urequests': b'302', packages may be partially installed
```

di:
Hi! These changes should make `upip` compatible with the new PyPI at https://pypi.org.

I also updated a few old links in `README.md` which were pointing to `pypi.python.org`.

Fixes #274.

### DIFF
--- a/upip/bootstrap_upip.sh
+++ b/upip/bootstrap_upip.sh
@@ -9,7 +9,7 @@ fi
 
 # Remove any stale old version
 rm -rf micropython-upip-*
-wget -nd -r -l1 https://pypi.python.org/pypi/micropython-upip/ --accept-regex ".*pypi.python.org/packages/source/.*.gz" --reject=html
+wget -nd -rH -l1 -D files.pythonhosted.org https://pypi.org/project/micropython-upip/ --reject=html
 
 tar xfz micropython-upip-*.tar.gz
 mkdir -p ~/.micropython/lib/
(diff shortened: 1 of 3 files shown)


upip/bootstrap_upip.sh
```diff
@@ -9,7 +9,7 @@ fi
 
 # Remove any stale old version
 rm -rf micropython-upip-*
-wget -nd -r -l1 https://pypi.python.org/pypi/micropython-upip/ --accept-regex ".*pypi.python.org/packages/source/.*.gz" --reject=html
+wget -nd -rH -l1 -D files.pythonhosted.org https://pypi.org/project/micropython-upip/ --reject=html
```
dpgeorge: Does it need the pythonhosted.org bit?

di: Yes, we only want to download links with the `files.pythonhosted.org` host because that's where the distributions  live (hence the `--accept-regex` no longer being necessary)

dpgeorge:
Thanks @di for this fix, much appreciated!  I verified it works and merged your fixes with modified commit messages in 49e8a839a003d0e2ebf3eee4204a5be9536036d6, b4c27ea1ea362ce5480de94c21f9cfe44bfb2f6f and e17efb08440846485f13d7790a9e45f1401bd5ff

PR #276, opened 2018-04-28
Handle case when server doesn't return anything

davispuh:
When server doesn't return anything
```
  File "urequests.py", line 100, in request
  File "urequests.py", line 84, in request
IndexError: list index out of range
```

### DIFF
--- a/urequests/urequests.py
+++ b/urequests/urequests.py
@@ -81,6 +81,8 @@ def request(method, url, data=None, json=None, headers={}, stream=None):
         l = s.readline()
         #print(l)
         l = l.split(None, 2)
+        if len(l) <= 1:
+            return Response(s)
         status = int(l[1])
         reason = ""
         if len(l) > 2:


andrewleech:
Hello I know this was a long time ago, but I'm trying to collate some urequests updates now. Do you remember, is there any valid case where this should happen? 
I see your change takes affect if there isn't even a status code in the response which appears to be completely invalid http, a quick test shows that cpython requests library gives a long list of exception messages in this case, I don't think we should simply return an empty reponse instead.
I've added a new commit to my working branch inspired by yours here, however I raise a `ValueError` with an error message instead. Does this sound appropriate to you?

davispuh:
It's been way too long so I really don't remember :D but I think I encountered some API that sometimes had issues so it would immediately close connection and you wouldn't get any data. Because I don't even remember what project it was this seems to not matter anymore and I think your approach should be fine.

Generally even with broken HTTP servers sometimes you still might want to get whatever they returned.

andrewleech:
Thanks for the reply, yep that sounds reasonable. I updated my commit slightly to still raise the exception, but then also print out whatever was received as well.

andrewleech:
dbd8fff8 was merged as part of #500 which raises a ValueError as suggested above with a message that should make it clear the error that's occurred. Thanks again!

PR #277, opened 2018-05-06
logging: support for file based logging

opreab:
Added basic file logging support
Added unittest for basicConfig when using files
Added stress test for logging to file and std stream

### DIFF
--- a/logging/example_logging.py
+++ b/logging/example_logging.py
@@ -1,6 +1,38 @@
+import os
+import time
 import logging
+
+
+def reset_log(log_file):
+    try:
+        os.remove(log_file)
+    except OSError:
+        pass
+
+
+def stress_test(iterations):
+    sample = "All work and no play, makes Jack a dull boy!"
+    log_file = 'file.log'
+    reset_log(log_file)
+    logging.basicConfig(filename=log_file)
+    start_time = time.time()
+    for i in range(iterations):
+        logging.info("%d %s" % (i, sample))
+    file_delta = time.time() - start_time
+    reset_log(log_file)
+
+    logging.basicConfig(filename=None)
+    start_time = time.time()
+    for i in range(iterations):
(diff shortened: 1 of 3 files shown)


logging/example_logging.py
```diff
@@ -1,9 +1,40 @@
+import os
+import time
 import logging
 
+
+def reset_log(log_file):
+    try:
+        os.remove(log_file)
+    except OSError:
+        pass
```
tomasz2101: Is it good approach to silently pass error?

logging/example_logging.py
```diff
@@ -1,9 +1,40 @@
...
+    file_delta = time.time() - start_time
+    reset_log(log_file)
+
+    logging.basicConfig(filename=None)
+    start_time = time.time()
+    for i in range(iterations):
+        logging.info("%d %s" % (i, sample))
+    stdout_delta = time.time() - start_time
+    print("File logging time %f for %i iterations" % (file_delta, iterations))
+    print("Stdout logging time %f for %i iterations" % (stdout_delta, iterations))
```
tomasz2101: Why do you want to put this function in example ?

logging/logging.py
```diff
@@ -38,11 +38,16 @@ def isEnabledFor(self, level):
...
+                _log_file.write(("%s:%s:" + msg + "\n") % ((self._level_str(level), self.name) + args))
+                _log_file.flush()
             else:
-                print(msg % args, file=_stream)
+                _stream.write("%s:%s:" % (self._level_str(level), self.name))
+                if not args:
+                    print(msg, file=_stream)
+                else:
+                    print(msg % args, file=_stream)
+
```
tomasz2101: Can we write to stream and file in same time?

jimmo:
File based logging was added in #507.

PR #278, opened 2018-05-07
urequests: Add iteration methods to process large responses and support 'with' to close response

chrisb2:
This change adds the following methods from the [Response class of the Python requests](https://github.com/requests/requests/blob/239ee194318e0b88d6a7b292ec4b68451ceefe68/requests/models.py#L578) module to the urequests Response class:
- iter_lines()
- iter_content()
- \_\_enter__()
- \_\_exit__()
- \_\_iter__()

This enables two things:
* processing of large HTTP responses
* succinct handling of the need to close Responses. 

So for example:
```
# Iterate over individual lines of the response
with urequests.get('http://jsonplaceholder.typicode.com/users') as response:
    for line in response.iter_lines():
        print(line.decode(response.encoding))
		
# Iterate over 'chunks' of the response
with urequests.get('http://jsonplaceholder.typicode.com/users') as response:
    for chunk in response.iter_content():
        print(chunk.decode(response.encoding))
```
The \_\_iter__() method of Response allows this example:
```
for chunk in requests.get('http://jsonplaceholder.typicode.com/users'):
    print(chunk.decode('UTF-8'))
```
I chose to reduce the default size of ITER_CHUNK_SIZE from 512 to 128 to conserve resources, not sure this makes sense?

I have tested this new functionality on ESP8266 and ESP32.

Finally, this obviously increases code size, so maybe the scope can be reduced? The \_\_iter__() method could be dropped as this would not reduce the actual capability. Beyond that we could maybe drop the iter_lines() method? Both \_\_enter__() and \_\_exit__() are small and get rid of nasty try finally syntax so I think these should stay?

### DIFF
--- a/urequests/example_iter.py
+++ b/urequests/example_iter.py
@@ -1,0 +1,15 @@
+# Example of processing large responses, the response is automatically closed
+try:
+    import urequests as requests
+except ImportError:
+    import requests
+
+# Iterate over individual lines of the response
+with requests.get('http://jsonplaceholder.typicode.com/users') as response:
+    for line in response.iter_lines():
+	    print(line.decode(response.encoding))
+		
+# Iterate over 'chunks' of the response
+with requests.get('http://jsonplaceholder.typicode.com/users') as response:
+    for chunk in response.iter_content():
+	    print(chunk.decode(response.encoding))
--- a/urequests/metadata.txt
+++ b/urequests/metadata.txt
@@ -1,4 +1,4 @@
 srctype = micropython-lib
 type = module
-version = 0.6
+version = 0.7
 author = Paul Sokolovsky
--- a/urequests/setup.py
+++ b/urequests/setup.py
@@ -7,7 +7,7 @@ sys.path.append("..")
 import sdist_upip
 
 setup(name='micropython-urequests',
-      version='0.6',
+      version='0.7',
       description='urequests module for MicroPython',
       long_description="This is a module reimplemented specifically for MicroPython standard library,\nwith efficient and lean design in mind. Note that this module is likely work\nin progress and likely supports just a subset of CPython's corresponding\nmodule. Please help with the development if you are interested in this\nmodule.",
       url='https://github.com/micropython/micropython-lib',
--- a/urequests/urequests.py
+++ b/urequests/urequests.py
@@ -1,11 +1,23 @@
 import usocket
 
+ITER_CHUNK_SIZE = 128
+
 class Response:
 
     def __init__(self, f):
         self.raw = f
         self.encoding = "utf-8"
+        self._content_consumed = False
         self._cached = None
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
+    def __iter__(self):
+        return self.iter_content()
 
     def close(self):
         if self.raw:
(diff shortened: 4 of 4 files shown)


rdagger:
This feature looks very helpful.  I'm having memory errors with urequests.  Are there plans to merge?

ehagerty:
@chrisb2 just a thank you for the code here, it not only works but it does sort of blow one's mind that in 2021 micropython doesn't bake this in.

NoumanSaleem:
Looks great! @chrisb2 would you consider publishing this under a different name? There are at least three variations of the urequests package living in pip already (https://pypi.org/search/?q=urequests)
Possibly: `chrisb2-urequests`

davidpcahill (approved):
Please merge / add to latest!

weeenerplastics:
please merge, now using https://github.com/SpotlightKid/mrequests

Gadgetoid:
I was about to raise a PR with just the context manager portion:

```python
    def __enter__(self):
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.close()
```

MicroPython typically has a hidden five socket "hard" limit (set by lwIPs TCP socket memory pool) which can be raised with `MEMP_NUM_TCP_PCB`, but more often than not leads to users (usually prodding `response.raw` because they want to stream a file to flash or otherwise) encountering cryptic `ENOMEM` errors on a board that clearly still has RAM left, simply because they weren't diligent about calling `response.close()`. It's also not obvious that a `close` happens implicitly when calling the `json()` and `text()` methods (a pattern than makes thoroughly no conceptual sense to anyone who hasn't studiously read the docs or code).

This is Pythonic, and sensible:

```python
with requests.get("https://example.com") as response:
    # do something with response.raw
```

TLDR: 👍  (for the context manager change)



Edit: To level a critique at the rest of this PR, it's probably what hung up the context manager change (which IMO should have been included a long, long time ago) the following is an anti-pattern for large file iteration since it will alloc "chunk" every iteration:

```python
    def iter_content(self, chunk_size=ITER_CHUNK_SIZE):
        def generate():
            while True:
                chunk = self.raw.read(chunk_size)
                if not chunk:
                    break
                yield chunk
```

It should ideally use a fixed size chunk buffer and `readinto` along these lines:

```python
    def iter_content(self, buffer=None):
        buffer = buffer or bytearray(ITER_CHUNK_SIZE)
        def generate():
            while True:
                length = self.raw.readinto(buffer)
                if length == 0:
                    break
                yield buffer, length
```

Which you'd consume with `memoryview(buffer)[:length]` to avoid a buffer copy.

If you're streaming a large file to disk you will eventually have a bad day not re-using the same chunk of RAM.

Unfortunately the fact is that *how* you stream data varies widely based on what you're doing with the stream, I don't think there's a single answer that suits all cases so this kind of stuff probably isn't worth putting directly into `requests`.

In almost all cases you'd be better doing your own loop, it's not a lot of code:

```python
buffer = bytearray(1024)
with requests.get("some-resource") as response:
    while length := response.raw.readinto(buffer)
        output.write(memoryview(buffer)[:length])
```

dpgeorge:
As @Gadgetoid demonstrates above, there are easy ways to do this in the user/application code which would better suit how the data is consumed.

So I'll close this PR (context manager was merged in acf98ad2c0c9727cae06c6748d0c30da5d4b7682).

Feel free to open a PR (or reopen this one) with just an example added on how to do streaming.

PR #280, opened 2018-05-14
unittest: fix context manager to store exception for later retrieval

oliver-joos:
The statement "with assertRaises(errtype) as ctxt" checks the type of a raised exception, but did not store the exception into ctxt like unittest of CPython. The exception instance  is usually used to check its message or other args.

This commit is a one-line fix at the cost of a bit more RAM usage while running unittests.


dpgeorge:
Thanks for the contribution.  Rebased and merged in 9bc0b15f11cb5089ac4b8666fc9a8d181b11f853

PR #282, opened 2018-05-24
Add uuid4 implementation

dhylands:


dpgeorge:
Sorry this didn't get any attention.

This would be nice to have, but the version contributed at https://github.com/pfalcon/pycopy-lib/commit/dcfa34b6c842385b59ca3f0be2c759d755e01970 looks better: it masks the correct bits, is a bit more efficient because it stores the bytes directly, and has tests.

So I'll close this PR.  If someone would like to make a PR to bring in the above commit, feel free.

PR #288, opened 2018-06-20
uaiohttpclient: Allow non-standard port number

rtau:
Parse the port number in URL and use that port number (if specified) when creating connection.


dpgeorge:
Sorry we never got to this PR.  This ended up getting fixed in 149226d3f743e800dc6629c81c832b4a2164dd8f

PR #289, opened 2018-06-21
Handle timeouts in MQTTClient

devxpy:


dpgeorge:
Please see #237 for an alternative approach that uses `uselect.poll` and would be more portable.

devxpy:
Thanks for the update..

PR #292, opened 2018-07-01
uasyncio core.py Fix for micropython iss #3241.

peterhinch:
If `create_task` or `run_until_complete` is called with a generator function instead of a
generator (coro) the code silently fails to run. This patch issues an assertion failure
and so provides a traceback in the event of this common error. [Original issue.](https://github.com/micropython/micropython/issues/3241).

### DIFF
--- a/uasyncio.core/uasyncio/core.py
+++ b/uasyncio.core/uasyncio/core.py
@@ -4,6 +4,7 @@ import ucollections
 
 
 type_gen = type((lambda: (yield))())
+type_genf = type((lambda: (yield)))  # Type of a generator function
 
 DEBUG = 0
 log = None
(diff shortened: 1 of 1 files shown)


peterhinch:
Arguably this should test that the passed object is a coro/generator rather than not being a generator function. The reasoning was that `create_task` will work if passed a callable despite this being poor practice; I didn't want to break working code. Passing a generator function (an easy error to make) achieves nothing useful and results in a silent fail. Comments requested.

jimmo:
This was fixed in asyncio v3 (it's handled in `asyncio.create_task`, `asyncio.run`, and `loop.run_until_complete`).

PR #299, opened 2018-08-07
urequests: request function can take custom response class

SpotlightKid:
* Add `response_class` keyword arg to `urequests.request`,
  which defaults to `urequests.Response` and is used to
  instantiate the object handling the HTTP response.
* Call `set_header` method on instance of given response
  class for every received header line, which takes an
  unparsed header line as a byte string.
* Add no-op `set_header` method to `urequests.Response`.

This could be used to store the response headers in a custom response class as demonstrated in this [comment](https://github.com/micropython/micropython-lib/pull/217#issuecomment-411205264) for #217. It is the responsibility of the custom response class to properly parse and assemble headers (e.g. ones with continuation lines).

### DIFF
--- a/urequests/urequests.py
+++ b/urequests/urequests.py
@@ -32,7 +32,8 @@ class Response:
         return ujson.loads(self.content)
 
 
-def request(method, url, data=None, json=None, headers={}, stream=None):
+def request(method, url, data=None, json=None, headers={}, stream=None,
+            response_class=Response):
     try:
         proto, dummy, host, path = url.split("/", 3)
     except ValueError:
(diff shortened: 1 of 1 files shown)


SpotlightKid:
Simplified the patch by checking for the `set:header` method on the response object first (only once), which makes the dummy method on `urequests.Response` superfluous.

andrewleech:
Thanks to the dynamic nature of (micro)python the class replacement can already be done with normal monkey-patching, eg.

``` python
import urequests

class Response(urequests.Response)
    def set_header(self, line):
        pass # do stuff

urequests.Response = Response
```

The other change here to extend headers handling I think would be covered by 5854ae12 recently merged as part of #500.

Feel free to re-open / update the PR if I've missed something.

PR #302, opened 2018-08-26
umqtt.simple: allow no password in connect packet 

SpotlightKid:
The `umqtt.simple.MQTTClient` incorrectly assumes that if the user is set, the password is not `None`. If password is set to `None`, the password flag in the connect headers should not be set and no password should be sent to the server.

Further discussion here: https://forum.micropython.org/viewtopic.php?p=29636#p29598


SpotlightKid:
Closing this since I'm cleaning my unmerged PRs and this hasn't been acted upon in years.

PR #310, opened 2018-10-04
urequests: Added Basic Authentication to requests.

fschmi:


urequests/urequests.py
```diff
@@ -33,6 +33,9 @@ def json(self):
 
 
 def request(method, url, data=None, json=None, headers={}, stream=None):
+    if auth is not None:
+        headers.update(encode_basic_auth(auth))
```
SpotlightKid: * An `auth` parameter needs to be added to the `request` function signature (defaulting to `None`).

* If following my suggested changes for `encode_basic_auth`, this should be changed to the following then:

      if auth:
          headers.update(encode_basic_auth(auth[0], auth[1]))

  (Also changing `if auth is not None` to `if auth` so that passing an empty tuple works as well.)

SpotlightKid: It might be also a good idea to allow passing a callable for auth (but this would be an extension not supported by the original `requests.request()`.

    if auth:
        headers.update(auth() if callable(auth) else encode_basic_auth(auth[0], auth[1]))

fschmi: Obviously that mistake had to happen to me 🤦
Seems like I messed up when developing...

urequests/urequests.py
```diff
@@ -122,3 +125,22 @@ def patch(url, **kw):
...
+    if not isinstance(s, (bytes, bytearray)):
+        raise TypeError("expected bytes, not %s" % s.__class__.__name__)
+    # Strip off the trailing newline
+    encoded = ubinascii.b2a_base64(s)[:-1]
+    if altchars is not None:
+        if not isinstance(altchars, bytes_types):
+            raise TypeError("expected bytes, not %s"
+                            % altchars.__class__.__name__)
+        assert len(altchars) == 2, repr(altchars)
+        return encoded.translate(bytes.maketrans(b'+/', altchars))
```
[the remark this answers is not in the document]
fschmi: Good to know. Shouldn't we resolve it [here](https://github.com/micropython/micropython-lib/blob/master/base64/base64.py) then? As mentioned, it's a simply a copy.

urequests/urequests.py
```diff
@@ -122,3 +125,22 @@ def patch(url, **kw):
...
+    return encoded
```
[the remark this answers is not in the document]
fschmi: Very good remark. Seems like I wasn't paying much attention to the functionality of `b64encode`. Simply knew I needed it ;-)

urequests/urequests.py
fschmi: > ... this would be an extension not supported by the original `requests.request()`.

How are we dealing with that? Do we actually want to extend the original or simply replicate it's functionality? Is there any guidelines for that?

fschmi:
I have decided to open a new pull request with a working version and some of the changes suggested by @SpotlightKid.

jrborbars:
I'm getting this error with urequests. (micropython 1.12 from repo, ESP32)
```
File "urequests.py", line 108, in get
TypeError: unexpected keyword argument 'auth'
```
Sorry about to open this thread again, but I cannot find the "auth" parameter in the actual code, even in the #311.
Thank you in advance.

SpotlightKid:
https://github.com/micropython/micropython-lib/blob/46e0fa18e76af9ce571d97590874282847d8e60f/urequests/urequests.py#L35

jrborbars:
The line in the master branch is as follow:
```
def request(method, url, data=None, json=None, headers={}, stream=None):
```
This pull request was not applied?
Thanks in advance.

SpotlightKid:
> This pull request was not applied?

No, as you can see from the PR still being open.

FalconBFR:
To whoever it was who asked me whether I found a solution or not, the answer is sort of. I am using JWT (Token Auth) instead of Basic Auth for my esp32 Micropython project now. It gets and refreshes the token every 5 minutes from the server with the username and password passed as clear text params and then the tokens are used in the header of each request. Hope it helps.

jrborbars:
Good catch. Can you provide some basic example?
For documentation purposes.
Thank you.

PR #311, opened 2018-10-04
urequests: Added Basic Authentication (as in requests.request())

fschmi:
Attempt number 2. Fixed some major bugs.

### DIFF
--- a/urequests/metadata.txt
+++ b/urequests/metadata.txt
@@ -1,4 +1,4 @@
 srctype = micropython-lib
 type = module
-version = 0.6
+version = 0.6.1
 author = Paul Sokolovsky
--- a/urequests/setup.py
+++ b/urequests/setup.py
@@ -7,7 +7,7 @@ sys.path.append("..")
 import sdist_upip
 
 setup(name='micropython-urequests',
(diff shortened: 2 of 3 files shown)


FalconBFR:
Hello. I am doing a urequest with basic auth with the latest version of MicroPython in 2020. It is telling me that the keyword 'auth' is unexpected. It is a post request:
```
res = urequests.post(request_url, headers={'content-type': 'application/json'}, data=post_data, auth=('username','password')).json()
        print(res.text)
```
Is this not implemented? 
Thank you very much for your kind help in advance @fschmi

FalconBFR:
@fschmi

andrewleech:
Hi, a version of this PR has been merged in #500, commit e7e8eff8.

@fschmi I hope you don't mind, I maintained your authorship on the commit even though I changed the code slightly based on review comments.
* replace `<str>.decode()` with `str(<str>, "ascii")` as not all ports have the decode functione enabled.
* inline the encode_basic_auth function into an inline block - while the function is arguably clearer to read it results in slightly larger bytecode, the inline version saves flash/ram.

PR #315, opened 2018-10-31
Fix false negative in assertRaises when exception type is AssertionError

sasiso:
The test expected that exception will be raised and test shall fail if the exception is not raised.

But when the user passes exception type AssertionError,  and code under test do not produce this exception. It is expected that the test would fail. but in fact, the test passes.

This is because of line  `assert False, "%r not raised" % exc`
inside the line in the try-except block produces similar excepton and test framework thinks it was raised by user code.

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
-  </component>
-</project>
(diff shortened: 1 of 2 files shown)


pfalcon:
Thanks for the patch. Please note that up to date unittest version is at https://github.com/pfalcon/micropython-lib . Please also note that there're guidelines which patches should adhere to: https://github.com/micropython/micropython-lib/wiki/ContributorGuidelines (again, more up to date version at https://github.com/pfalcon/micropython-lib/wiki/ContributorGuidelines).

pfalcon:
Updated to follow code guidelines and merged to upstream, thanks.

PR #323, opened 2019-01-09
umqtt documentation

shinn16:
I was recently using the library and thought it might be helpful to include some more documentation in the code. The documentation I added is pep8 compliant and gives some information on what variable types the various inputs are. It's not much, but I hope it helps!


pfalcon:
This would increase the size of modules, effectively taking away storage space from users' needs.

shinn16:
Would it be better to move the docstrings' contents over to a readme instead? Or

PR #326, opened 2019-01-12
uasyncio: fixed remove_reader crash

Lupino:


jimmo:
This applies to uasyncio v2 which has since been removed, and replaced by v3 in the main repo.

PR #338, opened 2019-04-27
add tzinfo constructor (fixes #319)

### LINKED ISSUE #319: datetime 'tzinfo' has no attribute '__new__'
When I try to import datetime using latest build of LoBo MicroPython I get the following error:

> >>> import datetime
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "datetime.py", line 1919, in <module>
>   File "datetime.py", line 1836, in _create
> AttributeError: type object 'tzinfo' has no attribute '__new__'
> >>>

miguelgrinberg:
This patch addresses the following `datetime` error:

```
AttributeError: type object 'tzinfo' has no attribute 'new'
```

### DIFF
--- a/datetime/datetime.py
+++ b/datetime/datetime.py
@@ -914,6 +914,11 @@ class tzinfo:
     Subclasses must override the name(), utcoffset() and dst() methods.
     """
     __slots__ = ()
+
+    def __new__(cls):
+        self = object.__new__(cls)
+        return self
+
     def tzname(self, dt):
         "datetime -> string name of time zone."
         raise NotImplementedError("tzinfo subclass must override tzname()")


agners:
This fixes #319 for me nicely, thx!

datetime/datetime.py
```diff
@@ -914,6 +914,11 @@ class tzinfo:
     Subclasses must override the name(), utcoffset() and dst() methods.
     """
     __slots__ = ()
+
+    def __new__(cls):
```
stinos: If you run the test suite this will fail at some point because more arguments are passed. I'd suggest you go with 
```
    def __new__(cls, *args, **kwargs):
        return object.__new__(cls)
```

stinos:
@dpgeorge this is low-hanging fruit to get more CPython compatibility. Even though with a fix the test suite doesn't pass completely (on CPython nor on MicroPython) can we this fix in so it's at least possible to import datetime?

datetime/datetime.py
18570461620: thank you ,that works

projectgus:
Sorry this sat around for so long. I believe fc86070ffb8aeea replaced the code that requires the fix.

PR #343, opened 2019-05-28
urequest socket instance issue

jotathebest:
Recently, new usocket release does not accept four parameters to retrieve the address information using DNS. This PR fixes the method instance (with backward compatibility) and the usocket instantiation


dpgeorge:
All `socket.getaddrinfo` should accept 4 args, and all `socket.socket` constructors should accept 3.  If that's not the case on a particular port, that port needs fixing.

PR #344, opened 2019-06-06
uasyncio.core.call_later_ms: sometimes bad delay type could happen.

pmp-p:
Sometimes it's possible to see this kind of error at runtime, at least on esp8266:
```
  File "uasyncio/core.py", line 159, in run_forever
  File "uasyncio/core.py", line 58, in call_later_ms
TypeError: can't convert float to int
```
possibly not the ideal place to fix it, but worked for me


jimmo:
This applies to uasyncio v2 which has since been removed, and replaced by v3 in the main repo.

PR #346, opened 2019-06-27
sdist_upip.py: fix list index out of range

smlng:
When looping over the list of files in MANIFEST it rsplits the file suffix
and compares against `py`, however this results in an error for file without
a suffix such as `LICENSE`. By changing index from `1` to `-1` it supports
files with and without suffix.


dpgeorge:
This script is now gone, so this PR is no longer required.

PR #347, opened 2019-07-19
Reenable use of string methods in cmd.py

winks:
string.py got the features that this file claimed it didn't in the meantime


jimmo:
Thanks @winks , but despite the comments in the code I think we should leave this as is:

- If someone doesn't already have `string` installed (or aren't otherwise importing it), this adds an extra dependency and more memory used.
- If they do, then the resulting concatenated string has to be allocated separately, so there's no gain from using the string.x components.

I will update the comments in this file to remove the notes about `string`.

PR #358, opened 2019-11-25
upip numerical version extention

Stephan-kashkarov:
I have written the functionality to provide a numerical package version according to PEP508 and PEP440. I have not implemented letter based versioning systems such as >=1.2.5rc5 but versions can have infinite subversions. The implemented operators include >, >=, <, <=, ==, ~=, ===, !=. Extra arguments are dumped as I am unsure how meta tags for upython are handled, i.e. `upip install example;python_version<"2.7"`. Wildcard characters are also included. Thanks for looking at this pull request

### DIFF
--- a/upip/upip.py
+++ b/upip/upip.py
@@ -254,7 +254,7 @@ def parse_version(string):
         Parse Version
 
         This function takes a string and gets all versioning infromation for
-        pypi according to PEP508
+        pypi according to PEP508 and PEP440
 
         Written by: Stephan Kashkarov 2019
     """
@@ -268,6 +268,7 @@ def parse_version(string):
         versioning = versioning.split(";")[0].split(",")
 
     # split version parameters ie "<=3.2.3, > 2.7.1, !=3.0.1" -> [('<=', [3, 2, 3]), ('>', [2, 7, 1]), ('!=', [3, 0, 1])]
+    # Does not support letters in the version name (i.e. 1.2.5rc3)
     parameters = []
     index = 0
     while len(versioning):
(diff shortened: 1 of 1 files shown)


Stephan-kashkarov:
Not tested in micropython but tested in python3.7 on a per-function basis. I have not tested the wildcard support however

upip/upip.py
```diff
@@ -243,6 +249,113 @@ def cleanup():
         except OSError:
             print("Warning: Cannot delete " + fname)
 
+def parse_version(string):
```
mattytrentini: Should probably have unit tests around this method. I may be able to help there...
Also, consider changing the variable `string` since it's very generic. Perhaps `version_string`?

I'll try and test this (on the unix and ESP32 ports) soon. I've also only had a superficial glance over the parsing logic; will try to understand it better.

upip/upip.py
```diff
@@ -243,6 +249,113 @@ def cleanup():
...
+            parameters.extend([("<=", version[:-1]), (">=", version)])
+            continue
+
+    if not versions:
+        raise NoVersionError
+    # returns max suitable operator
+    pkg['version'] = max(versions)
+    return pkg, data
+
+def ver_list_cmp(list1, list2):
```
mattytrentini: Again, unit testing this would be relatively straightforward and beneficial!
Also, I believe that if you used tuples then the built-in tuple comparisons will work correctly. Haven't yet validated that though...

upip/upip.py
```diff
@@ -268,7 +381,7 @@ def main():
         help()
         return
 
-    if sys.argv[1] != "install":
+    if sys.argv[1] not in ["install"]:
```
mattytrentini: I think the original is more readable. The 'not in' form is great for testing if an item is in a collection but that's not the case here.

jimmo:
Thanks @Stephan-kashkarov -- since this PR was raised, upip has been removed and replaced with mip which has much simpler handling of package versions.

PR #359, opened 2019-11-26
Allow overriding PREFIX in Makefile

dlech:
This allows installing somewhere other than the default location


mbuesch:
I don't think this change is needed.
The PREFIX can be overridden as such:

```
make install PREFIX=/tmp/foo
```

dlech:
For some reason, I thought only assignment with `?=` could be overridden from the command line. But of course, you are right.

PR #360, opened 2019-12-03
normpath integration

bradley499:
Directory normalisation for the normpath attribute to the os.path module, this would allow the function to return a normalised value for the directory provided, instead of returning the provided directory path; this now enacts on the directory path string and generates the normalised directory path string from a relative or absolute directory start.

### DIFF
--- a/os.path/os/path.py
+++ b/os.path/os/path.py
@@ -15,6 +15,8 @@ def normpath(s):
                 sp.pop()
         elif d != ".":
             sp.append(d)
+    if len(sp) == 0:
+        return "/"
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks for the contribution, and sorry this never got any attention.

It's a nice addition, but has a few issues, eg:
- `normpath('/.')` returns `'.'` when it should return `'/'`
- `path.normpath('a//b')` returns `'a//ba'` when it should return `'a/b'`
- it needs tests

So I'll close it for now.  Feel free to reopen with the above addressed.

PR #378, opened 2020-03-24
urequests.py: cookies, redirect, url-parameters

mardigras2020:
* added optional parameters: params for URL, cookies, followredirect
* cookie support (necessary for som oauth2 workflows)
* redirect: follow redirects, store redirect url
* parameters for URL support added


dpgeorge:
Sorry we never got to this PR.  There are too many changes here, and the style doesn't match existing code.

PR #382, opened 2020-04-14
umqtt.simple: Close old socket on connect

x3a:
This closes the old socket of the client if one is present during a connection attempt to prevent the initialization of multiple sockets when the function is called repeatedly. It comes in handy if properly disconnecting from the broker isn't possible.

### DIFF
--- a/umqtt.simple/umqtt/simple.py
+++ b/umqtt.simple/umqtt/simple.py
@@ -53,6 +53,8 @@ class MQTTClient:
         self.lw_retain = retain
 
     def connect(self, clean_session=True):
+        if self.sock:
+            self.sock.close()
         self.sock = socket.socket()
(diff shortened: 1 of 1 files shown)


elvis-epx:
This PR is important because in ESP32 the maximum number of sockets is around 8 or 10, so if the old sockets are not explicitly closed or garbage-collected, they pile up until no new connections are possible.

dpgeorge (approved):
This is a reasonable thing to do to improve usability, and it's a minimal change.

(It was also proposed in #586, which adds weight to its usefulness.)

dpgeorge:
Rebased and merged in 5ba4b3b39f47561d72a3d0c21236e221972f9043 (with a bump to the patch version of the library).

Thanks for the contribution and sorry it took 6 years to get to!

PR #385, opened 2020-05-05
umqtt.robust: fix check_msg() hanging after reconnect

EddieParis:
The patch consist of memorizing the fact that the reconnect comes from a check_msg() call.

Tested for month here on esp8266 connected to a mosquitto server on the local network


dpgeorge:
Thanks for the contribution.  But a simpler fix was merged in 4dc2d5e17f1dfa0ca8af731cb1b6eec437731b25

PR #392, opened 2020-06-18
close socket after each request

frahlg:
When running this library on a LoPy4 it crashes after a couple of requests. The problem has been described here: https://forum.pycom.io/topic/1747/urequests-with-ussl-causes-an-oserror/4 and it seems to be that the memory is limited, and the socket needs to be closed after each request. I have added a s.close() , and the library is now working for me.


Kieran7741:
Another option is to close the socket on the response object. I just use my own custom get function that closes the socket.

```python
import urequests

def get(endpoint):
    res = urequests.get(endpoint)
    res.close()
    return res
```

PR #393, opened 2020-06-19
urequests: fix empty body response with some formatings

jesuslg123:
Some engines generate responses with break lines using only \n instead of \r\n, so in order to support that kind of request response body format I have also added this check. Without the fix the body response comes empty as it does not stop on the if and empty the buffer with `readline()`

### DIFF
--- a/urequests/urequests.py
+++ b/urequests/urequests.py
@@ -87,7 +87,7 @@ def request(method, url, data=None, json=None, headers={}, stream=None):
             reason = l[2].rstrip()
         while True:
             l = s.readline()
-            if not l or l == b"\r\n":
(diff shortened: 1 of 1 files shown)


andrewleech:
I'm not too sure about this, the http RFC appears to be pretty clear that `\r\n` is expected: https://stackoverflow.com/questions/5757290/http-header-line-break-style

I'd certainly prefer that the server gets fixed rather than patching a workaround here - do you know if this kind of issue is common?

jesuslg123:
Hi @andrewleech ,

I agree, It is a pain to need manually replace the library as the server implementation won't be updated but I have to agree with you that the implementation should follow the specs. 

I will close the PR.

PR #394, opened 2020-06-25
Fixing TypeError in error message

sfromis:
It is pretty annoying to get
```
File "urequests.py", line 95, in request
TypeError: can't convert 'bytes' object to str implicitly
```
instead of being told what was not supported. :-) Fixing


sfromis:
Fixed in https://github.com/micropython/micropython-lib/commit/106c28a4d7f2358ae0e8c3272739aa504822aab7

PR #397, opened 2020-09-29
urequests: added chunked requests and redirection support

diefesson:


stinos:
The Github diff shows the entrire file changed, which isn't actually the case, so probably all the newlines in the file changed. Can you check your git settings (e.g. turning off `core.autocrlf`) or text editor, adjust and push again? There should also be only 1 commit, since ther's only one change, so try to get rid of the merge commits and write a commit message following [the guidelines](https://github.com/micropython/micropython/blob/master/CODECONVENTIONS.md)

diefesson:
ok

PR #398, opened 2020-09-30
urequests: added support for chunked requests and redirection

diefesson:
Added support for chunked requests, allowing urequests to send data of
unknown size (eg: generated or collected data that would be too big to
be stored in RAM), and HTTP redirection ("Location" header) support
for temporary and permanent redirection.


andrewleech:
Hi, a version of this PR has been merged in #500, commits be327a7b, b17e9aaf & b29cffb3.

@diefesson I hope you don't mind, I maintained your authorship on the commits even though I changed them slightly based on review commits and split the redirects & chunk handling into two separate commits for clarity of change.

PR #399, opened 2020-09-30
ffilib:fix Traceback in Unix port due to missing 'maxsize'

pvinci:
>>> from machine import I2C
  File "/root/.micropython/lib/machine/__init__.py", line 2, in <module>
  File "/root/.micropython/lib/machine/timer.py", line 1, in <module>
  File "/root/.micropython/lib/ffilib.py", line 43, in <module>
AttributeError: 'module' object has no attribute 'maxsize'

### DIFF
--- a/ffilib/ffilib.py
+++ b/ffilib/ffilib.py
@@ -1,4 +1,4 @@
-import sys
+import usys as sys
 try:
     import ffi
 except ImportError:
--- a/ffilib/ffilib.py
+++ b/ffilib/ffilib.py
@@ -1,4 +1,4 @@
-import usys as sys
+import sys
 try:
     import ffi
 except ImportError:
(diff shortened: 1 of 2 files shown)


jimmo:
I think the issue here is that micropython-lib's `sys` module needs to be updated since https://github.com/micropython/micropython/pull/6164 renamed the builtin from `sys` to `usys`.

So what used to happen is that `import sys` got the builtin, whereas since https://github.com/micropython/micropython/pull/6164 it now gets micropython-lib's empty sys.py

The fix is that sys.py should "inherit" from usys by having `from usys import *` at the start. This is what's done for other modules. Are you able to test that and see if it resolves the issue?

The other fix is that micropython-lib shouldn't contain these empty placeholder modules.... see #376 for some thoughts on that.

PR #401, opened 2020-10-24
signal: Add alarm() and pause() function

Hagb:
This PR adds `alarm` and `pause` function to call these two syscalls.

### DIFF
--- a/unix-ffi/signal/manifest.py
+++ b/unix-ffi/signal/manifest.py
@@ -1,4 +1,4 @@
-metadata(version="0.3.2")
+metadata(version="0.4.0")
 
 # Originally written by Paul Sokolovsky.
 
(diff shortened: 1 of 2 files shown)


signal/signal.py
```diff
@@ -6,16 +6,25 @@
...
 SIGINT = 2
 SIGPIPE = 13
+SIGALRM = 14
 SIGTERM = 15
 
 libc = ffilib.libc()
 
 signal_i = libc.func("i", "signal", "ii")
 signal_p = libc.func("i", "signal", "ip")
+alarm_i = libc.func("I", "alarm", "I")
```
dlech, line 16: since there is no special handling needed for arguments, this could just be

    alarm = libc.func("I", "alarm", "I")

and skip the wrapper below. The same applies to `pause`.

Hagb: Thanks. The file have been changed.

dpgeorge (approved):
This looks fine, and I verified the behaviour is the same as CPython.

dpgeorge:
Sorry this took nearly 6 years to get to, but it's now merged!

PR #411, opened 2021-02-03
xmltok: fix StopIteration deprecated in PEP-479

ThunderEX:
Due to change in micropython for [PEP-479](https://www.python.org/dev/peps/pep-0479/), StopIteration has been deprecated for return. This results in xmltok to raise RuntimeError every time.
This is a simple fix to just change from StopIteration to EOFError and then return it in generator.


dpgeorge:
Thanks for the fix, it's definitely needed and works well (the tests pass again).

I rebased on master and fixed the conflicts.

PR #418, opened 2021-04-22
Add support for Uefi to various fileio libraries

Zitt:


glob/glob.py
```diff
@@ -2,7 +2,15 @@
 
 import os
 import os.path
-import re
+import sys
+
+if (sys.platform == 'uefi') and (sys.implementation.name == 'micropython'):
```
stinos, line 7: Canonical way of doing this is
```
try
  import _re
except ImportError:
  import ure
except importError:
  import re
```
which doesn't have the downside of hardcoding these names.
However, the `import re` was there already: why is it now needed to try the `ure` import (not related to uefi I assume)?

fnmatch/fnmatch.py
```diff
@@ -26,10 +35,10 @@ def fnmatch(name, pat):
     [seq]   matches any character in seq
     [!seq]  matches any char not in seq
 
-    An initial period in FILENAME is not special.
+    An initial period in FILENAME is not special
```
stinos, line 38: Unrelated formatting would better go in separate sommits. Thoug I'm not sure this is even need?

fnmatch/fnmatch.py
```diff
@@ -50,9 +59,15 @@ def filter(names, pat):
     result = []
     pat = os.path.normcase(pat)
     match = _compile_pattern(pat)
-    for name in names:
-        if match(os.path.normcase(name)):
-            result.append(name)
+    if os.path is posixpath:
+        # normcase on posix is NOP. Optimize it away from the loop.
```
stinos, line 63: Have you measured if this has an actual impact?

fnmatch/fnmatch.py
```diff
@@ -64,6 +79,29 @@ def fnmatchcase(name, pat):
     match = _compile_pattern(pat)
     return match(name) is not None
 
+_re_special_chars_map = {i: '\\' + chr(i) for i in b'()[]{}?*+-|^$\\.&~# \t\n\r\v\f'}
+def _re_translate(p, chmap):
+    oldp = p
+    for idx,c in enumerate(p):
+     print('[%d]=%s' % (idx, c))
```
stinos, line 86: Leftover? Also please remove the other debug statements.

fnmatch/fnmatch.py
```diff
@@ -64,6 +79,29 @@ def fnmatchcase(name, pat):
...
+     if ord(c) in chmap:
+      newp = p.replace(c, chmap[ord(c)])
+      #print( '%s -> %s' % (oldp, newp) )
+      return newp
+    #print( '%s -> %s' % (oldp, p) )
+    return p
+
+def _re_escape(pattern):
```
stinos, line 94: Would be best to add tests for this.

fnmatch/fnmatch.py
```diff
@@ -64,6 +79,29 @@ def fnmatchcase(name, pat):
...
+    """
+    Escape special characters in a string.
+    """
+    print( _re_special_chars_map )
+    if isinstance(pattern, str):
+        #return pattern.translate(_re_special_chars_map)
```
stinos, line 100: dead code

fnmatch/fnmatch.py
```diff
@@ -64,6 +79,29 @@ def fnmatchcase(name, pat):
...
+      #print( '%s -> %s' % (oldp, newp) )
+      return newp
+    #print( '%s -> %s' % (oldp, p) )
+    return p
+
+def _re_escape(pattern):
+    """
+    Escape special characters in a string.
+    """
+    print( _re_special_chars_map )
```
stinos, line 98: not needed

fnmatch/fnmatch.py
```diff
@@ -64,6 +79,29 @@ def fnmatchcase(name, pat):
...
     return match(name) is not None
 
+_re_special_chars_map = {i: '\\' + chr(i) for i in b'()[]{}?*+-|^$\\.&~# \t\n\r\v\f'}
+def _re_translate(p, chmap):
+    oldp = p
+    for idx,c in enumerate(p):
+     print('[%d]=%s' % (idx, c))
+     if ord(c) in chmap:
+      newp = p.replace(c, chmap[ord(c)])
+      #print( '%s -> %s' % (oldp, newp) )
```
stinos, line 89: remve this and return in one line

fnmatch/fnmatch.py
```diff
@@ -64,6 +79,29 @@ def fnmatchcase(name, pat):
...
+def _re_escape(pattern):
+    """
+    Escape special characters in a string.
+    """
+    print( _re_special_chars_map )
+    if isinstance(pattern, str):
+        #return pattern.translate(_re_special_chars_map)
+        return _re_translate(pattern, _re_special_chars_map )
+    else:
+        pattern = str(pattern, 'latin1')
```
stinos, line 103: Hardcoding an encoding might not be ok here? And why latin1?

fnmatch/fnmatch.py
```diff
@@ -99,6 +137,11 @@ def translate(pat):
                     stuff = '\\' + stuff
                 res = '%s[%s]' % (res, stuff)
         else:
+         if (sys.platform == 'uefi') and (sys.implementation.name == 'micropython') :
```
stinos, line 140: Probably niver t make this a 'constant' at the top of the file. However, is this check always correct?

os.path/os/path.py
```diff
@@ -1,7 +1,38 @@
...
-
+import sys
 
 sep = "/"
+altsep = "\\"
+if (sys.platform == 'uefi') and (sys.implementation.name == 'micropython'):
+ sep = '\\'
+ altsep = '/'
+ import _re as re
+else: import ure as re
```
stinos, line 10: Code style not consistent.

os.path/os/path.py
```diff
@@ -1,7 +1,38 @@
...
+    if isinstance( s, (str, bytes) ): return s
+    raise TypeError( '%s is unknown type' % s )
+
+def _get_bothseps(path):
+    if isinstance(path, bytes):
+        return b'\\/'
+    else:
+        return '\\/'
+
+def isabs(s):
```
stinos, line 22: Tests would be nice.

os.path/os/path.py
```diff
@@ -1,7 +1,38 @@
 import os
-
+import sys
 
 sep = "/"
+altsep = "\\"
+if (sys.platform == 'uefi') and (sys.implementation.name == 'micropython'):
+ sep = '\\'
```
stinos, line 7: This is actually also true for windows. Does this mean all the uefi stuff here is appliccable for windows as well?

glob/glob.py
Zitt: Good point regarding re vs ure. I guess my thought was *if* someone were to try and run the micropython under cpython; wouldn't it be better to fall back to standard re which would be available on that platform?

fnmatch/fnmatch.py
Zitt: I didn't specifically add the 'period' at the end; but it could have been a typo on my part.

fnmatch/fnmatch.py
Zitt: I didn't specifically make this change either. Wonder what is going on.

fnmatch/fnmatch.py
Zitt: Correct.

fnmatch/fnmatch.py
Zitt: I'm not familiar with tests; would need someone to assist on writing tests.

fnmatch/fnmatch.py
Zitt: A copy from cpython implementation. Felt is was better to leave as-is; rather than risk a bug later if/when encoding is supported.

fnmatch/fnmatch.py
Zitt: This code is correct; all though; there seems to be some indentation issues. 
I'll re-check on my master file. That said the "else" goes with the 
`elif stuff[0] == '^':`
above it so that the _re_escape(c) is called in the case of uefi implementation.

fnmatch/fnmatch.py
Zitt: Why do you believe this is dead code? what if pattern is a bytestring?
Again; this code is present in cpython IIRC.

os.path/os/path.py
Zitt: I don't follow. What code style am I violating?

os.path/os/path.py
Zitt: Possibly; however, I don't have a specific need for windows at the moment? Is there a developer who wants to take on windows?

glob/glob.py
stinos: If you look in the source code, using the try/except is the canonical way to do this for MicroPython so it is only vague if you haven't seen any of the other code. Though I get what you mean. Still, I don't think we can just go changing these `import re` everywhere. Does it not work for you?

fnmatch/fnmatch.py
stinos: See test_fnmatch.py in this directory

fnmatch/fnmatch.py
stinos: I'm talking about the comment: doesn't do anything, unclear why it is commented out.

fnmatch/fnmatch.py
stinos: Is it ok to just copy from CPython without acnkowledging that? What does its license say?

os.path/os/path.py
stinos: Well, MicroPython uses black for formatting. But only for the micropython repository currently, not yet for micropython-lib. In any case: all statements should go on their own line for consistency and readability

os.path/os/path.py
stinos: I almost exclusively use Windows so if there's a major effort towards getting this path problem fixed it would be nice and I could contribute. That's also why I think the 'if platform is X and something else is Y' should be abstracted away in one place instead of repeating. However: in the end I just resolved to changing all backslahes into forward slashes whenevr I use paths: IIRC there are other places apart from os.path and fnmatch which have problems with backslashes.

dpgeorge:
Are you still interested in this PR?  It will need some work to rebase it on the latest master commit.

PR #419, opened 2021-04-24
- show traceback when test failed

maho:
i was surprised that exception in tested function is just silently redirected to nowhere. 

I think it's good to show the traceback.


stinos:
Is this the same as what CPython does?

maho:
Let's check:

test_ee.py

```python
import unittest

class MyTestCase(unittest.TestCase):

    def _fn2(self):
        raise Exception("this shouldn't happend, but it did")

    def _fn(self):
        return self._fn2()


    def test_exception(self):
        self._fn()

if __name__ == '__main__':
    unittest.main()
       

```

micropython:

```bash
% micropython test_ee.py                                                                                                                                                                  :( 1 21-04-25 - 17:39:17
test_exception (MyTestCase) ... FAIL
  File "/home/maho/workspace/servo/dispenser/.micropython-lib/unittest/unittest.py", line 201, in run_class
  File "test_ee.py", line 13, in test_exception
  File "test_ee.py", line 9, in _fn
  File "test_ee.py", line 6, in _fn2
Exception: this shouldn't happend, but it did
Ran 1 tests

FAILED (failures=1, errors=0)
```

CPython 3.8:

```bash
% python3 /tmp/test_ee.py                                                                                                                                                                 :( 1 21-04-25 - 17:39:23
E
ERROR: test_exception (__main__.MyTestCase)
  File "/tmp/test_ee.py", line 13, in test_exception
    self._fn()
  File "/tmp/test_ee.py", line 9, in _fn
    return self._fn2()
  File "/tmp/test_ee.py", line 6, in _fn2
    raise Exception("this shouldn't happend, but it did")
Exception: this shouldn't happend, but it did

Ran 1 test in 0.000s

FAILED (errors=1)
```

So this change makes it closer to behaviour of CPython.

stinos:
Makes sense. Adding a newline would be nice though because raising multiple exceptions results in text which gets hard to read.

unittest/unittest.py
```diff
@@ -200,8 +206,7 @@ def run_class(c, test_result):
             except:
                 print(" FAIL")
                 test_result.failuresNum += 1
-                # Uncomment to investigate failure in detail
-                #raise
+                print_exc()
```
stinos, line 209: would be simpler to `except Exception as e` and then pass e to `print_exc`: no need for `sys.exc_info()[1]`

unittest/unittest.py
```diff
@@ -180,6 +180,12 @@ def __init__(self):
     def wasSuccessful(self):
         return self.errorsNum == 0 and self.failuresNum == 0
 
+
+def print_exc():
+    e = sys.exc_info()[1]
+    sys.print_exception(e, sys.stdout)
```
stinos, line 186: Not all platforms have `print_exception` so that makes this a hard error on those.

unittest/unittest.py
```diff
@@ -180,6 +180,12 @@ def __init__(self):
     def wasSuccessful(self):
         return self.errorsNum == 0 and self.failuresNum == 0
 
+
+def print_exc():
```
stinos, line 184: Personal opinion: no need to obfuscate, line is short enough, just `print_exception` is ok?

maho: applied all comments

unittest/unittest.py
```diff
@@ -180,6 +180,17 @@ def __init__(self):
...
         return self.errorsNum == 0 and self.failuresNum == 0
 
+
+def print_exception(e):
+    try:
+        print("-"*80)
+        sys.print_exception(e, sys.stdout)
+    except Exception:
+        # just print exception
+        print("EXCEPTION: ", e)
```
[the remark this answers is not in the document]
maho: you're absolutely right. Again :). Somehow my flake8 didn't launch in vim in this project.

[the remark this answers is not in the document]
stinos: As you might know the micropython repository itself uses black for formatting and has tools/autoformat.py to mass format everything, eventually something like that should end up in this repository as well and then formatting is a thing of the past :)

maho:
should I do any more changes or it's ok?

maho:
@stinos? Anyone?

stinos:
I think this is ok, but it's not up to me to merge it :)

dpgeorge:
Sorry this got neglected.  The feature was implemented in 7d4d02edfc689811ac9fb120a188498686bfc56d

PR #424, opened 2021-05-27
python-ecosys/iperf3: Add iperf3.

dpgeorge:
This finally has a home here...

Maybe it should be renamed `iperf3` because it's really trying to be compatible with standard `iperf3`?

### DIFF
--- a/python-ecosys/iperf3/iperf3.py
+++ b/python-ecosys/iperf3/iperf3.py
@@ -1,0 +1,537 @@
+"""
+Pure Python, iperf3-compatible network performance test tool.
+
+MIT license; Copyright (c) 2018-2019 Damien P. George
+
+Supported modes: server & client, TCP & UDP, normal & reverse
+
+Usage:
+    import iperf3
+    iperf3.server()
+    iperf3.client('192.168.1.5')
+    iperf3.client('192.168.1.5', udp=True, reverse=True)
+"""
+
+import sys, os, struct
+import time, select, socket
+import json
+
+DEBUG = False
+
+# iperf3 cookie size, last byte is null byte
+COOKIE_SIZE = 37
+
+# iperf3 commands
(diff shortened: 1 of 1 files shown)


jimmo:
> Maybe it should be renamed `iperf3` because it's really trying to be compatible with standard `iperf3`?

Yeah I think this makes sense...

I'd like the `u` prefix to be used to convey "ufoo library is inspired by but is fundamentally different to foo". Whereas this is a drop-in replacement. I think in general the rule should be "if it needs to start with a 'u', then it should be in the micropython dir instead".

(Related... I'd like to rename `urequests` to `requests` by the above logic)

dpgeorge:
This library has already been published to PyPI as `uiperf3`: https://pypi.org/project/uiperf3/, so that might constrain us here.

jimmo:
I'm not sure I've really seen the value of putting MicroPython packages in PyPI... And at least not in the same namespace as regular Python packages.

Obviously having a way to conveniently install packages is important for boards with WiFi, and perhaps an integration with `mpremote` for boards that don't. But I'm not sure PyPI (and in particular the global PyPI namespace) is a necessary part of that.

dpgeorge:
> I'm not sure I've really seen the value of putting MicroPython packages in PyPI... And at least not in the same namespace as regular Python packages.

In this particular case `uiperf3` is providing a pure-Python implementation of `iperf3` (which is a wrapper around the iperf3 C library), so it has some merit in that sense.  But maybe it should be called `iperf3-python` if that's it's main goal.  This is a related (but slightly different) goal to the goal of having it run under MicroPython, and easily installable on a MP-based system.

But the naming of modules in this repo should follow the logic/aims of this repo, in which case it should be called `iperf3`.  What we can then do is separately publish it to PyPI as `uiperf3` (via a script that automatically renames it, for example).

PR #426, opened 2021-05-30
python-ecosys/urequests: add streaming API support and cache standard responses

mkomon:
As the title says.

### DIFF
--- a/micropython/bluetooth/aioble/multitests/ble_shutdown.py
+++ b/micropython/bluetooth/aioble/multitests/ble_shutdown.py
@@ -1,0 +1,128 @@
+# Test for shutting down and restarting the BLE stack.
+
+import sys
+
+sys.path.append("")
+
+from micropython import const
+import time, machine
+
+import uasyncio as asyncio
+import aioble
+import bluetooth
+
+TIMEOUT_MS = 5000
+
+SERVICE_UUID = bluetooth.UUID("A5A5A5A5-FFFF-9999-1111-5A5A5A5A5A5A")
+CHAR_UUID = bluetooth.UUID("00000000-1111-2222-3333-444444444444")
+
+_L2CAP_PSN = const(22)
+_L2CAP_MTU = const(128)
+
+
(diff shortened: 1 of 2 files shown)


mkomon:
By the way I am using urequests with these two commits with lichess.org API and all works well, both standard and streaming API requests. Without this patch streaming requests don't work at all as they read only the immediate response and close the socket immediately.

python-ecosys/urequests/urequests.py
```diff
@@ -7,6 +7,17 @@ def __init__(self, f):
...
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        data = b""
+        chunk = self.raw.read()
+        while chunk:
+            data += chunk
+            chunk = self.raw.read()
+        return data
```
dpgeorge, line 19: Does it make sense to combine all data here?  Why not just return the result of the first read?  The caller must anyway keep calling this function until it has all data.

python-ecosys/urequests/urequests.py
```diff
@@ -105,6 +116,12 @@ def request(method, url, data=None, json=None, headers={}, stream=None):
     resp = Response(s)
     resp.status_code = status
     resp.reason = reason
+
+    if stream:
+        resp.raw.setblocking(False)
+    else:
+        resp.content
```
dpgeorge, line 123: I'm not sure it makes sense to pre-read the content here.  Some uses of this may not need the content and then it shouldn't waste time and memory retrieving it.

python-ecosys/urequests/urequests.py
mkomon: Good point, that makes sense and the loop just adds an unnecessary read. I will test it with `return self.raw.read()' and report back.

python-ecosys/urequests/urequests.py
mkomon: I have not considered such use case but it makes total sense. There is no need to read the response like the original requests package does. I'll remove 9b5d30c.

PR #427, opened 2021-05-30
unix-ffi/datetime: make the package importable

mkomon:
-add constructor to tzinfo class so that the package can be imported without errors

### DIFF
--- a/unix-ffi/datetime/datetime.py
+++ b/unix-ffi/datetime/datetime.py
@@ -958,6 +958,10 @@ class tzinfo:
 
     __slots__ = ()
 
+    def __new__(self, *args, **kwargs):
+        """Constructor."""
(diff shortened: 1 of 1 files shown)


dpgeorge:
What error do you see without this method?

stinos:
```
  File "datetime.py", line 1919, in <module>
  File "datetime.py", line 1836, in _create
AttributeError: type object 'tzinfo' has no attribute '__new__'
```

dpgeorge:
I was going to suggest it might be better to change the call in `_create` so it did `tzinfo()`, but `__new__` is used a lot in this module so probably not.

dpgeorge:
Merged in 2e91b924135deee7e287fb20bdd6b158a2914b66

PR #428, opened 2021-06-09
Brought back from an old commit

dstromberg:
Restoring the Makefile, to facilitate installing all of micropython-lib.


stinos:
This uses max-depth=1, does this still work?
In any case imo ths should be replaced by Python code so it runs on Windows as well. And the code should allow selecting which of the subdirectories to install.

In a later installment I guess upip should allow installation from a local respository (and from github as well perhaps) like pip and then this tool here would call into upip to do the installation of each module.

dstromberg:
I've rewritten it in Python 3.4.

Thanks :)

jimmo:
Thanks @dstromberg 

So I can understand this better, what's the underlying workflow that you're using this for? I'm guessing you have the unix build in your `PATH`, and you want to be able to use micropython-lib packages from anywhere? i.e. you're using micropython's unix port as a cpython-replacement, and `~/.micropython/lib` as your site-packages equivalent? Or something else?

My experience is far more embedded-focused, so this isn't a scenario that comes up very often.

dlech:
> you're using micropython's unix port as a cpython-replacement, and `~/.micropython/lib` as your site-packages equivalent?

Yes. The `Makefile` was quite convenient for [creating a debian package](https://github.com/pybricks/micropython-lib/blob/ev3dev-stretch/debian/rules) for `micropython-lib` that included everything.

dstromberg:
On Thu, Jun 10, 2021 at 5:09 PM Jim Mussared ***@***.***>
wrote:

> Thanks @dstromberg <https://github.com/dstromberg>
>
> So I can understand this better, what's the underlying workflow that
> you're using this for? I'm guessing you have the unix build in your PATH,
> and you want to be able to use micropython-lib packages from anywhere? i.e.
> you're using micropython's unix port as a cpython-replacement, and
> ~/.micropython/lib as your site-packages equivalent? Or something else?
>
> My experience is far more embedded-focused, so this isn't a scenario that
> comes up very often
>
I like exploring different dialects of Python.

And yes, I want to be able to easily use micropython on a Linux system.
Actually, I've been patching my micropython binaries to get their library
from, EG, /usr/local/micropython-1.15/lib.

I've been intrigued by the possibility that micropython might thread better
than CPython and Pypy, so I've had it in the back of my mind to write a
little parallel benchmark or two and graph the performance curves.

Thanks.


-- 
Dan Stromberg

installer
```diff
@@ -0,0 +1,139 @@
+#!/usr/local/cpython-3.4/bin/python3
+
+"""
+Copy modules from here to --prefix.
+
+Note that we do not scan for directories; instead we have a hardcoded list in all_directories.
```
stinos, line 6: Why?

installer
```diff
@@ -0,0 +1,139 @@
...
+import shutil
+import sys
+
+
+prefix = os.path.expanduser('~/.micropython/lib')
+directories = []
+all_directories = ['micropython', 'python-ecosys', 'python-stdlib', 'tools', 'unix-ffi']
+
+
+def usage(retval):
```
stinos, line 19: Minor, but imo clearer to pass the  `write` argument directly and have it default to sys.stdout.write: obvious, clearer below in the case stderr gets used beause one doesn't have to wonder what passing `1` means

installer
```diff
@@ -0,0 +1,139 @@
...
+    else:
+        write = sys.stderr.write
+
+    write('Usage: {} --prefix ~/.micropython/lib --directories {}\n'.format(sys.argv[0], ' '.join(all_directories)))
+    write('--directories, if provided, must be the last argument specified\n')
+
+    sys.exit(retval)
+
+
+# The getopt-like module du jour would be a little more concise, but that's much harder for a static analyzer to check.
```
stinos, line 32: argparse is hardly 'du jour' and arguably the statis analyzer is not good if it cannot handle that

installer
```diff
@@ -0,0 +1,139 @@
...
+
+    pass
+
+
+def copytree(src, dst, symlinks=False, ignore=None, copy_function=shutil.copy2,
+             ignore_dangling_symlinks=False):
+    """
+    Recursively copy a directory tree.
+
+    This is a slightly-modified copy of copytree from CPython 3.4.  The motivation was to allow dst to preexist.
```
stinos, line 62: Surely there are simpler ways, no? Doesn't seem right to introduce dependency on such a large function simply to copy some files

installer
```diff
@@ -0,0 +1,139 @@
...
+    cmd = "find . -maxdepth 1 -mindepth 1 ( -name '*.py' -not -name 'test_*' -not -name 'setup.py' )
+        -or ( -type d -not -name 'dist' -not -name '*.egg-info' -not -name '__pycache__' )
+        | xargs --no-run-if-empty cp -r -t $(PREFIX)"
+    Deleted the backslashes from the above shell command for the benefit of syntastic
+    """
+    ignore = shutil.ignore_patterns('*.py', 'test_*', 'setup.py', 'dist', '*.ogg-info', '__pycache__')
+    copytree(source_dir, dest_dir, symlinks=True, ignore=ignore, ignore_dangling_symlinks=False)
+
+
+os.makedirs(prefix, exist_ok=True)
```
stinos, line 132: can you wrap this in a `if __main__` so it doesn't always execute? Much easier for testing

installer
```diff
@@ -0,0 +1,139 @@
...
+    if errors:
+        raise Error(errors)
+    return dst
+
+
+def copy(source_dir, dest_dir):
+    """
+    Copy modules to a lib location.
+
+    This was a shell command:
```
stinos, line 122: Not really interesting to know, but I wouldn't repeat this here because it's a matter of time before it becomes obsolete, because more ignore patterns need to be added. Just calling this copy_module and mentioning it will exclude anything but module files is clear enough?

stinos:
> So I can understand this better, what's the underlying workflow that you're using this for? I'm guessing you have the unix build in your `PATH`, and you want to be able to use micropython-lib packages from anywhere?

That's what we do as well, roughly: to deploy applications which use select items from the micropyton-lib repository we copy over those directories to a lib directory. Which can then get redistributed as-is to other computers. All of this is simpler and faster than using pip, and works without global internet access as well. But as said, in the end it would probably be better to have `upip` handle this.

> I've been patching my micropython binaries

You can set the `MICROPYPATH` environment variable for that, equivalen of `PYTHONPATH`, no need for patching.

installer
dstromberg: No particular reason.  One way you could accidentally copy a new doc/, the other you could accidentally not copy a new from-github.

installer
dstromberg: It's like sys.exit(0)

installer
dstromberg: If you know of a static analyzer that does this, I'll probably want to switch.

installer
dstromberg: It's not really a dependency, but this is a library function that lacks a dirs_exist_ok option.  That's not present until Python 3.8+ though.

installer
dstromberg: if'd.

installer
dstromberg: It's gone.

dstromberg:
I'd really rather patch the binary, so anyone on the system can use it conveniently.

installer
```diff
@@ -0,0 +1,131 @@
+#!/usr/local/cpython-3.4/bin/python3
```
stinos, line 1: Too strict probably, and for consistency with other tools should be `#!/usr/bin/env python3`, and the filename should have the `.py` extension, and perhaps even be placed in tools/ directory though that makes it harder to discover.

Also the file should be formatted using tools/autoformat.py so that one might need to be changed as well (didn't test though).

installer
```diff
@@ -0,0 +1,131 @@
...
+"""
+
+import os
+import shutil
+import sys
+
+
+prefix = os.path.expanduser('~/.micropython/lib')
+directories = []
+all_directories = ['micropython', 'python-ecosys', 'python-stdlib', 'tools', 'unix-ffi']
```
stinos, line 16: tools shouldn't be copied probably, since it copies it codeformat.py which only works for this repository

installer
stinos: What needs to be checked, and what problem are you having with argparse?

installer
stinos: Ok. But it's a dependncy on a lot of non-trivial code unrelated to the actual task at hand. Plus I'm not sure one can just copy from CPython without mentioning its license? Anyway, dirs_exist_ok is actaully a problem in my experience: what typcially happens is a file gets deleted from the source (this repository) but then it just remains in the target and causes havoc there. Deleting all directories first (as in: loop over the first level and delete everything encounterd in prefix) is better and safer. Plus makes copy_tree usable again.

installer
```diff
@@ -0,0 +1,131 @@
...
+        if getattr(why, 'winerror', None) is None:
+            errors.append((src, dst, str(why)))
+    if errors:
+        raise Error(errors)
+    return dst
+
+
+def copy(source_dir, dest_dir):
+    """Copy modules to a lib location."""
+    ignore = shutil.ignore_patterns('test_*', 'setup.py', 'dist', '*.ogg-info', '__pycache__')
```
stinos, line 120: `ogg-info` -> `egg-info` 

But there are some other problems: the original copies only .py files, now everything gets copies, which results in metadata.txt getting copies, and the README.md of all subdirectories, but they are all copies into the same location so one ends up with unix-ffi's readme.md in the root of the prefix.

installer
```diff
@@ -0,0 +1,131 @@
...
+    copytree(source_dir, dest_dir, symlinks=True, ignore=ignore, ignore_dangling_symlinks=False)
+
+
+if __name__ == '__main__':
+    os.makedirs(prefix, exist_ok=True)
+
+    for directory in directories:
+        print('Copying directory {}'.format(directory))
+        os.chdir(directory)
```
stinos, line 129: Would be better to use full paths here, less chance of things not working as expected.

installer
dstromberg: Code needs to be checked, and I haven't used an argument parser since getopt.

installer
dstromberg: The code can be trivially removed when 3.8 is adopted.

CPython is fully opensource.

Goodness, no, deleting a hierarchy isn't even close to safer.

installer
dstromberg: I do a specific #! sometimes, to document what version of Python I was testing against.  I agree that using /usr/bin/env is better for portability, even though it's less clear, and shows up in top less nicely.

installer
dstromberg: It's gone now.

installer
dstromberg: Typo fixed.

As far as what to copy, I've just been rewriting the Makefile.  Maybe we should handle things like the stray README.md as a separate commit?  This is already more arduous than it should be.

installer
dstromberg: No, not really.  Relative paths have their place.  So do absolute paths.  But converting all to the other is not a good thing.

installer
stinos: Can you be more concrete? I.e. this comment sounds like something doesn't work for a particular static analyzer. As such it raises more questions that it answers. My question really is: what does not work, and for which analyzer.

installer
stinos: > Goodness, no, deleting a hierarchy isn't even close to safer.

How so (in this particular case)? Do you know another way to deal with the problem I mentioned?

installer
stinos: Please explain. Right now the tool is only usable from the micropython-lib directory which is quite limiting. As in, first thing I tried was to just run it from another place, like code-format.py, and that didn't work.

installer
stinos: But the stray README is because this does something very different from the Makefile, which used to copy *.py only.

dbrignoli:
The original Makefile, which I use to install micropython-lib, doesn't really use or need any features of _make_ and I agree it would be nice to have something equivalent in Python. However, after reading this conversation and comparing the original shell script from the Makefile and the proposed solution in Python, I wonder if the practical and concise solution could be to have a shell script for bash/posix and one for windows.

stinos:
Note that if it weren't for the copy-tree implementation combined with using argparse this would be pretty concise.

But the more I think about it, the more I feel the way forward is to use `upip`. It already knows which files to skip and how to copy the files etc, just needs some extra bits to not unpack a tar from a URL but to use files on disk. Then unless I'm missing something the installation here would be just 'for subdir in listdir(mainDirs): upip(subdir)'. But changing upip might be non-trivial, don't know.

dpgeorge:
I think this functionality is now provided by `micropython -m mip install ...`.

PR #431, opened 2021-06-27
Add ulogger to micropython-lib

Youkii-Chen:
This is a logging module specially designed for micropython

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "micropython/ulogger"]
+	path = micropython/ulogger
+	url = https://github.com/Li-Lian1069/micropython-ulogger


andrewleech:
Thanks for the contribution, sorry for the lack of response earlier!  This repo avoids using submodules for organising modules however. 
Anyone is more than welcome to publish packages to pypi and they can then be used with `upip` (sorry I dont' really have any docs to describe how to do this though).

We're looking at better way / writing up guidance for how users can better publish package for micropython, I'll try to remember to update here when we get this sorted.

Youkii-Chen:
I'm sorry for not keeping your promise.

But I encountered a problem when packaging to pypi: how to only package the compiled `.mpy` file as an archive and upload it to pypi, so as to improve the execution speed of the code on micropython. The existing pypi packaging scheme only allows `.py` files to be added to the archive。I have a headache about this。

If you have a solution, can you add it to the guide?

andrewleech:
Hi yes that's a good point, there currently isn't any way to have binary packages for micropython uploaded to pypi. 

Publishing mpy files is on the TODO list for this repo and I'm actually investigating solutions to this limitation currently, but there's nothing ready yet unfortunately.

I see now you have pushed to pypi already, sorry I missed that earlier! Your package structure and detail in the readme look great!

One note however, while I personally support your use of GPL Licence, it's technically incompatible with the Python/BSD licences used here. You're still more than welcome to publish / use / share it with the world, we just can't directly include it here!

Issue #384, opened 2020-04-24
HMAC AttributeError


cositotito:
hello all,
I have a problem when using 'import hmac' and I try to use it on my esp32

* before using it, I install it and it installs correctly, the code is the following:
import upip
upip.install ("hmac")

* the code is the following to use hmac:
digest = hmac.new (s.decode ("base64"), a, hashlib.sha3_256) .digest ()

* the error that prints to me on my console:
Traceback (most recent call last):
   File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'new'

How can I use the new or remove this error?
I would appreciate the help

cositotito:
* Code example.py:

import hmac
import hashlib
import time
import urequests
import json

secret="345fg-dfg"
word="536"
signature = hmac.new(secret, word.encode(), digestmod=hashlib.sha256).hexdigest().upper()


* return Error:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/lib/hmac.py", line 149, in new
  File "/lib/hmac.py", line 63, in __init__
AttributeError: 'sha256' object has no attribute 'digest_size'

mvo5:
Fwiw, I ran into the same issue and pushed a PR that fixes this for me. Let's see what the code review says.

dpgeorge:
Should be fixed by f95260d7e3c123a8b1907ded095e4c25e4251085

PR #435, opened 2021-07-21
hashlib: import pure python hash first to unbreak the hmac module

mvo5:
With the current import strategy that favours the uhashlib
implementation before using the "internal" hashlib version external
modules like "hmac" break because the uhashlib versions lack properties
like `{digest,block}_size, copy()`.

This commit changes the import so that the internal version of the
hash is tried to be imported first and only if there is no internal
version then the uhashlib version is used.

This should fix https://github.com/micropython/micropython-lib/issues/384

I guess as an alternative we could try to make uhashlib compatible with hmac.
This should be relatively straightforward and was done in https://github.com/micropython/micropython/pull/4539 but wasn't received favorable there.


mvo5:
I love micropython and would love to see this issue fixed. Is there anything I can do/should do to move this forward? I'm happy to write test cases or anything else that might help :)

tfindlay-au:
@jimmo Can you provide some feedback on this PR ?
I know a few people that are held up by the issue that this would resolve if its accepted.

jimmo:
Thanks @mvo5 and sorry this has taken so long.

It's preferable to use the built-in one where possible -- e.g. an app that only needs to do sha256 hmacs shouldn't need to install hashlib from micropython-lib if the device already has it.

I agree it would be good to update uhashlib's ones to support hmac (as per https://github.com/micropython/micropython/pull/4539) but hmac strictly doesn't need these extensions to function. So instead I think it's better to solve this by making the `hmac` module able to use the built-in ones as they are. See #515

mvo5:
Thanks @jimmo - using the buildin hashes by default of course makes a ton of sense - thank so much for fixing hmac with the buildin hashes too, really appreicated!

PR #437, opened 2021-07-23
aioble: Add support for write-with-update.

jimmo:
This allows a server to automatically notify/indicate subscribed clients on characteristic writes, implemented in ubluetooth in https://github.com/micropython/micropython/pull/7564

Alternative to #436

@andrewleech @MarceauFillon

PR #439, opened 2021-08-17
aioble updates - generic attribute service/service changed

andrewleech:
This MR includes a few changes in a working branch.
* https://github.com/micropython/micropython-lib/pull/437
* aioble: Add timeout to device.exchange_mtu()
* aioble/security: Store secrets in list. 
    - Ensure newest keys are stored at top to be used/loaded first.
* aioble/server: Log warning on out-of-order indication.
* aioble/service_changed: Add support for BLE service_changed. 

Service Changed is an important feature for any project where the BLE pairing&bonding is used and services/characteristics could be changed, either from modifying application settings, to rolling out in-field updates.

Most BLE clients (eg android/ios phones) will cache the service discovery when bonded to a device. This means that if these are later changed on the device, the phone doesn't know about it and will send the wrong data to a different characteristic.

The BLE service: `Generic Attribute Service` contains a characteristic that can be indicated immediately after connection to tell the client that some/all the services/characteristics have changed and should be re-discovered.

For newer (BLE5.1) clients, there's a separate Database Hash characteristic that contains a hash of all services/characteristics configuration which the client can check directly and check against its stored hash for any change, also prompting re-discovery.

Both are supported in this change, however it needs the existing default gatt service entry removed with https://github.com/micropython/micropython/pull/7675

### DIFF
--- a/micropython/bluetooth/aioble/aioble/device.py
+++ b/micropython/bluetooth/aioble/aioble/device.py
@@ -262,7 +262,7 @@ class DeviceConnection:
     def timeout(self, timeout_ms):
         return DeviceTimeout(self, timeout_ms)
 
-    async def exchange_mtu(self, mtu=None):
+    async def exchange_mtu(self, mtu=None, timeout_ms=1000):
         if not self.is_connected():
             raise ValueError("Not connected")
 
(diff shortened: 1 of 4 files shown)


andrewleech:
This change has been deprecated in favour of https://github.com/micropython/micropython/pull/7877

PR #440, opened 2021-08-21
WIP: Update deque

mattytrentini:
MicroPython's built-in deque lacks _many_ features but is implemented in C so is very lean and fast. 
The micropython-lib version is implemented in Python and has more features - but should have even more. This is a start toward achieving feature-parity with CPython's deque.

### DIFF
--- a/python-stdlib/collections.deque/collections/deque.py
+++ b/python-stdlib/collections.deque/collections/deque.py
@@ -1,27 +1,40 @@
 class deque:
-    def __init__(self, iterable=None):
+    def __init__(self, iterable=None, maxlen=None):
         if iterable is None:
             self.q = []
         else:
             self.q = list(iterable)
+        if maxlen is not None:
+            if not isinstance(maxlen, int):
+                raise TypeError("" "an integer is required")
+            if maxlen < 0:
+                raise ValueError("" "maxlen must be non-negative")
+        self.__maxlen = maxlen
 
     def popleft(self):
         return self.q.pop(0)
-
-    def popright(self):
-        return self.q.pop()
 
     def pop(self):
         return self.q.pop()
 
     def append(self, a):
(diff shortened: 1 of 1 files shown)


mattytrentini:
This could possibly be merged as-is (I had to extend deque to supporting porting the [progress library](https://github.com/mattytrentini/progress/tree/micropython-support) to MicroPython) but I'm going to try and add [all the other features of deque](https://docs.python.org/3/library/collections.html#collections.deque) as well as use as much as possible of the [CPython test cases](https://github.com/python/cpython/blob/main/Lib/test/test_deque.py).

python-stdlib/collections.deque/collections/deque.py
```diff
@@ -1,28 +1,44 @@
...
-    def popright(self):
-        return self.q.pop()
-
     def pop(self):
         return self.q.pop()
 
     def append(self, a):
         self.q.append(a)
+        if self.__maxlen is not None and len(self.q) > self.__maxlen:
+            self.popleft()
```
dpgeorge, line 23: you could do this bit before appending, to use less memory for the list

BrianPugh:
chiming in that this helped make some of my cpython code be micropython-compatible.

mattytrentini:
Note that the built-in deque has had some substantial improvements, see:

https://github.com/micropython/micropython/commit/7dff38fdc190d7b731fad8319d2ae8aa13fde18a

There is little reason to use this Python version now - which is great!

dpgeorge:
Now that the core `collections.deque` class is more fully implemented, I think we should just remove the `deque` implementation here.  That will prevent any confusion, everyone will just use the built-in version.

mattytrentini:
I agree; the differences are minimal, so I'll close this. Fantastic to see the built-in deque improved!

PR #446, opened 2021-09-09
uaescmac: New module for aes cypher and aes-cmac algorithm

MarceauFillon:
New module to provide aes cypher and aes-cmac algorithm.
This module is based on a pure python solution and adapted
to match the other micropython hashers methods.
This type of algorithm is used to compute the database hash
applied to Bluetooth services definition.

Signed-off-by: Marceau Fillon <marceau.fillon@planetinnovation.com.au>


dpgeorge:
Thanks for the contribution and sorry no one looked at this PR.

Does it still make sense to have it?  Or could it be mostly replaced by the `cryptolib` module (which implements AES in C)?

PR #447, opened 2021-09-16
pprint: add simple pprint(), pp(), and pformat() methods for simple container types

mikebentley15:
# Description

This change is to provide a simple implementation of pprint to replace the dummy placeholder implementation.  This implementation should not be too slow nor take too much space on disk nor memory (which is the point of Micropython).

This does not implement the full specification from CPython, but enough for simple uses.  It simply expands lists, sets, and dictionaries to multiple lines if there are more than one element.  Useful for simple printing of lists, dictionaries, and sets to the console, especially when using the REPL.

I added one additional function that is not in the CPython implementation: `ppdir()`.  This is a convenience function for calling `pprint(dir(obj))`, except by default, attributes that begin with an underscore will not be displayed (which can be changed by specifying `hidden=True`).  I personally find this useful when using the REPL, similar to how printing works by default in iPython.  If this additional convenience function is not wanted, it can be removed.

_Note: this is my first pull request for Micropython.  In fact, I've only been playing with it for about a week now.  Please be gentle :)_

### DIFF
--- a/python-stdlib/pprint/pprint.py
+++ b/python-stdlib/pprint/pprint.py
@@ -53,9 +53,13 @@ def pp(obj, *args, sort_dicts=False, **kwargs):
 def pprint(obj, stream=None, indent=1, depth=1, sort_dicts=True):
     """Simple implementation of a pretty-printer.
 
-    For simplicity, this does not recurse down.  It does not produce the same
-    output as the CPython implementation.  Instead, it is intended to simply
-    expand lists and dictionaries to multiple lines for easy viewing.
+    It does not produce the same output as the CPython implementation, but it's
+    close.  Instead, it is intended to simply expand lists and dictionaries to
+    multiple lines for easy viewing.  When the recursion depth has been
+    reached, the elements will be printed with repl(item) instead of the
+    CPython implementation of eliding the results (e.g., "{...}").  This is
+    because this implementation is not as safe as the CPython implementation
+    for deep recursion, which is detailed two paragraphs lower.
 
     Only expands containers.  Currently supported containers are set, list,
     dict, and tuple.
--- a/python-stdlib/pprint/pprint.py
+++ b/python-stdlib/pprint/pprint.py
@@ -1,6 +1,10 @@
-from io import StringIO as _StringIO
-from sys import stdout as _stdout
-from machine import const as _const
+import io as _io
+import sys as _sys
+
+try:
+    from micropython import const as _const
+except ImportError:
+    _const = lambda x: x
 
 
 class PrettyPrinter:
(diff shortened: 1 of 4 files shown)


python-stdlib/pprint/pprint.py
```diff
@@ -1,6 +1,57 @@
-def pformat(obj):
-    return repr(obj)
+from sys import stdout as _stdout
 
+def pformat(obj, indent=1, sort_dicts=True):
+    'Simply calls pprint() but capturing prints into a string and returning'
```
stinos, line 4: docstrings style should be like `"""Simply calls pprint()"""` or if multiline

```
"""Short description.

Long description here.
"""
```

python-stdlib/pprint/pprint.py
```diff
@@ -1,6 +1,57 @@
...
-    print(repr(obj))
+def pp(obj, *args, sort_dicts=False, **kwargs):
+    'Same as pprint() except the default value of sort_dicts is False'
+    pprint(obj, *args, sort_dicts=sort_dicts, **kwargs)
+
+def pprint(obj, stream=None, indent=1, sort_dicts=True):
+    '''
+    Simple implementation of a pretty-printer.
+
+    For simplicity, this does not recurse down.  It does not produce the same
```
stinos, line 18: Would it be possible to make it recurse anyway, as that's much more versatile? Shouldn't be too hard

mikebentley15: Perhaps not too hard.  The main problem is that the `pprint` function from the standard library has safety checks such as

- make sure not to overwhelm the output stream (by having a maximum width and meticulously checking the output and collapsing it)
- ensure no infinite recursion (such as a list being a member of itself or a descendant of itself)

There are other tricky things like ensuring the indentation is handled properly.  I do not intend to implement those checks at this time as they will add significant complexity.

If you think it would be good to have the option, then I could implement it to have a default of no recursion, but the user can specify a desired depth, at their own peril (if they're not careful) ;)

python-stdlib/pprint/pprint.py
mikebentley15: Done.

python-stdlib/pprint/pprint.py
stinos: I wouldn't care about the checks etc, but I mainly use pprint for looking at JSON data on the REPL (instead of e.g. opening in a text editor and pretty printing there) and I encounter nested data way more often than not. But that is just my experience.

mikebentley15: Okay, I've refactored to include depth recursion.  However, since safety checks are not performed, I set the default to a depth of one instead of infinite depth.

mikebentley15:
This pull request has been updated to not just handle a depth of 1, but handles any depth.  The types that are expanded are limited to

- `list`
- `tuple`
- `set`
- `dict`

Without checks, it is not generally safe to set the depth to infinite, so the default is 1 in this implementation.

I have also added the convenience class `PrettyPrinter` which provides the ability to cache parameters and forward them on to `pprint()` and `pformat()`.

dpgeorge (approved):
Thanks for the contribution and sorry this didn't get any further attention after the initial review.

It looks good, relatively simply and provides CPython functionality.

PR #448, opened 2021-09-23
aioble/security: Control order of bond database and allow limiting of the number of pairs.

andrewleech:
Requires: https://github.com/micropython/micropython/pull/7845

In my application, we have a stm327f765 chip running micropython, with a build of the official nimble uart/hci application running on a separate nrf52810 (https://github.com/apache/mynewt-nimble/tree/master/apps/blehci)

We use this with aioble and paring/bonding.

It's been found that this nimble hci radio only has buffer/storage for 4 RPA / IRK addressing lookups, so if you've got more devices than this stored in the pairing/bonding database, any new devices that come along and try to pair/bond seem to work initially, but upon disconnect/reconnect the pairing details are lost.
Also, if you've got more devices than this stored in the bond database, with the current inplementation of aioble the fact these are stored in a dict/json file means the order is not strictly controlled, so you can't know for certain which devices keys will be loaded first/last into the radio - which ever ones don't fit can fail to connect correctly.

This PR aims to control the order of keys to maintain knowledge of newest / oldest keys. 

A limit on the number of peers can be configured eg. `aioble.security.limit_peers = 3`
In this case, only the newest 3 peers will be loaded into the radio at startup.
If a new peer tries to pair/bond and this limit has already been reached, the oldest device in the database will be dropped / unpaired to make space for the new one.

### DIFF
--- a/micropython/bluetooth/aioble/aioble/device.py
+++ b/micropython/bluetooth/aioble/aioble/device.py
@@ -262,7 +262,7 @@ class DeviceConnection:
     def timeout(self, timeout_ms):
         return DeviceTimeout(self, timeout_ms)
 
-    async def exchange_mtu(self, mtu=None):
+    async def exchange_mtu(self, mtu=None, timeout_ms=1000):
         if not self.is_connected():
             raise ValueError("Not connected")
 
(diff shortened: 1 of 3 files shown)


andrewleech:
This PR has been split out into separate ones for each feature

PR #450, opened 2021-09-27
New module `suntime` for sunrise and sunset time

lorcap:
Approximated calculation of sunrise and sunset time. Adapted from https://github.com/SatAgro/suntime.


mattytrentini:
This looks like a useful addition, however [SatAgro/suntime](https://github.com/SatAgro/suntime) is LGPL which is incompatible with micropython-lib's MIT license...

lorcap:
I started from scratch with the formula provided by Wikipedia's [Sunrise equation](https://en.wikipedia.org/wiki/Sunrise_equation) and it works. I'll provide a new clean and neat code under MIT license in the following days.

lorcap:
I just pushed a commit with the new module implementation. I rebased it on `datetime` branch (see PR #449) because it depends on that.

lorcap:
Just rebased on top of `datetime`.

jonnor:
I do not see why this should be in micropython-lib, as opposed to being a module maintained as a third-party Github repository.
If one add a package.json then it is just as easy to install with mip etc.

Proposing to close.

lorcap:
I created a new [repository](/lorcap/micropython-suntime), updated the code to work with the latest `datetime` version and added the JSON files for mip as suggested.

This PR can be closed.

PR #451, opened 2021-09-30
Add chunk and aifc modules

mattytrentini:
Added the [chunk](https://github.com/python/cpython/blob/main/Lib/chunk.py) and [aifc](https://github.com/python/cpython/blob/main/Lib/aifc.py) modules from CPython with some minimal changes for MicroPython incompatibilities. Chunk is a dependency of aifc.

Note that the tests weren't ported since they use the CPython test infrastructure which is going to be challenging to run under MicroPython. A simple test for chunk was added. 

Unfortunately, some of the audio files that aifc can process requires an additional library, [audioop](https://github.com/python/cpython/blob/main/Modules/audioop.c). It's implemented in C and will require a rewrite (it may be a good candidate for a [native module](https://docs.micropython.org/en/latest/develop/natmod.html)). So, at least for now, if an attempt is made to load those particular audio files then a runtime exception will be raised ("Import Error: no module named audioop").

Successfully tested with stereo 8, 12, 16, 24, and 32 bit integer AIFF and AIFF-C files. Fails with alaw compressed data, ulaw compressed data, and float data.


mattytrentini:
Also, I presume the [Python license](https://github.com/python/cpython/blob/main/LICENSE) is compatible..?

mattytrentini:
[PEP 594](https://peps.python.org/pep-0594/) suggests that this library is to be deprecated so perhaps we should forget about this library...

dpgeorge:
Yes, agreed, let's not add libraries that have been deprecated by PEP 594.

PR #455, opened 2021-10-18
aioble: Add a write queue for gatt server.

jimmo:
This fixes a bug where an incoming write before `written` is awaited
causes `written` to return None. It also introduces a mechanism for
a server to "capture" all incoming written values (instead of only
having access to the most recent value).

### DIFF
--- a/micropython/bluetooth/aioble/aioble/server.py
+++ b/micropython/bluetooth/aioble/aioble/server.py
@@ -34,8 +34,13 @@ _FLAG_WRITE_ENCRYPTED = const(0x1000)
 _FLAG_WRITE_AUTHENTICATED = const(0x2000)
 _FLAG_WRITE_AUTHORIZED = const(0x4000)
 
+_FLAG_WRITE_CAPTURE = const(0x10000)
+
 _FLAG_DESC_READ = const(1)
 _FLAG_DESC_WRITE = const(2)
+
+
+_WRITE_CAPTURE_QUEUE_LIMIT = const(10)
 
 
 def _server_irq(event, data):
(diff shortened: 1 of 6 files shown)


jimmo:
Added multitest.

micropython/bluetooth/aioble/aioble/server.py
```diff
@@ -34,10 +34,15 @@
 _FLAG_WRITE_AUTHENTICATED = const(0x2000)
 _FLAG_WRITE_AUTHORIZED = const(0x4000)
 
+_FLAG_WRITE_CAPTURE = const(0x10000)
+
 _FLAG_DESC_READ = const(1)
 _FLAG_DESC_WRITE = const(2)
 
 
+_WRITE_CAPTURE_QUEUE_LIMIT = 10
```
dpgeorge, line 43: make it `const(10)`?

micropython/bluetooth/aioble/aioble/server.py
```diff
@@ -137,8 +179,13 @@ def __init__(
...
                 _FLAG_WRITE_NO_RESPONSE if write_no_response else 0
             )
-            self._write_connection = None
+            if capture:
+                # Capture means that we keep track of all writes, and capture
+                # their values (and connection) in a queue. Otherwise we just
+                # track the most recent connection.
+                flags |= _FLAG_WRITE_CAPTURE
             self._write_event = asyncio.ThreadSafeFlag()
+            self._write_queue = []
```
dpgeorge, line 188: did you consider using a `collections.deque` here?  it seems like that could be more efficient for append/popleft (although it would introduce a dependency on this class...)

dpgeorge:
I think `perf_gatt_notify.py` also needs to be fixed to use `aioble.stop()`.

micropython/bluetooth/aioble/aioble/server.py
jimmo: Yeah that's a good idea. All the BLE targets have collections.deque, so probably worth doing. It also simplifies the code a bit because the max-length is handled by the deque. I've done this in a separate commit, but feel free to squash.

jimmo:
> I think `perf_gatt_notify.py` also needs to be fixed to use `aioble.stop()`.

Done.

dpgeorge:
Merged in 5a86aa58662469c789effc7f76225889b6a969fe through 43cad179462d965014eaaa6567eabcae2c6f6f25 (with an additional fix to `perf_gatt_notify.py` to pass `connection` to `discover_server()`).

PR #459, opened 2021-11-08
aioble: Fix notified/indicated event waiting.

jimmo:
After a client does a successful `await char.notified()`, then before the
next call to `notified()` a notification arrives, then they call
`notified()` twice before the _next_ notification, the second call will
return None rather than waiting.

This applies the same fix as in 5a86aa58662469c789effc7f76225889b6a969fe
which solved a similar problem for server-side `char.written()`. Using
a deque is slightly overkill here, but it's consistent with the server
side, and also makes it very easy to support having a notification queue
in the future.

Also makes the client characteristic properly flags/properties-aware (i.e.
explicitly fail operations that aren't supported).

Signed-off-by: Jim Mussared <jim.mussared@gmail.com>

### DIFF
--- a/micropython/bluetooth/aioble/README.md
+++ b/micropython/bluetooth/aioble/README.md
@@ -3,7 +3,7 @@ aioble
 
 This library provides an object-oriented, asyncio-based wrapper for MicroPython's [ubluetooth](https://docs.micropython.org/en/latest/library/ubluetooth.html) API.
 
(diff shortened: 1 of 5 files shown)


jimmo:
Added multitest for this.

Also added multitest for subscription (so I've pulled in the commit from https://github.com/micropython/micropython-lib/pull/437).

dpgeorge:
Thank you!  The multitests pass for me on PYBD+PYBD.

PR #457, opened 2021-10-29
micropython/aioble: Add l2cap channel disconnected().

jimmo:
Allows `await channel.disconnected()`.

This also fixes a bug where connection._l2cap_channel wasn't being set to None on disconnect.

Also updates the README.md to show a minimal l2cap channel example.

PR #460, opened 2021-11-11
aioble: Add shutdown handler for cleanup.

jimmo:
This allows `aioble.stop()` to reset all internal state.

(The reported issue here was that after a shutdown, l2cap's internal `_listening` state was still true, so it never re-started accepting).

Note this includes the two commits from #457 as the test depends on it.

### DIFF
--- a/micropython/bluetooth/aioble/README.md
+++ b/micropython/bluetooth/aioble/README.md
@@ -123,6 +123,25 @@ while True:
     data = await temp_char.notified()
 ```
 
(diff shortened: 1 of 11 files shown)


dpgeorge:
Tested on PYBD+PYBD, works correctly.

Merged in 3ea74867f39bd42a10c7388c305a12bf59a42553 through cdd260f0792d04a1ded99171b4c7a2582b7856b4

PR #462, opened 2021-11-18
aioble/security: Only schedule save when needed.

andrewleech:
Was getting occasional: RuntimeError: schedule queue full if there was a flood of BLE activity.

This also consolidates the disk write for back-to-back security operations which are common during pairing.


dpgeorge:
Can probably remove the `if not _modified:` bit from `_save_secrets()`.

dpgeorge (approved):

PR #465, opened 2021-11-18
aioble: Add timeout to device.exchange_mtu()

andrewleech:
Similar to many other BLE operations that are waiting for a response; the timeout here can prevent the device locking up indefinitely waiting for a response to a failed command.


jimmo:
Thanks -- merged in `[09f0e47](https://github.com/micropython/micropython-lib/commit/09f0e47386c283b17e44cfeee0329165b654d6e2)`

Sorry about the delay on this... initially my thought was that the caller would opt-in to timeout by using the `with conn.timeout(...)` around the call to the various operations, but I've made having a `timeout_ms` arg the precedent in so many other places that this definitely makes sense to do it here too. And I think having a default timeout on all ops is a good thing.

PR #469, opened 2021-11-27
python-ecosys/urequests: Workaround for faulty getaddrinfo() with ".local" adresses on ESP32.

karfas:
On the ESP32, socket.getaddrinfo() might return SOCK_DGRAM instead of SOCK_STREAM.
As a HTTP request is always a TCP stream, we don't need to rely on the values returned by getaddrinfo.


andrewleech:
This looks like a worthwhile safe change. While normally I'd avoid workarounds in a library for a port-specific issue, as you say a HTTP request should always be set up with SOCK_STREAM so this change makes sense to me.

andrewleech (approved):

andrewleech:
Merged in a725c420 as part of #500.  
@karfas I updated the commit message to match the style of the project but maintained you as the author. Thanks for the change!

PR #470, opened 2021-12-06
aioble/server: Add missing import for GattError.

andrewleech:
GattError is used in `async def indicate()` but was missing from the import from `.core`


andrewleech:
Has been fixed airway already

PR #471, opened 2021-12-07
aioble/peripheral: Handle immediate disconnect.

andrewleech:
If the client disconnects immediately after connection, the irq can be run before the initial connect handler has finished.


puppet13th:
+1

jimmo:
Thanks @andrewleech 

I think maybe this could just be simplified by having `DeviceConnection::__init__` always unconditionally create the event. I'm not sure I can see any code paths now that don't end up initialising `self._event` anyway. i.e. construction of a `DeviceConnection` for either a central or a peripheral will always end up needing to initialise the event. So we'll always have to do this allocation regardless, might as well do it up-front.

Then we can get rid of all the various places that create-or-reuse the event...less code overall.

PR #476, opened 2022-01-04
aioble/peripheral.py: Reduce priority of name in the passive scan response.

mattytrentini:
A number of fields (services, appearance, manufacturer, name) can appear when a
scan is requested. However there is only so much space in the header so, if a
user has configured multiple fields, some may be 'pushed' into the active
scan response which requires additional communication.

name is, arguably, the least useful of the fields but was given the highest
priority and would 'bump' the others if configured. This change simple reduces
the priority of name and makes it less likely to push the more useful fields
into the active scan response.

### DIFF
--- a/micropython/bluetooth/aioble/aioble/peripheral.py
+++ b/micropython/bluetooth/aioble/aioble/peripheral.py
@@ -126,9 +126,6 @@ async def advertise(
             struct.pack("B", (0x01 if limited_disc else 0x02) + (0x18 if br_edr else 0x04)),
         )
 
-        if name:
-            resp_data = _append(adv_data, resp_data, _ADV_TYPE_NAME, name)
-
         if services:
             for uuid in services:
                 b = bytes(uuid)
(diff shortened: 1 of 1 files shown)


dpgeorge:
I think the ordering needs to be configurable.  Some projects may want `name` first, and some may want it last.

See #489 for an alternative.

micropython/bluetooth/aioble/aioble/peripheral.py
```diff
@@ -153,6 +150,9 @@ async def advertise(
                 struct.pack("<H", manufacturer[0]) + manufacturer[1],
             )
 
+        if name:
```
dpgeorge, line 153: Please move it to just after `services`, and add a comment, something like "Services are prioritised to go in the advertising data because iOS supports filtering scan results by service only, so service must come first".

dpgeorge: This change has been applied.

dpgeorge:
Merged in 0c31e0b3d712fcd21982daec8bfca48ad5527a72

PR #479, opened 2022-01-26
uaiohttpclient: add https support

eliclement:
Add https support like the urequests lib does.


andrewleech:
Does this PR work as-is? I would expect the `ussl` module would also need to be used to wrap the socket like https://github.com/micropython/micropython-lib/commit/d120749191851fb1e00c3632204908fce55694a9

jczic:
And wrap a socket with the uaio is special with the handshake...
Bad commits..

PR #483, opened 2022-02-21
aioble/aioble: Characteristic _tuple function now returns descriptor's tuples as well

chrisovergaauw:
I could not test the multitest and it is a copy of the characterstic test.
especially the instance1 task is not modified yet, but I don't know how to run the test.

If given some pointers, I could finish the multitest and update the PR.

### DIFF
--- a/micropython/bluetooth/aioble/aioble/server.py
+++ b/micropython/bluetooth/aioble/aioble/server.py
@@ -199,6 +199,9 @@ class Characteristic(BaseCharacteristic):
         self._value_handle = None
         self._initial = initial
 
+    def _tuple(self):
+        return (self.uuid, self.flags, tuple(d._tuple() for d in self.descriptors))
+
     def notify(self, connection, data=None):
         if not (self.flags & _FLAG_NOTIFY):
             raise ValueError("Not supported")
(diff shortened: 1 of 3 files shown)


andrewleech:
Hi, yeah the multitests would benefit from some docs :-)  it took me a while to figure out too - I should have written said docs when I got it running...

To start with, you'll need two BLE capable micropython devices plugged in. It's possible to use the unix port as one of the instances, however this requires a fair bit of extra (non-documented) setup, so it'll generally be much easier to use two physical devices (eg. pyb-d, stm32wb55, esp32)

Next, the multitest runner scripts are in the micropython repo, so you'll need a copy of that alongside micropython-lib

Then from linux (or wsl with usb devices passed through)
```
cd micropython/tests
python3 run-multitests.py -i pyb:/dev/ttyACM0  -i pyb:/dev/ttyACM1 ../../micropython-lib/micropython/bluetooth/aioble/multitests/*
```
or from windows 
```
cd micropython\tests
py.exe run-multitests.py -i pyb:COM12  -i pyb:COM17 ..\..\micropython-lib\micropython\bluetooth\aioble\multitests\ble_descriptor.py
```

`-i` is used to specify the two instances used for each end of the test, in this case `pyb` refers to using `pyboard.py` behind the scenes as the interface. The instance types can be seen in code here: https://github.com/micropython/micropython/blob/94a9b5066827d7dc5b1844ae28fad79778314185/tests/run-multitests.py#L502

dpgeorge:
Support for this was added in 765f14b501785e6f7bbf3b6be5827c48d52a3eb8

PR #487, opened 2022-03-17
Unittest updates from pycopy

andrewleech:
The unittest library has had a number of quality-of-life improvements from a couple of contributors on the pycopy fork of micropython-lib. 
This PR cherry picks / rebases them on current master, with black run over each commit interactivly, eg
```
git rebase -Xtheirs -i --exec 'black --fast --line-length=99 python-stdlib/unittest;git add python-stdlib/unittest; git commit --amend --no-edit' 4c98765e
```

For now, this keeps it largely in sync with https://pypi.org/project/micropython-unittest/

### DIFF
--- a/python-stdlib/unittest/metadata.txt
+++ b/python-stdlib/unittest/metadata.txt
@@ -1,3 +1,3 @@
 srctype = micropython-lib
 type = module
-version = 0.4.1
+version = 0.4.2
--- a/python-stdlib/unittest/metadata.txt
+++ b/python-stdlib/unittest/metadata.txt
@@ -1,3 +1,3 @@
 srctype = micropython-lib
 type = module
-version = 0.3.2
+version = 0.4
--- a/python-stdlib/unittest/metadata.txt
+++ b/python-stdlib/unittest/metadata.txt
@@ -1,3 +1,3 @@
 srctype = micropython-lib
 type = module
-version = 0.4.2
+version = 0.4.3
--- a/python-stdlib/unittest/metadata.txt
+++ b/python-stdlib/unittest/metadata.txt
@@ -1,3 +1,3 @@
 srctype = micropython-lib
 type = module
-version = 0.4
+version = 0.4.1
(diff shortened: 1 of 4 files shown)


stinos:
Good idea but I wonder whether we need to pull in all those no-code 'release xxx' commits?

andrewleech:
Yeah I probably should have marked this pr as draft as I haven't reviewed all the changes myself yet.

This was intended as a first step in a series of updates I have planned for unittest in the coming days; there's some bugs relating to test error handling / reporting then I want to add an implementation of https://docs.python.org/3/library/unittest.html#test-discovery

Before I went to far into that I wanted to check what had already been done on the other fork; turns out some of it was things I'd submitted myself back in the day.

On a more specific note though, even those version commits are making changes in the sense they're updating the reported version. While these changes are in sync with what happens to be on pypi I feel it makes sense to keep the version numbers in sync too? Once features diverge, then it's a different story.

stinos:
The idea of versions (as I perceive it begin used here) if to bump them to get new pypi packages indeed. So in this case I'd go for a bunch of commits improving unittest, then one single commit bumping version?

andrewleech:
Very true, the version updates are to push to pypi - which we can't do for this package here because the pypi name is already taken. That's a separate issue to deal with elsewhere though.

I'll review the rest of the changes here tomorrow either way - if some of them don't make sense to keep then I'll probably strip down / reorder everything here. However if everything looks relevant and worthwhile then I think it's less confusing to anyone coming back and comparing the two forks in the future; the more they match match commit by commit (within a particular package at least) the easier it is to review.

andrewleech:
The initial branch I pushed had some rebase errors which have now been fixed. I went through the commits one by one and I think all the code changes make sense and generally improve the compatibility with cpython/unittest. I've continued to build new features on top of this branch in #488

mattytrentini:
This looks like it's in good shape - and adds a lot of useful features.

dpgeorge:
Closing in favour of #488.

PR #491, opened 2022-03-21
os.path: Add uos compatibility.

andrewleech:
The os.access function is only provided on unix/ffi os module, so replace with basic os.stat call.
Everything else in this module is already compatible with on-device uos module.


andrewleech:
Superceded by https://github.com/micropython/micropython-lib/pull/543

PR #492, opened 2022-03-21
tools/codeformat: By default only check/update files modified on current git branch.

andrewleech:
The codeformat tool is a fantasitc addition to the repo, ensuring users can format their code to match standards.
However, it can be difficult to use when there are other files in the repository not currently matching the standard.

For developers, running over the entire repo can throw up a large list of changes in a local git which can lead to inclusion of unrelated changes in commits if they're added accidentally. 
If the files arg is used to trim down the list of files scanned, it runs a risk of missing some files they've modified.

In CI, it means that PR's can fail on codeformat for issues that aren't related to that PR.

This change adds a git query in the codeformat tool by default to only work on the list of files that have been modified in the current branch. 

This can still be overridden by the files arg to run over all files still, eg. `python3 tools/codeformat.py -v '**'`

### DIFF
--- a/tools/codeformat.py
+++ b/tools/codeformat.py
@@ -54,6 +54,10 @@ C_EXTS = (
 PY_EXTS = (".py",)
 
 
+MAIN_BRANCH = "master"
+BASE_BRANCH = os.environ.get("GITHUB_BASE_REF", MAIN_BRANCH)
+
+
 def list_files(paths, exclusions=None, prefix=""):
     files = set()
     for pattern in paths:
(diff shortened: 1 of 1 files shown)


dpgeorge:
Can we make it so that if you're on `master` then it checks everything by default?  Or does that defeat the purpose of this PR?

The reason is so that one can run `tools/codeformat.py` on a fresh clone and it should check everything, rather than having to remember to add `'**'` to check everything.

andrewleech:
> Can we make it so that if you're on `master` then it checks everything by default? Or does that defeat the purpose of this PR?

Sorry yes, I intended to make it work this way, I'll fix it now!

dpgeorge:
Looks like this is a lot more complicated now.... is it good for review?

andrewleech:
Yep the behaviour did get more complex, currently:
* If on clean master, scan entire repo.
* If on master but with local changes, just scan local changes .
* If on branch, scan both local changes and and files modified on branch.

It did take a couple of extra git commands to reliably detect if the current commit is on master branch or not (including if it's an older commit on master and/or a detached HEAD)

I've also been trying to think through the workflow of when I'm working on a feature, I'd often start with adding new files from a clean checkout of master, then want to get them properly codeformatted before committing them - hence the updates to address local changes. Otherwise it would need to be a "make new files", "commit them to new branch", "run codeformat", "ammend commit".

Otherwise yes I think it's good for review, it seems to be working correctly for me as per the above descriptions.

tools/codeformat.py
```diff
@@ -54,8 +54,13 @@
 PY_EXTS = (".py",)
 
 
+MAIN_BRANCH = "master"
+BASE_BRANCH = os.environ.get("GITHUB_BASE_REF", MAIN_BRANCH)
+
+
 def list_files(paths, exclusions=None, prefix=""):
     files = set()
+    print(paths)
```
dpgeorge, line 63: This looks like a stray addition.

tools/codeformat.py
```diff
@@ -105,12 +110,58 @@ def fixup_c(filename):
         assert not dedent_stack, filename
 
 
+def query_git_files(verbose):
+    def list_results(cmd):
```
dpgeorge, line 114: Maybe call it `cmd_result_set()`, to indicate it returns a set not a list?

andrewleech: Ah yes, it was previously a list... then I had some duplicates...

dpgeorge:
Thanks for updating, merged in ab1e6231dcfa031376e6b60e95ca24868311588d

I wouldn't want to make these changes to the main repo's version of `codeformat.py` just yet.  Wait to see how they go in everyday use.

andrewleech:
> I wouldn't want to make these changes to the main repo's version of `codeformat.py` just yet. Wait to see how they go in everyday use.

Thanks, yep I was going to suggest the same thing; use this repo as a bit of a test bed to see if it helps and is reliable / doesn't let problems slip through!

PR #493, opened 2022-03-23
tools/codeformat: Fix compatibility running from subfolder in repo.

andrewleech:
Didn't take me long to find more changes in this script...

With the version on master, if run from a subfolder in the repo it will try to process any local changes with a path relative to root of repo - which will fail.
It was also only processing files committed on the current branch that were also within the current folder.

Both of these issues have been fixed here to correctly process all expected files, with all paths correctly relative to the current directory.


dpgeorge:
Thank you!

PR #494, opened 2022-04-05
Supports class level setUp and Teardown

steveli777:
To support class level setUp and Teardown for Unittest

### DIFF
--- a/python-stdlib/unittest/unittest.py
+++ b/python-stdlib/unittest/unittest.py
@@ -188,8 +188,13 @@ class TestResult:
 # TODO: Uncompliant
 def run_class(c, test_result):
     o = c()
+    set_up_class = getattr(o, "setUpClass", lambda: None)
+    tear_down_class = getattr(o, "tearDownClass", lambda: None)
     set_up = getattr(o, "setUp", lambda: None)
     tear_down = getattr(o, "tearDown", lambda: None)
+
+    set_up_class()
+
     for name in dir(o):
         if name.startswith("test"):
             print("%s (%s) ..." % (name, c.__qualname__), end="")
@@ -210,6 +215,8 @@ def run_class(c, test_result):
                 continue
             finally:
                 tear_down()
+    
+    tear_down_class()
 
 
 def main(module="__main__"):


andrewleech:
Hi @steveli777 thanks for the contribution! As a general rule (not yet documented anywhere easy to find) the commits here should follow a pattern of `module: change description`, for example take a look at the commits in https://github.com/micropython/micropython-lib/pull/488/commits

python-stdlib/unittest/unittest.py
```diff
@@ -188,8 +188,14 @@ def wasSuccessful(self):
 # TODO: Uncompliant
 def run_class(c, test_result):
     o = c()
+    set_up_class = getattr(o, "setUpClass", lambda: None)
+    tear_down_class = getattr(o, "tearDownClass", lambda: None)
     set_up = getattr(o, "setUp", lambda: None)
     tear_down = getattr(o, "tearDown", lambda: None)
+
+    if set_up_class:
+        set_up_class()
```
andrewleech, line 197: In the `getattr` line above, if the function doesn't exist it defaults to `lambda: None`. This is done so that you don't need the `if set_up_class` here, you can just run `set_up_class()` and if it doesn't exist it'll just run the no-op lambda function

python-stdlib/unittest/unittest.py
```diff
@@ -198,9 +204,9 @@ def run_class(c, test_result):
             try:
                 test_result.testsRun += 1
                 m()
-                print(" ok")
+                print(" PASS")
             except SkipTest as e:
-                print(" skipped:", e.args[0])
+                print(" SKIPPED:", e.args[0])
```
andrewleech, line 209: These output format changes should be done in a separate commit - but only if the new format matches cpython unittest? I'd prefer to add this commit after the #488 changeset which has had some other changes made to the output format already to more closely match cpython.

python-stdlib/unittest/unittest.py
steveli777: Thanks for the suggestion, fixed.

python-stdlib/unittest/unittest.py
steveli777: You are right, I reverted those text changes.

mattytrentini (approved):
Clean and simple, nice!

andrewleech:
Hi @steveli777 as I've got a running unittest PR on the go, I've rebased your commit here on top of that one. 

It needed some manual intervention to deal with merge conflicts, but I think I got it right: https://github.com/micropython/micropython-lib/pull/488/commits/23866f3848e3f15878d0a69954268a81d96e4b1f

Even with the merge changes the commit is still comitted in your name, I hope that's good for you!

If you get a chance could you review / test this setUpClass functionality in the branch from https://github.com/micropython/micropython-lib/pull/488 ?

Otherwise do you have a simple test case I can run (or add as unit test) to cover this functionality?

dpgeorge:
This feature was implemented in 2d61dbdb93fa941c54b3990c2ee046074d2fce07

PR #496, opened 2022-04-05
aioble/examples/temp_client.py: check connection before reading temp

rknegjens:
Only read from the temp characteristic if the connection is still
active. Improves the example by avoiding a TypeError exception when the
sensor disconnects.


dpgeorge:
Sorry this took 2 years to get to!

Now merged.

PR #497, opened 2022-04-06
aioble/device.py: No default timeout for async disconnected() method.

rknegjens:
The value for the `timeout_ms` optional argument to `DeviceConnection.disconnected()` async method is changed from 60000 to None. This way users awaiting a device disconnection using `await connection.disconnected()` won't be surprised by a 1 minute timeout. For instance, in the `temp_sensor.py` usage example I think it's reasonable to expect the connection to the client to stay open indefinitely.

### DIFF
--- a/micropython/bluetooth/aioble-core/manifest.py
+++ b/micropython/bluetooth/aioble-core/manifest.py
@@ -1,4 +1,4 @@
-metadata(version="0.2.0")
+metadata(version="0.3.0")
 
 package(
     "aioble",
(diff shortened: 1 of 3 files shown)


andrewleech:
This does make sense to me, I know a couple of other people this timeout has tripped up too.

I'm not sure if there was a specific intention behind having a timeout here, can you comment @jimmo ?

andrewleech (approved):

peterson79:
This also tripped me up and caused a lot of debug time... can this get merged?

dpgeorge:
I agree, the default makes more sense as `None`, with no timeout.

In fact, the temp sensor example was just changed to explicitly specify this: d4362d5cc3a6d90fabc9a684e1e7c5a29cc47f6e

dpgeorge:
Merged!

PR #501, opened 2022-06-29
micropython/lora: Add async-capable LoRa modem driver

projectgus:
Working LoRa drivers for:

* SX1276, SX1277, SX1278, SX1279 modems
* SX1261, SX1262 modems.

Eventual goal (not in this PR) is to have enough support for MicroPython to be a LoRaWAN Class A device with a fully async API available.

## How to Test this PR

The [lora README added in the PR branch](https://github.com/projectgus/micropython-lib/blob/feature/lora/micropython/lora/README.md) has instructions for how to use the LoRa drivers.

However, until the PR is merged you need some additional steps to install the lora MicroPython packages.

### Installing with mip

The `--index` argument of mip install is [needed to install packages from a PR branch](https://github.com/micropython/micropython-lib/#installing-packages-from-forks).

Example:

```sh
$ mpremote connect /dev/ttyUSB0 mip install --index https://projectgus.github.io/micropython-lib/mip/feature/lora lora-sx127x lora-sync
```

(Replace `/dev/ttyUSB0` with the serial port you are using, and replace `lora-sx127x` and `lora-sync` with the appropriate packages for your LoRa hardware and programming style  - see the [lora README]((https://github.com/projectgus/micropython-lib/blob/feature/lora/micropython/lora/README.md)) for explanation.)

Or from a REPL:

```python
import mip
mip.install("lora-sync", index="https://projectgus.github.io/micropython-lib/mip/feature/lora")
```

(Replace the package name as needed.)

### Freezing lora into the firmware

If your board doesn't support mip, you can freeze the lora modules into the firmware.

* First, you need to be able to build the MicroPython firmware for your board locally, with the micropython source cloned from the github repo.
* Next, switch the branch of the `micropython-lib` git submodule, as found in the `lib/micropython-lib` directory of the micropython source, to the branch for this PR. [GitHub CLI client](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally?tool=cli) is the easiest way to check out a PR branch from the command line: Running commands similar to `cd lib/micropython; gh pr checkout 501` should do it.
* Update the board [manifest file](https://docs.micropython.org/en/latest/reference/manifest.html) to add the lora packages you need.
* Rebuild the firmware.

## TODOs

* [x] Examples are currently hardcoded to the LoRa modems I was using for testing. I think the best solution here is to just make them say `modem =   # TODO put your modem config here`, but haven't made this change yet.
* [x] Confirm the API & AsyncModem wrapper design is a reasonable approach
* [x] Support ThreadSafeFlag.clear() in MP
* [x] Figure out why the RxTimeout interrupt line is toggling during TX (currently using `_irq_en` as a workaround)
* [x] Split out the single configure() function to allow fast runtime configuration of parameters changed on the fly by the LoRAWAN MAC layer. (Done by having configure() keep the current value for any missing keys, instead of setting defaults.)
* [x] Decide about support for interleaving rx,tx as managed by the driver - i.e. it would be nice if one task could be continuously (or arbitrarily) calling `receive()` while another task sometimes called `transmit()`, and have the modem driver handle the state changes internally without the caller needing to do anything. This isn't necessary for LoRAWAN, but it would make some other types of application logic much simpler. (Implemented.)
* [x] Decide about support for `rx_continuous` mode. If the previous dot point can be implemented then it might not be needed to explicitly support putting the modem into continuous RX mode. (Implemented.)
* [x] Review use of `assert` in the code - some places are probably ValueErrors, some places may be unnecessary.
* [x] Add support for a second modem apart from SX1276. (Done, SX1262.)
* [x] Resolve issue where SX1276 invert_iq_tx setting seems to be inverted from all other modems (this may be a hardware compatibility issue, but it's unusual!)
* [x] Check TODOs in the code, mostly these are things that could be configurable but currently aren't.
* [x] Proper packaging layout
* [x] Documentation

## Follow-up work

Intended for future PRs:

* :heavy_check_mark:  ~~Document the `ant_sw` feature that allows callbacks for switching off-modem antenna switches for RX, TX.~~
* Document `calibrate()` and `calibrate_image()` functions.
* :heavy_check_mark: ~~Document the modem `crc_errors` counter field, and maybe break out CRC checking to its own doc section.~~
* Add some packages for common modules and boards (some have fixed pins, others have specific antenna switch configs). This will make it a lot easier to get up and running with a LoRa development board or module.
* :heavy_check_mark: ~~SX126x radios with DIO2 configured as "enable TCXO" line currently take around 18ms to start the TCXO, which can be made configurable.~~ We can also add a configuration option to not automatically stop the TCXO (and go to "RC Standby") when a receive or transmit ends, which allows tighter receive/transmit timing.
* STM32WL5x is a special case of the SX126x which requires hardware support in MicroPython, have moved to #572.
* OCP limit is not currently configured either automatically or manually by the driver, and possibly should be.
* Add a function to read the instantaneous RSSI at any time during receive.
* Add a documentation section about common hardware issues - clock drift, symptoms of incorrect antenna pin setting, known modem bugs, etc.

## Example API usage

This PR adds two examples, each with synchronous and async API versions:

* `simple_rxtx` example that just transmits and receives in a loop
* `reliable_delivery` which implements a relatively simple "ACK or retry" scheme to receive packets reliably from low power nodes.

### DIFF
--- a/micropython/lora/lora-async/lora/async_modem.py
+++ b/micropython/lora/lora-async/lora/async_modem.py
@@ -1,0 +1,144 @@
+# MicroPython LoRa async modem driver
+# MIT license; Copyright (c) 2023 Angus Gratton
+#
+# LoRa is a registered trademark or service mark of Semtech Corporation or its affiliates.
+
+import asyncio
+import time
+from micropython import const
+
+# Set to True to get some additional printed debug output.
+_DEBUG = const(False)
+
+# Some "belts and braces" timeouts when using IRQs, to wake up if an IRQ hasn't
+# fired as expected. Also a timeout for how rapidly to poll the modem during TX
+# if no IRQ is enabled.
+#
+# All in milliseconds
+_RX_POLL_WITH_IRQ = const(1000)
+_RX_POLL_NO_IRQ = const(50)
+_TX_POLL_LATE_IRQ = const(10)
+
+
+class AsyncModem:
+    # Mixin-like base class that provides coroutines for modem send, recv & recv_continuous.
+    #
+    # Don't instantiate this class directly, instantiate one of the 'AsyncXYZ'
+    # modem classes defined in the lora module.
+
+    def _after_init(self):
+        # Separate flags for (rx, tx) as only one task can block on a flag
+        self._flags = (asyncio.ThreadSafeFlag(), asyncio.ThreadSafeFlag())
+        self.set_irq_callback(self._callback)
+
+    async def recv(self, timeout_ms=None, rx_length=0xFF, rx_packet=None):
+        # Async function to receive a single LoRa packet, with an optional timeout
+        #
+        # Same API as the "Simple API" synchronous function
+        self._flags[0].clear()
+        will_irq = self.start_recv(timeout_ms, False, rx_length)
+        return await self._recv(will_irq, rx_packet)
+
+    def recv_continuous(self, rx_length=0xFF, rx_packet=None):
+        # Returns an Async Iterator that continuously yields LoRa packets, until
+        # the modem is told to sleep() or standby().
+        #
+        # Once MicroPython has PEP525 support (PR #6668) then this somewhat
+        # fiddly implementation can probably be merged with recv() into a much simpler
+        # single _recv() private coro that either yields or returns, depending on
+        # an argument. The public API can stay the same.
+        will_irq = self.start_recv(None, True, rx_length)
+        return AsyncContinuousReceiver(self, will_irq, rx_packet)
+
+    async def _recv(self, will_irq, rx_packet):
+        # Second half of receiving is implemented here to share code with AsyncContinuousReceiver,
+        # until future refactor is possible (see comment immediately above this one.)
+        rx = True
+        while rx is True:
+            await self._wait(will_irq, 0, _RX_POLL_WITH_IRQ if will_irq else _RX_POLL_NO_IRQ)
+            rx = self.poll_recv(rx_packet)
+        if rx:  # RxPacket instance
+            return rx
+
+    async def send(self, packet, tx_at_ms=None):
+        self._flags[1].clear()
+        self.prepare_send(packet)
+
+        timeout_ms = self.get_time_on_air_us(len(packet)) // 1000 + 50
+
+        if tx_at_ms is not None:
+            await asyncio.sleep_ms(max(0, time.ticks_diff(tx_at_ms, time.ticks_ms())))
+
+        if _DEBUG:
+            print("start_send")
+
+        will_irq = self.start_send()
+
+        tx = True
+        while tx is True:
+            await self._wait(will_irq, 1, timeout_ms)
+            tx = self.poll_send()
+
+            if _DEBUG:
+                print(f"poll_send returned tx={tx}")
+
+            # If we've already waited the estimated send time (plus 50ms) and the modem
+            # is not done, timeout and poll more rapidly from here on (unsure if
+            # this is necessary, outside of a serious bug, but doesn't hurt.)
+            timeout_ms = _TX_POLL_LATE_IRQ
+
+        return tx
+
+    async def _wait(self, will_irq, idx, timeout_ms):
+        # Common code path for blocking on an interrupt, if configured.
+        #
+        # idx is into _flags tuple and is 0 for rx and 1 for tx
+        #
+        # timeout_ms is the expected send time for sends, or a reasonable
+        # polling timeout for receives.
+        if _DEBUG:
+            print(f"wait will_irq={will_irq} timeout_ms={timeout_ms}")
+        if will_irq:
+            # In theory we don't need to ever timeout on the flag as the ISR will always
+            # fire, but this gives a workaround for possible race conditions or dropped interrupts.
+            try:
+                await asyncio.wait_for_ms(self._flags[idx].wait(), timeout_ms)
+            except asyncio.TimeoutError:
+                pass
+        else:
+            await asyncio.sleep_ms(timeout_ms)
+
+        if _DEBUG:
+            print(f"wait complete")
+
+    def _callback(self, _):
+        # IRQ callback from BaseModem._radio_isr. Hard IRQ context unless _DEBUG
+        # is on.
+        #
+        # Set both RX & TX flag. This isn't necessary for "real" interrupts, but may be necessary
+        # to wake both for the case of a "soft" interrupt triggered by sleep() or standby(), where
+        # both tasks need to unblock and return.
+        #
+        # The poll_recv() method is a no-op if _tx is set, so waking up both
+        # blocked tasks doesn't have much overhead (no second polling of modem
+        # IRQ flags, for example).
+        for f in self._flags:
+            f.set()
+
+
+class AsyncContinuousReceiver:
+    # Stop-gap async iterator implementation until PEP525 support comes in, see
+    # longer comment  on AsyncModem.recv_continuous() above.
+    def __init__(self, modem, will_irq, rx_packet):
+        self.modem = modem
+        self.will_irq = will_irq
+        self.rx_packet = rx_packet
+
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self):
+        res = await self.modem._recv(self.will_irq, self.rx_packet)
+        if not res:
+            raise StopAsyncIteration
+        return res
--- a/micropython/lora/lora/lora/modem.py
+++ b/micropython/lora/lora/lora/modem.py
@@ -1,0 +1,483 @@
+# MicroPython LoRa modem driver base class
+# MIT license; Copyright (c) 2023 Angus Gratton
+#
+# LoRa is a registered trademark or service mark of Semtech Corporation or its affiliates.
+import time
+from micropython import const, schedule
+
+# Set to True to get some additional printed debug output.
+_DEBUG = const(False)
+
+
+def _clamp(v, vmin, vmax):
+    # Small utility function to clamp a value 'v' between 'vmin' and 'vmax', inclusive.
+    return min(max(vmin, v), vmax)
+
+
+def _flag(value, condition):
+    # Small utility function for returning a bit 'value' or not, based on a
+    # boolean condition. Can help make expressions to build register values more
+    # readable.
+    #
+    # Note that for value==1, can also rely on int(bool(x)) with one or both
+    # conversions being implicit, as int(True)==1 and int(False)==0
+    #
+    # There is also (condition and value) but this is (IMO) confusing to read.
+    return value if condition else 0
+
+
+class ConfigError(ValueError):
+    # Raise if there is an error in lora_cfg, saves some duplicated strings
+    def __init__(self, field):
+        super().__init__("Invalid lora_cfg {}".format(field))
+
+
+class BaseModem:
+    def __init__(self, ant_sw):
+        self._ant_sw = ant_sw
+        self._irq_callback = None
+
+        # Common configuration settings that need to be tracked by all modem drivers
+        # (Note that subclasses may set these to other values in their constructors, to match
+        # the power-on-reset configuration of a particular modem.)
+        #
+        self._rf_freq_hz = 0  # Needs to be set via configure()
+        self._sf = 7  # Spreading factor
+        self._bw_hz = 125000  # Reset value
+        self._coding_rate = 5
+        self._crc_en = True  # use packet CRCs
+        self._implicit_header = False  # implict vs explicit header mode
+        self._preamble_len = 12
+        self._coding_rate = 5
+
+        # CRC error counter
+        self.crc_errors = 0
+        self.rx_crc_error = False
+
+        # Current state of the modem
+
+        # _rx holds radio recv state:
+        #
+        # - False if the radio is not receiving
+        # - True if the radio is continuously receiving, or performing a single receive with
+        #   no timeout.
+        # - An int if there is a timeout set, in which case it is the is the receive deadline
+        #   (as a time.ticks_ms() timestamp).
+        #
+        # Note that self._rx can be not-False even when the radio hardware is not actually
+        # receiving, if self._tx is True (send always pauses recv.)
+        self._rx = False
+
+        # _rx_continuous is True if the modem is in continuous receive mode
+        # (this value is only valid when self._rx is also True).
+        self._rx_continuous = False
+
+        # This argument is stored from the parameter of the same name, as set in
+        # the last call to start_recv()
+        self._rx_length = None
+
+        # _tx holds radio send state and is simpler, True means sending and
+        # False means not sending.
+        self._tx = False
+
+        # timestamp (as time.ticks_ms() result) of last IRQ event
+        self._last_irq = None
+
+        # values are:
+        # - lora_cfg["invert_iq_rx"]
+        # - lora_cfg["invert_iq_tx"]
+        # - Current modem Invert setting
+        self._invert_iq = [False, False, False]
+
+        # This hook exists to allow the SyncModem & AsyncModem "mixin-like"
+        # classes to have some of their own state, without needing to manage the
+        # fuss of multiple constructor paths.
+        try:
+            self._after_init()
+        except AttributeError:
+            # If this exception happens here then one of the modem classes without a SyncModem or AsyncModem "mixin-like" class
+            # has been instantiated.
+            raise NotImplementedError(
+                "Don't instantiate this class directly, "
+                "instantiate a class from the 'lora' package"
+            )
+
+    def standby(self):
+        # Put the modem into standby. Can be used to cancel a continuous recv,
+        # or cancel a send before it completes.
+        #
+        # Calls the private function which actually sets the mode to standby, and then
+        # clears all the driver's state flags.
+        #
+        # Note this is also called before going to sleep(), to save on duplicated code.
+        self._standby()
+        self._rx = False
+        self._tx = False
+        self._last_irq = None
+        if self._ant_sw:
+            self._ant_sw.idle()
+        self._radio_isr(None)  # "soft ISR"
+
+    def _get_t_sym_us(self):
+        # Return length of a symbol in microseconds
+        return 1000_000 * (1 << self._sf) // self._bw_hz
+
+    def _get_ldr_en(self):
+        # Return true if Low Data Rate should be enabled
+        #
+        # The calculation in get_n_symbols_x4() relies on this being the same logic applied
+        # in the modem configuration routines.
+        return self._get_t_sym_us() >= 16000
+
+    def _get_pa_ramp_val(self, lora_cfg, supported):
+        # Return the PA ramp register index from the list of supported PA ramp
+        # values. If the requested ramp time is supported by the modem, round up
+        # to the next supported value.
+        #
+        # 'supported' is the list of supported ramp times, must be sorted
+        # already.
+        us = int(lora_cfg["pa_ramp_us"])
+
+        # Find the index of the lowest supported ramp time that is longer or the
+        # same value as 'us'
+        for i, v in enumerate(supported):
+            if v >= us:
+                return i
+        # The request ramp time is longer than all this modem's supported ramp times
+        raise ConfigError("pa_ramp_us")
+
+    def _symbol_offsets(self):
+        # Called from get_time_on_air_us().
+        #
+        # This function provides a way to implement the different SF5 and SF6 in SX126x,
+        # by returning two offsets: one for the overall number of symbols, and one for the
+        # number of bits used to calculate the symbol length of the payload.
+        return (0, 0)
+
+    def get_n_symbols_x4(self, payload_len):
+        # Get the number of symbols in a packet (Time-on-Air) for the current
+        # configured modem settings and the provided payload length in bytes.
+        #
+        # Result is in units of "symbols times 4" as there is a fractional term
+        # in the equation, and we want to limit ourselves to integer arithmetic.
+        #
+        # References are:
+        # - SX1261/2 DS 6.1.4 "LoRa Time-on-Air"
+        # - SX1276 DS 4.1.1 "Time on air"
+        #
+        # Note the two datasheets give the same information in different
+        # ways. SX1261/62 DS is (IMO) clearer, so this function is based on that
+        # formula. The result is equivalent to the datasheet value "Nsymbol",
+        # times 4.
+        #
+        # Note also there are unit tests for this function in tests/test_time_on_air.py,
+        # and that it's been optimised a bit for code size (with impact on readability)
+
+        # Account for a minor difference between SX126x and SX127x: they have
+        # incompatible SF 5 & 6 modes.
+        #
+        # In SX126x when using SF5 or SF6, we apply an offset of +2 symbols to
+        # the overall preamble symbol count (s_o), and an offset of -8 to the
+        # payload bit length (b_o).
+        s_o, b_o = self._symbol_offsets()
+
+        # calculate the bit length of the payload
+        #
+        # This is the part inside the max(...,0) in the datasheet
+        bits = (
+            # payload_bytes
+            8 * payload_len
+            # N_bit_crc
+            + (16 if self._crc_en else 0)
+            # (4 * SF)
+            - (4 * self._sf)
+            # +8 for most modes, except SF5/6 on SX126x where b_o == -8 so these two cancel out
+            + 8
+            + b_o
+            # N_symbol_header
+            + (0 if self._implicit_header else 20)
+        )
+        bits = max(bits, 0)
+
+        # "Bits per symbol" denominator is either (4 * SF) or (4 * (SF -2))
+        # depending on Low Data Rate Optimization
+        bps = (self._sf - (2 * self._get_ldr_en())) * 4
+
+        return (
+            # Fixed preamble portion (4.25), times 4
+            17
+            # Remainder of equation is an integer number of symbols, times 4
+            + 4
+            * (
+                # configured preamble length
+                self._preamble_len
+                +
+                # optional extra preamble symbols (4.25+2=6.25 for SX1262 SF5,SF6)
+                s_o
+                +
+                # 8 symbol constant overhead
+                8
+                +
+                # Payload symbol length
+                # (this is the term "ceil(bits / 4 * SF) * (CR + 4)" in the datasheet
+                ((bits + bps - 1) // bps) * self._coding_rate
+            )
+        )
+
+    def get_time_on_air_us(self, payload_len):
+        # Return the "Time on Air" in microseconds for a particular
+        # payload length and the current configured modem settings.
+        return self._get_t_sym_us() * self.get_n_symbols_x4(payload_len) // 4
+
+    # Modem ISR routines
+    #
+    # ISR implementation is relatively simple, just exists to signal an optional
+    # callback, record a timestamp, and wake up the hardware if
+    # needed. ppplication code is expected to call poll_send() or
+    # poll_recv() as applicable in order to confirm the modem state.
+    #
+    # This is a MP hard irq in some configurations, meaning no memory allocation is possible.
+    #
+    # 'pin' may also be None if this is a "soft" IRQ triggered after a receive
+    # timed out during a send (meaning no receive IRQ will fire, but the
+    # receiver should wake up and move on anyhow.)
+    def _radio_isr(self, pin):
+        self._last_irq = time.ticks_ms()
+        if self._irq_callback:
+            self._irq_callback(pin)
+        if _DEBUG:
+            # Note: this may cause a MemoryError and fail if _DEBUG is enabled in this base class
+            # but disabled in the subclass, meaning this is a hard irq handler
+            try:
+                print("_radio_isr pin={}".format(pin))
+            except MemoryError:
+                pass
+
+    def irq_triggered(self):
+        # Returns True if the ISR has executed since the last time a send or a receive
+        # started
+        return self._last_irq is not None
+
+    def set_irq_callback(self, callback):
+        # Set a function to be called from the radio ISR
+        #
+        # This is used by the AsyncModem implementation, but can be called in
+        # other circumstances to implement custom ISR logic.
+        #
+        # Note that callback may be called in hard ISR context, meaning no
+        # memory allocation is possible.
+        self._irq_callback = callback
+
+    def _get_last_irq(self):
+        # Return the _last_irq timestamp if set by an ISR, or the
+        # current time.time_ms() timestamp otherwise.
+        if self._last_irq is None:
+            return time.ticks_ms()
+        return self._last_irq
+
+    # Common parts of receive API
+
+    def start_recv(self, timeout_ms=None, continuous=False, rx_length=0xFF):
+        # Start receiving.
+        #
+        # Part of common low-level modem API, see README.md for usage.
+        if continuous and timeout_ms is not None:
+            raise ValueError()  # these two options are mutually exclusive
+
+        if timeout_ms is not None:
+            self._rx = time.ticks_add(time.ticks_ms(), timeout_ms)
+        else:
+            self._rx = True
+
+        self._rx_continuous = continuous
+        self._rx_length = rx_length
+
+        if self._ant_sw and not self._tx:
+            # this is guarded on 'not self._tx' as the subclass will not immediately
+            # start receiving if a send is in progress.
+            self._ant_sw.rx()
+
+    def poll_recv(self, rx_packet=None):
+        # Should be called while a receive is in progress:
+        #
+        # Part of common low-level modem API, see README.md for usage.
+        #
+        # This function may alter the state of the modem - it will clear
+        # RX interrupts, it may read out a packet from the FIFO, and it
+        # may resume receiving if the modem has gone to standby but receive
+        # should resume.
+
+        if self._rx is False:
+            # Not actually receiving...
+            return False
+
+        if self._tx:
+            # Actually sending, this has to complete before we
+            # resume receiving, but we'll indicate that we are still receiving.
+            #
+            # (It's not harmful to fall through here and check flags anyhow, but
+            # it is a little wasteful if an interrupt has just triggered
+            # poll_send() as well.)
+            return True
+
+        packet = None
+
+        flags = self._get_irq()
+
+        if _DEBUG and flags:
+            print("RX flags {:#x}".format(flags))
+        if flags & self._IRQ_RX_COMPLETE:
+            # There is a small potential for race conditions here in continuous
+            # RX mode. If packets are received rapidly and the call to this
+            # function delayed, then a ValidHeader interrupt (for example) might
+            # have already set for a second packet which is being received now,
+            # and clearing it will mark the second packet as invalid.
+            #
+            # However it's necessary in continuous mode as interrupt flags don't
+            # self-clear in the modem otherwise (for example, if a CRC error IRQ
+            # bit sets then it stays set on the next packet, even if that packet
+            # has a valid CRC.)
+            self._clear_irq(flags)
+            ok = self._rx_flags_success(flags)
+            if not ok:
+                # If a non-valid receive happened, increment the CRC error counter
+                self.crc_errors += 1
+            if ok or self.rx_crc_error:
+                # Successfully received a valid packet (or configured to return all packets)
+                packet = self._read_packet(rx_packet, flags)
+                if not self._rx_continuous:
+                    # Done receiving now
+                    self._end_recv()
+
+        # _check_recv() will return True if a receive is ongoing and hasn't timed out,
+        # and also manages resuming any modem receive if needed
+        #
+        # We need to always call check_recv(), but if we received a packet then this is what
+        # we should return to the caller.
+        res = self._check_recv()
+        return packet or res
+
+    def _end_recv(self):
+        # Utility function to clear the receive state
+        self._rx = False
+        if self._ant_sw:
+            self._ant_sw.idle()
+
+    def _check_recv(self):
+        # Internal function to automatically call start_recv()
+        # again if a receive has been interrupted and the host
+        # needs to start it again.
+        #
+        # Return True if modem is still receiving (or sending, but will
+        # resume receiving after send finishes).
+
+        if not self._rx:
+            return False  # Not receiving, nothing to do
+
+        if not self.is_idle():
+            return True  # Radio is already sending or receiving
+
+        rx = self._rx
+
+        timeout_ms = None
+        if isinstance(rx, int):  # timeout is set
+            timeout_ms = time.ticks_diff(rx, time.ticks_ms())
+            if timeout_ms <= 0:
+                # Timed out in software, nothing to resume
+                self._end_recv()
+                if _DEBUG:
+                    print("Timed out in software timeout_ms={}".format(timeout_ms))
+                schedule(
+                    self._radio_isr, None
+                )  # "soft irq" to unblock anything waiting on the interrupt event
+                return False
+
+        if _DEBUG:
+            print(
+                "Resuming receive timeout_ms={} continuous={} rx_length={}".format(
+                    timeout_ms, self._rx_continuous, self._rx_length
+                )
+            )
+
+        self.start_recv(timeout_ms, self._rx_continuous, self._rx_length)
+
+        # restore the previous version of _rx so ticks_ms deadline can't
+        # slowly creep forward each time this happens
+        self._rx = rx
+
+        return True
+
+    # Common parts of send API
+
+    def poll_send(self):
+        # Check the ongoing send state.
+        #
+        # Returns one of:
+        #
+        # - True if a send is ongoing and the caller
+        #   should call again.
+        # - False if no send is ongoing.
+        # - An int value exactly one time per transmission, the first time
+        #   poll_send() is called after a send ends. In this case it
+        #   is the time.ticks_ms() timestamp of the time that the send completed.
+        #
+        # Note this function only returns an int value one time (the first time it
+        # is called after send completes).
+        #
+        # Part of common low-level modem API, see README.md for usage.
+        if not self._tx:
+            return False
+
+        ticks_ms = self._get_last_irq()
+
+        if not (self._get_irq() & self._IRQ_TX_COMPLETE):
+            # Not done. If the host and modem get out
+            # of sync here, or the caller doesn't follow the sequence of
+            # send operations exactly, then can end up in a situation here
+            # where the modem has stopped sending and has gone to Standby,
+            # so _IRQ_TX_DONE is never set.
+            #
+            # For now, leaving this for the caller to do correctly. But if it becomes an issue then
+            # we can call _get_mode() here as well and check the modem is still in a TX mode.
+            return True
+
+        self._clear_irq()
+
+        self._tx = False
+
+        if self._ant_sw:
+            self._ant_sw.idle()
+
+        # The modem just finished sending, so start receiving again if needed
+        self._check_recv()
+
+        return ticks_ms
+
+
+class RxPacket(bytearray):
+    # A class to hold a packet received from a LoRa modem.
+    #
+    # The base class is bytearray, which represents the packet payload,
+    # allowing RxPacket objects to be passed anywhere that bytearrays are
+    # accepted.
+    #
+    # Some additional properties are set on the object to store metadata about
+    # the received packet.
+    def __init__(self, payload, ticks_ms=None, snr=None, rssi=None, valid_crc=True):
+        super().__init__(payload)
+        self.ticks_ms = ticks_ms
+        self.snr = snr
+        self.rssi = rssi
+        self.valid_crc = valid_crc
+
+    def __repr__(self):
+        return "{}({}, {}, {}, {}, {})".format(
+            "RxPacket",
+            repr(
+                bytes(self)
+            ),  # This is a bit wasteful, but gets us b'XYZ' rather than "bytearray(b'XYZ')"
+            self.ticks_ms,
+            self.snr,
+            self.rssi,
+            self.valid_crc,
+        )
--- a/micropython/lora/README.md
+++ b/micropython/lora/README.md
@@ -1,0 +1,1156 @@
+# LoRa driver
+
+This MicroPython library provides synchronous and asynchronous wireless drivers
+for Semtech's LoRa (Long Range Radio) modem devices.
+
+(LoRa is a registered trademark or service mark of Semtech Corporation or its
+affiliates.)
+
+## Support
+
+Currently these radio modem chipsets are supported:
+
+* SX1261
+* SX1262
+* SX1276
+* SX1277
+* SX1278
+* SX1279
+
+Most radio configuration features are supported, as well as transmitting or
+receiving packets.
+
+This library can be used on any MicroPython port which supports the `machine.SPI`
+interface.
+
+## Installation
+
+First, install at least one of the following "base" LoRa packages:
+
+- `lora-sync` to use the synchronous LoRa modem API.
+- `lora-async` to use the asynchronous LoRa modem API with
+  [asyncio](https://docs.micropython.org/en/latest/library/asyncio.html). Support
+  for `asyncio` must be included in your MicroPython build to use `lora-async`.
+
+Second, install at least one of the following modem chipset drivers for the
+modem model that matches your hardware:
+
+- `lora-sx126x` for SX1261 & SX1262 support.
+- `lora-sx127x` for SX1276-SX1279 support.
+
+It's recommended to install only the packages that you need, to save firmware
+size.
+
+Installing any of these packages will automatically also install a common
+base package, `lora`.
+
+For more information about how to install packages, or "freeze" them into a
+firmware image, consult the [MicroPython documentation on "Package
+management"](https://docs.micropython.org/en/latest/reference/packages.html).
+
+## Initializing Driver
+
+### Creating SX1262 or SX1261
+
+This is the synchronous modem class, and requires `lora-sync` to be installed:
+
+```py
+from machine import SPI, Pin
+import lora import SX1262  # or SX1261, depending on which you have
+
+def get_modem():
+    # The LoRa configuration will depend on your board and location, see
+    # below under "Modem Configuration" for some possible examples.
+    lora_cfg = { 'freq_khz': SEE_BELOW_FOR_CORRECT_VALUE }
+
+    # To instantiate SPI correctly, see
+    # https://docs.micropython.org/en/latest/library/machine.SPI.html
+    spi = SPI(0, baudrate=2000_000)
+    cs = Pin(9)
+
+    # or SX1261(), depending on which you have
+    return SX1262(spi, cs,
+                 busy=Pin(2),  # Required
+                 dio1=Pin(20),   # Optional, recommended
+                 reset=Pin(15),  # Optional, recommended
+                 lora_cfg=lora_cfg)
+
+modem = get_modem()
+```
+
+### Creating SX127x
+
+This is the synchronous modem class, and requires `lora-sync` to be installed:
+
+```py
+from machine import SPI, Pin
+# or SX1277, SX1278, SX1279, depending on which you have
+from lora import SX1276
+
+def get_modem():
+    # The LoRa configuration will depend on your board and location, see
+    # below under "Modem Configuration" for some possible examples.
+    lora_cfg = { 'freq_khz': SEE_BELOW_FOR_CORRECT_VALUE }
+
+    # To instantiate SPI correctly, see
+    # https://docs.micropython.org/en/latest/library/machine.SPI.html
+    spi = SPI(0, baudrate=2000_000)
+    cs = Pin(9)
+
+    # or SX1277, SX1278, SX1279, depending on which you have
+    return SX1276(spi, cs,
+                  dio0=Pin(10),  # Optional, recommended
+                  dio1=Pin(11),  # Optional, recommended
+                  reset=Pin(13),  # Optional, recommended
+                  lora_cfg=lora_cfg)
+
+modem = get_modem()
+```
+
+*Note*: Because SX1276, SX1277, SX1278 and SX1279 are very similar, currently
+the driver uses the same code for any. Dealing with per-part limitations (for
+example: lower max frequency, lower maximum SF value) is responsibility of the
+calling code. When possible please use the correct class anyhow, as per-part
+code may be added in the future.
+
+### Notes about initialisation
+
+* See below for details about the `lora_cfg` structure that configures the modem's
+  LoRa registers.
+* Connecting radio "dio" pins as shown above is optional but recommended so the
+  driver can use pin interrupts for radio events. If not, the driver needs to
+  poll the chip instead. Interrupts allow reduced power consumption and may also
+  improve receive sensitivity (by removing SPI bus noise during receive
+  operations.)
+
+### All constructor parameters
+
+Here is a full list of parameters that can be passed to both constructors:
+
+#### S1261/SX1262
+
+(Note: It's important to instantiate the correct object as these two modems have small differences in their command protocols.)
+
+| Parameter                 | Required               | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |
+|---------------------------|------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `spi`                     | Yes                    | Instance of a `machine.SPI` object or compatible, for the modem's SPI interface (modem MISO, MOSI, SCK pins).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
+| `cs`                      | Yes                    | Instance of a `machine.Pin` input, as connected to the modem's NSS pin.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
+| `busy`                    | Yes                    | Instance of a `machine.Pin` input, as connected to the modem's BUSY pin.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
+| `dio1`                    | No                     | Instance of a `machine.Pin` input, as connected to the modem's DIO1 pin. If not provided then interrupts cannot be used to detect radio events.                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
+| `dio2_rf_sw`              | No, defaults to `True` | By default, configures the modem's DIO2 pin as an RF switch. The modem will drive this pin high when transmitting and low otherwise. Set this parameter to False if DIO2 is connected elsewhere on your LoRa board/module and you don't want it toggling on transmit.                                                                                                                                                                                                                                                                                                                                                 |
+| `dio3_tcxo_millivolts`    | No                     | If set to an integer value, DIO3 will be used as a variable voltage source for the modem's main TCXO clock source. DIO3 will automatically disable the TCXO to save power when the 32MHz clock source is not needed. The value is units of millivolts and should be one of the voltages listed in the SX1261 datasheet section 13.3.6 "SetDIO3AsTCXOCtrl". Any value between `1600` and `3300` can be specified and the driver will round down to a lower supported voltage step if necessary. The manufacturer of the LoRa board or module you are using should be able to tell you what value to pass here, if any. |
+| `dio3_tcxo_start_time_us` | No                     | This value is ignored unless `dio3_tcxo_millivolts` is set, and is the startup delay in microseconds for the TCXO connected to DIO3. Each time the modem needs to enable the TCXO, it will wait this long. The default value is `1000` (1ms). Values can be set in multiples of `15.625`us and range from 0us to 262 seconds (settings this high will make the modem unusable).                                                                                                                                                                                                                                           |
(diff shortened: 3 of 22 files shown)


beyonlo:
Hello @projectgus 

Great. This is a great lib for the MicroPython system! 

1. This lib support all features of the SX1276?
2. This lib is really specific to the SX1276 or support other models?

Thank you very much.

beyonlo:
Sorry, I see above now that is specific for the SX1276.

beyonlo:
I was questioning that because I would like to use in the feature the 2.4Ghz LORA https://www.semtech.com/products/wireless-rf/lora-24ghz for indoor and outdoor location - It has ~3 meters accuracy, like as the [SX1280](https://www.semtech.com/products/wireless-rf/lora-24ghz/sx1280)

projectgus:
@beyonlo The idea is to have a generic LoRA modem API that can be extended to other modem models in the future, such as the SX1280. However at least for now the focus is on messaging and supporting LoRAWAN nodes, rather than ToF range measurement.

beyonlo:
@projectgus No problem. To have a Micropython driver for  messaging on LORA nodes, will be great :)

dpgeorge:
I've had a play with this driver using 2x SX1276 modules connected to a PYBLITEv1.0 and a PYBD-SF2.  Then they run the `rxtx_sync.py` demo code.  I didn't wire up the D0/D1 or RST pins, so it's just using SPI and CS.

It works!  Both devices can send and receive to/from each other.

For the power consumption, it seems quite high.  I measure about 13mA when in RX mode (on top of the pyboard current), and about 30-45mA when transmitting a message (again, on top of pyboard current).  Is that expected?

dpgeorge:
> Decide about support for interleaving rx,tx as managed by the driver - i.e. it would be nice if one task could be continuously (or arbitrarily) calling `receive()` while another task sometimes called `transmit()`, and have the modem driver handle the state changes internally without the caller needing to do anything.

I think this would be very useful, for the driver to automatically manage switching between TX and RX.

For LoRaWAN, or for a custom LoRa protocol, is the idea to have RX constantly on (except when transmitting), or for RX only to be active periodically to reduce power consumption?

micropython/lora/sx1276.py
```diff
@@ -0,0 +1,620 @@
...
+    return min(max(vmin, v), vmax)
+
+
+class SX1276:
+    def __init__(self, spi, nss, dio0=None, dio1=None, radio_reset=None, lora_cfg=None):
+        self._spi = spi
+        self._nss = nss
+
+        self._dio0 = Pin(dio0, mode=Pin.IN) if (dio0 is not None) else None
+        self._dio1 = Pin(dio1, mode=Pin.IN) if (dio1 is not None) else None
```
dpgeorge, line 199: The user should pass in `Pin` objects themselves, rather than pin id's.  That way, this driver doesn't need to access the SPI and Pin classes at all.

(And the user can still pass in `None` here if they don't want to use these pins.)

micropython/lora/sx1276.py
```diff
@@ -0,0 +1,620 @@
...
+                self._dio1.irq(self._dio_isr, trigger=Pin.IRQ_RISING, hard=not __DEBUG__)
+
+        self._bw_hz = 125000  # Reset value
+        self._t_sym_ms = 1  # Reset value, SF=7 BW=125kHz
+
+        # TODO: decide if it's sensible to put pin config here for board-level SX1276 features like TXCO Enable, RF Antenna switch
+
+        if radio_reset:
+            # If the user supplies a reset pin argument, reset the radio
+            radio_reset = Pin(radio_reset, mode=Pin.OPEN_DRAIN, value=1)
```
dpgeorge, line 215: As above, the user should pass in a `Pin` object, not an id.

micropython/lora/sx1276.py
```diff
@@ -0,0 +1,620 @@
...
+
+_REG_VERSION = const(0x42)
+
+
+def _clamp(v, vmin, vmax):
+    return min(max(vmin, v), vmax)
+
+
+class SX1276:
+    def __init__(self, spi, nss, dio0=None, dio1=None, radio_reset=None, lora_cfg=None):
```
dpgeorge, line 194: IMO `nss` should be called `cs` to be consistent with "chip select" used in other drivers.

micropython/lora/sx1276.py
```diff
@@ -0,0 +1,620 @@
...
+        return buf[0]
+
+    def _reg_readinto(self, reg, buf):
+        try:
+            self._nss(0)
+            self._spi.read(1, reg)
+            self._spi.readinto(buf)
+            if __DEBUG__:
+                print("R {:#x} <== {}".format(reg, binascii.hexlify(buf)))
+        finally:
```
dpgeorge, line 445: Does it need a try-finally?  I don't think SPI read functions can raise (??).

micropython/lora/sx1276.py
```diff
@@ -0,0 +1,620 @@
...
+
+    def _reg_read(self, reg):
+        buf = bytearray(1)
+        self._reg_readinto(reg, buf)
+        return buf[0]
+
+    def _reg_readinto(self, reg, buf):
+        try:
+            self._nss(0)
+            self._spi.read(1, reg)
```
dpgeorge, line 441: To eliminate a memory allocation with this read, you can do:
```py
self._spi.readinto(self._buf1, reg)
```
where `self._buf1 = bytearray(1)` is in the constructor.

micropython/lora/sx1276.py
```diff
@@ -0,0 +1,620 @@
...
+        old_value = self._reg_read(reg)
+        if update_mask:
+            value = ((old_value & ~update_mask) & 0xFF) | (new_value & update_mask)
+            self._reg_write(reg, value)
+        return old_value
+
+    def _reg_read(self, reg):
+        buf = bytearray(1)
+        self._reg_readinto(reg, buf)
+        return buf[0]
```
dpgeorge, line 436: This function can be made more efficient, eg:
```py
self._buf2[0] = reg
self._cs(0)
self._spi.write_readinto(self._buf2, self._buf2)
self._cs(1)
return self._buf2[1]
```

micropython/lora/sx1276.py
```diff
@@ -0,0 +1,620 @@
...
+            print("ISR {}".format(pin))
+        if self._irq_callback:
+            self._irq_callback()
+
+    def receive(self, timeout_ms=None, rx_continuous=False):
+        # TODO: check for any pending TX
+        will_irq = self.start_receive(timeout_ms, rx_continuous)
+        while True:
+            if will_irq:
+                machine.idle()
```
dpgeorge, line 487: IMO this `idle()` should always be called, even if `will_irq` is False.  Because otherwise this is a very busy loop that consumes lots of power on the MCU side.

micropython/lora/async_modem.py
```diff
@@ -0,0 +1,47 @@
+# MicroPython LoRa async modem driver
+# MIT license; Copyright (c) 2022 Angus Gratton
+#
+# LoRa is a registered trademark or service mark of Semtech Corporation or its affiliates.
+
+import uasyncio
+
+
+def _clear_irqflag(thread_safe_flag):  # TODO: implement ThreadSafeFlag.clear() in MP
```
dpgeorge, line 9: `ThreadSafeFlag.clear()` is now implemented.

micropython/lora/sx1276.py
projectgus: Good point, will only fail of `buf` or `reg` are invalid and these are supplied by the object itself.

micropython/lora/sx1276.py
projectgus: @dpgeorge I guess this explains the high power consumption. I was worried on some ports this might get stuck in a WFI or similar if no interrupt was set, but I suppose there's always a timer tick or something to wake the CPU up again?

projectgus:
> For the power consumption, it seems quite high. I measure about 13mA when in RX mode (on top of the pyboard current), and about 30-45mA when transmitting a message (again, on top of pyboard current). Is that expected?

That sounded high to me also, but SX1276 datasheet says "Supply current in receive mode" between 10.8 and 11.5mA for AU915 depending on which antenna is used. Transmit current 20-120mA depending on transmit power level.

So probably to be expected.

> >    Decide about support for interleaving rx,tx as managed by the driver - i.e. it would be nice if one task could be continuously (or arbitrarily) calling receive() while another task sometimes called transmit(), and have the modem driver handle the state changes internally without the caller needing to do anything.
>  For LoRaWAN, or for a custom LoRa protocol, is the idea to have RX constantly on (except when transmitting), or for RX only to be active periodically to reduce power consumption?

LoRaWAN specifies particular time slices when the node has to receive. 

Given how much of a power drain constant receiving is, it won't be desirable for battery-powered devices. It's likely most custom LoRa setups will have at least one end on battery power, so it's probably not that useful to have seamless RX/TX switching in the driver. Will pull it out.

micropython/lora/sx1276.py
projectgus: Was reminded the original version of this code called `machine.lightsleep(10)` here. Maybe:

```python
if will_irq:
   machine.idle()
else:
   machine.lightsleep(1)
```

is a reasonable compromise? Max 1ms of lag on receive shouldn't be a problem even for a Class B LoRaWAN "Beacon" devices, I don't think.

dpgeorge: `machine.idle()` is perhaps a bit under-specified in the docs... we don't really have any tickless ports (*) so `idle()` should always wake after some short period.  Thus it would be safe to use it unconditionally here.

If you're concerned, then `time.sleep_ms(1)` would be an alternative.  Don't use `machine.lightsleep()`!  That will put it in a very low power consumption mode that usually disables USB.

(*) zephyr port is perhaps tickless, and `machine.idle()` there calls `k_yield()`.  I'm not sure exactly how that works if there are no interrupts to wake it up.

dpgeorge:
> So probably to be expected.

Yeah, I came to the same conclusion reading the SX datasheet.

> Will pull it out.

OK.  We can always put it back in when we have more experience with this driver.  For now, best to keep it simple/minimal.

projectgus:
@dpgeorge Thanks for the review, I've updated based on your comments and did some additional memory usage optimisations (especially in sx1262.py).

The stm32wl5 support has now been split into an sx1262.py driver, tested with a local module, and I've moved the small amount of additional changes needed for the stm32wl5 into another branch for a separate PR. This means this PR no longer requires any MicroPython changes, and @jimmo your modules should now work properly with it!

dpgeorge:
Looking good!

One thing that's needed is a `manifest.py` for packaging.  Also to consider how to organise the files in this package, and whether they should be split out into separate packages or components within the one package.  For example, as a user I probably only want either the SX1276 or SX1262 driver, but not both.  And I probably only want to use AsyncModem or Modem, but not both.

Probably the directory `micropython/lora` should have subdirs in it for the actual packages, like `micropython/lora/sx1276` and `micropython/lora/lora-modem` etc.   See `aioble` for a good example of how to structure things.  @jimmo will probably have opinions here :smile:

projectgus:
@dpgeorge Can do! Sorry, I did know I needed to do this but I hadn't put it on the TODO list. Will crib from `aioble` and come up with something soon.

`Modem` is actually an abstract base class, so everyone will need that - I will rename it to BaseModem or something and make sure everything is laid out to make this clearer too.

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+        return self._reg_read(_REG_OPMODE) & _OPMODE_MODE_MASK
+
+    def _set_mode(self, mode, lora_en=True):
+        lora_val = _OPMODE_LONGRANGEMODE_LORA if lora_en else 0
+        old_value = self._reg_update(
+            _REG_OPMODE, _OPMODE_LONGRANGEMODE_LORA | _OPMODE_MODE_MASK, mode | lora_val
+        )
+        old_value &= _OPMODE_MODE_MASK
+        if old_value == _OPMODE_MODE_SLEEP:
+            time.sleep_ms(1)  # waking from sleep takes TS_OSC = 250us (typical)
```
tve, line 468: why is this delay needed?

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+            # is probably a driver bug, and once it's fixed then _irq_en can be
+            # fully removed as the modem IRQ mask should determine which IRQs
+            # are driven high
+            if _DEBUG:
+                print("ISR ignored {}".format(pin))
+            return
+        if _DEBUG:
+            print("ISR {}".format(pin))
+        if self._irq_callback:
+            self._irq_callback()
```
tve, line 490: How is the callback supposed to distinguish successful TX, successful RX, RX with CRC error, and RX timeout? Why not pass it that info, which the driver already has?

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+
+    def start_receive(self, timeout_ms=None):
+        mode = self._get_mode()
```
tve, line 493: unused?

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+            # We need to use continuous mode for a no-timeout receive, so stop receiving now
+            self._set_mode(_OPMODE_MODE_STDBY)
+
+        flags = self._reg_update(
+            _REG_IRQ_FLAGS,
+            _IRQ_RX_DONE | _IRQ_RX_TIMEOUT,
+            _IRQ_RX_DONE | _IRQ_RX_TIMEOUT,
+        )  # clear flags
+        if flags & _IRQ_RX_TIMEOUT:  # Modem timed out
+            return None
```
tve, line 546: This code assumes that end_receive is being called because receive-complete got signalled. What if the upper layer just wants to TX a packet and thus calls end_receive to stop the current RX mode? As far as I can tell, it's incorrect to just call start_transmit...

tve (commented):
Nice code overall!

I'm having a hard time picturing how to use this driver, there are a bunch of things missing and few that make usage difficult. If this driver is supposed to be a low level driver that can be used for different high-level LoRa uses then I would expect it to look somewhat different:
- break all sub-parts of `config()` out into separate functions so they can be called when needed, examples are changing BW/SF and TX power in response to ACKs (or lack thereof) or changing channels (freq) (see https://github.com/lemariva/uPyLoRaWAN/blob/master/sx127x.py for example where the init is a collection of calls to these broken out functions)
- ability to have continuous RX as the default mode as opposed to switching back to standby (that's what I tend to use for powered devices)
- provide functionality to check for RX in-progress before TX
- provide a function to calculate packet duration (with LoRa it's not that trivial and often needed since air time can be quite long)
- provide timing information, in particular when a packet was received (I find this essential to sync nodes e.g. to define an ACK window when there is no other shared clock, such as GPS)
- provide functions to enter and leave sleep mode, incl. setting all pins to high impedance

I find the way the IRQ flags are managed rather fragile. All the masking/unmasking only works if start_xx/end_xx are always correctly called in the right sequence. If they're not, one is guaranteed odd difficult to troubleshoot results. I'm more used to clear _all_ flags _before_ switching to TX/RX mode so I'm guaranteed I'm starting from a known state for that phase. Also, I've never used the IRQ masks on the semtech chips and rather do a dispatch in the IRQ handler to figure out what happened. You seem to experience some spurious interrupt according tot he comments and I wouldn't be surprised if it had to do with changing mode, changing IRQ masks, and changing DIO Mapping in back-to-back SPI transactions. There's a wimpy little uC in that sx1276 that has to keep up and do the right thing kepping in mind that the mode change may take way longer than the subsequent SPI transactions (we know that little uC has bugs...)

WRT polling for RX I would not provide any examples or high level code that polls the radio for RX complete. It's a noise disaster in my experience. (Maybe different in stm32wle.) I believe the only appropriate non-interrupt usage pattern for RX is to have an RX window, start the RX at the beginning of the window, wait, and then check whether something arrived _once_ at the end of the window.

I did not see high-level code to operate the driver in always-RX mode and I believe it's rather tricky. Specifically because the IRQ handler gets no info about why it got called and there's an issue with calling RX when no RX-complete indication has been provided (e.g. because one wants to TX).

Documentation???

micropython/lora/lora/async_modem.py
```diff
@@ -0,0 +1,42 @@
...
+        if will_cb:
+            await self._flag.wait()
+            assert self._modem.is_transmit_completed()
+        else:
+            while not self._modem.is_transmit_completed():
+                await uasyncio.sleep_ms(0)
+        self._modem.end_transmit()
+
+    def _callback(self):
+        self._flag.set()
```
tve, line 42: What happens if I first call `receive` and then while it's waiting I call `transmit`?

tve (commented):
Nice code overall!

I'm having a hard time picturing how to use this driver, there are a bunch of things missing and few that make usage difficult. If this driver is supposed to be a low level driver that can be used for different high-level LoRa uses then I would expect it to look somewhat different:
- break all sub-parts of `config()` out into separate functions so they can be called when needed, examples are changing BW/SF and TX power in response to ACKs (or lack thereof) or changing channels (freq) (see https://github.com/lemariva/uPyLoRaWAN/blob/master/sx127x.py for example where the init is a collection of calls to these broken out functions)
- ability to have continuous RX as the default mode as opposed to switching back to standby (that's what I tend to use for powered devices)
- provide functionality to check for RX in-progress before TX
- provide a function to calculate packet duration (with LoRa it's not that trivial and often needed since air time can be quite long)
- provide timing information, in particular when a packet was received (I find this essential to sync nodes e.g. to define an ACK window when there is no other shared clock, such as GPS)
- provide functions to enter and leave sleep mode, incl. setting all pins to high impedance

I find the way the IRQ flags are managed rather fragile. All the masking/unmasking only works if start_xx/end_xx are always correctly called in the right sequence. If they're not, one is guaranteed odd difficult to troubleshoot results. I'm more used to clear _all_ flags _before_ switching to TX/RX mode so I'm guaranteed I'm starting from a known state for that phase. Also, I've never used the IRQ masks on the semtech chips and rather do a dispatch in the IRQ handler to figure out what happened. You seem to experience some spurious interrupt according tot he comments and I wouldn't be surprised if it had to do with changing mode, changing IRQ masks, and changing DIO Mapping in back-to-back SPI transactions. There's a wimpy little uC in that sx1276 that has to keep up and do the right thing kepping in mind that the mode change may take way longer than the subsequent SPI transactions (we know that little uC has bugs...)

WRT polling for RX I would not provide any examples or high level code that polls the radio for RX complete. It's a noise disaster in my experience. (Maybe different in stm32wle.) I believe the only appropriate non-interrupt usage pattern for RX is to have an RX window, start the RX at the beginning of the window, wait, and then check whether something arrived _once_ at the end of the window.

I did not see high-level code to operate the driver in always-RX mode and I believe it's rather tricky. Specifically because the IRQ handler gets no info about why it got called and there's an issue with calling RX when no RX-complete indication has been provided (e.g. because one wants to TX).

Documentation???

projectgus:
Hey @tve,

Thanks for taking a detailed look through this. I appreciate that you seem to be pretty familiar with these radios, it's great to get that perspective as I'm no LoRa expert at this stage.

May defer a few of the in-code replies until I have a chance to check properly on the answers.

>     break all sub-parts of config() out into separate functions so they can be called when needed, examples are changing BW/SF and TX power in response to ACKs (or lack thereof) or changing channels (freq) (see https://github.com/lemariva/uPyLoRaWAN/blob/master/sx127x.py for example where the init is a collection of calls to these broken out functions)

This was modelled after how the micropython-lib bluetooth config() function works, with the idea that any keys which aren't set in the argument dict are kept at their current values. So it's possible for one call to either change one or two parameters, or set all the parameters, depending on the arguments.

I believe this makes for smaller code size, but I could have that wrong. @dpgeorge @jimmo is that right, and is this still a good approach?

>    ability to have continuous RX as the default mode as opposed to switching back to standby (that's what I tend to use for powered devices)
>   provide functionality to check for RX in-progress before TX

The first version of this PR had some of this, but I dropped it as I wasn't sure how useful it was, and it adds a bunch of complexity to manage the concurrency around resuming RX cleanly. Will look at adding it back.

My thought was this can be done in this simplified driver by calling receive() in a loop with a relevant timeout, and if also transmitting then call transmit() in between calls to receive() as needed. I was mostly thinking from the point of view of either fairly simple simplex schemes, or LoRaWAN nodes which have specific short receive and transmit windows but the radio is otherwise idle and window timing is controlled by the host.

If you have a different use case with more complex transmit/receive windows and timing that you'd like this to support the please post more details - I'd like to properly understand it.

>    provide a function to calculate packet duration (with LoRa it's not that trivial and often needed since air time can be quite long)
>    provide functions to enter and leave sleep mode, incl. setting all pins to high impedance
 >   provide a function to read the FEI (frequency error indication), I find it pretty useful to tune the frequency of sensor nodes when using FSK mode; I don't know whether LoRa is more robust WRT freq deviation (hmmm, certainly not in the very narrow band modes) and the FEI is an easy way to ensure the nodes agree with a central gateway (I'm not using LoRaWAN)

All good suggestions, will do!

>  I find the way the IRQ flags are managed rather fragile. All the masking/unmasking only works if start_xx/end_xx are
> always correctly called in the right sequence. If they're not, one is guaranteed odd and difficult to troubleshoot results.
> I'm used to clear all flags before switching to TX/RX mode so I'm guaranteed I'm starting from a known state for that
> phase. Also, I've never used the IRQ masks on the semtech chips and rather do a dispatch in the IRQ handler to figure
> out what happened. You seem to experience some spurious interrupt according tot he comments and I wouldn't be
> surprised if it had to do with changing mode, changing IRQ masks, and changing DIO Mapping in back-to-back SPI
> transactions. There's a wimpy little uC in that sx1276 that has to keep up and do the right thing kepping in mind that the > mode change may take way longer than the subsequent SPI transactions (we know that little uC has bugs...)

Goal here was mostly to keep the code small as possible, but this is a reasonable point - especially if it does turn out that masking is causing some of the spurious interrupt issue.

I really only anticipated modem users calling transmit()/receive() directly (sync or async). The start_xx/end_xx /is_receive_finished/is_transmit_finished are primarily an "internal" API for each modem chip. Of course in that case they should be prefixed as such, will change that.

I know there are use cases where the desire is to call start_xyz and end_xyz manually and interleave other computation in between. My feeling was that switching to async and `await m.transmit()` or `await m.receive(timeout)` instead will make the whole thing simpler.

> WRT polling for RX I would not provide any examples or high level code that polls the radio for RX complete. It's a noise 
> disaster in my experience. (Maybe different in stm32wle.) I believe the only appropriate non-interrupt usage pattern for > RX is to have an RX window, start the RX at the beginning of the window, wait, and then check whether something
> arrived once at the end of the window.

Interesting, I agree supporting non-interrupt completion is definitely a sub-optimal fallback. Can you elaborate on "noise disaster"? Do you mean you get significant RF noise injected from the SPI transactions?

> I did not see high-level code to operate the driver in always-RX mode and I believe it's rather tricky. Specifically because
> the IRQ handler gets no info about why it got called and there's an issue with calling RX when no RX-complete indication
> has been provided (e.g. because one wants to TX).

Agree this is not really possible in this version, see the answer above. If you've got some examples of exactly how you'd like to use the driver, that's probably the best way for me to understand what you'd like it to support internally.

> Documentation???

Yes, definitely. This is a Draft PR and there is a TODO checklist in the first post which includes an item for documentation. I have them half-written now but I wanted to get feedback on the API and approach as early as possible. So thanks for providing good feedback on those things here, it might put the actual docs back a little bit though(!) depending if any changes are made.

micropython/lora/lora/async_modem.py
projectgus: Yes, this is not at all supported right now. Need to either make it clear that the same task should be calling both receive and transmit, or re-add support for cancelling receive.

micropython/lora/lora/sx1276.py
projectgus: The purpose here is to wake up the code blocked in `receive()` or `transmit(), which then reads the interrupt status (and it already knows this information, as you note.)

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+
+_REG_VERSION = const(0x42)
+
+
+def _clamp(v, vmin, vmax):
+    return min(max(vmin, v), vmax)
+
+
+class SX1276(BaseModem):
+    def __init__(self, spi, cs, dio0=None, dio1=None, radio_reset=None, lora_cfg=None):
```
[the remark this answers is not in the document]
projectgus: The fields for lora_cfg definitely need to be fully documented!

There can't really be a default because there isn't a single frequency that works everywhere in the world, although that minimal config `{ 'freq_khz': xyz }` should give a working radio in many cases (but recommend to also set `sf`, `bw`, and probably other settings!).

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+        self._irq_callback = None
+        if self._dio0:
+            self._dio0.irq(self._dio_isr, trigger=machine.Pin.IRQ_RISING, hard=not _DEBUG)
+            if self._dio1:
+                self._dio1.irq(self._dio_isr, trigger=machine.Pin.IRQ_RISING, hard=not _DEBUG)
+
+        self._bw_hz = 125000  # Reset value
+        self._t_sym_ms = 1  # Reset value, SF=7 BW=125kHz
+
+        # TODO: decide if it's sensible to put pin config here for board-level SX1276 features like TXCO Enable, RF Antenna switch
```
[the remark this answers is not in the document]
projectgus: SX1276 actually has this (an optional `ant_sw` parameter which is an object with `tx()` and `rx()` functions. Need to add the same support here, and then can remove this TODO.

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
+# MicroPython LoRa SX1276 driver
+# MIT license; Copyright (c) 2022 Angus Gratton
+#
+# LoRa is a registered trademark or service mark of Semtech Corporation or its affiliates.
+
+from micropython import const
+from .modem import BaseModem, ConfigError
```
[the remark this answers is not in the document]
projectgus: That's right. I'm not very happy with this hierarchy either, but it seemed like the best of a few poor options.

Could make both `SyncModem()` and `AsyncModem()` into wrappers so you always do something like `SyncModem(SX1726(...))` or `AsyncModem(SX1276(...))`. This does save a very small amount of code size for the "async" user, but it also makes the plain modem classes kind of code-smelly (they're no longer useful for anything on their own).

Plus I think sooner or later there will be functionality that needs to go in a common `BaseModem` class under both SX1276 and SX1262 (maybe the packet time calculation function you requested, for example.)

If you can see a possible hierarchy I've missed, I'm up for thinking about it!

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+            pa_config |= (
+                _clamp(lora_cfg["max_power"], 0x0, _PA_CONFIG_MAXPOWER_MASK)
+                << _PA_CONFIG_MAXPOWER_SHIFT
+            )
+            update_mask |= _PA_CONFIG_MAXPOWER_MASK << _PA_CONFIG_MAXPOWER_SHIFT
+        except KeyError:
+            pass
+        try:  # output_power
+            pa_config |= (
+                _clamp(lora_cfg["output_power"], 0x0, _PA_CONFIG_OUTPUTPOWER_MASK)
```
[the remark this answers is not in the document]
projectgus: yeah, this is very reasonable - will put it on the list to implement.

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+        try:
+            self._reg_update(
+                _REG_FSKOOK_IMAGE_CAL,
+                _IMAGE_CAL_AUTO,
+                _IMAGE_CAL_AUTO if lora_cfg["auto_image_cal"] else 0,
+            )
+        except KeyError:
+            pass
+
+        # Update _REG_PA_CONFIG
```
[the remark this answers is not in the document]
projectgus: see general reply about single `config()` with optional keys vs individual functions

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+
+        self._reg_update(_REG_PA_CONFIG, update_mask, pa_config)
+
+        # Not changing PaRAMP value from the default, yet...
+
+        self._reg_write(
+            _REG_IRQ_FLAGS, 0xFF
+        )  # clear all IRQ flags (should be cleared if we have hard reset the radio, but might not be otherwise)
+
+        # Update MODEM_CONFIG1
```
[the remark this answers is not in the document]
projectgus: see general reply about single `config()` with optional keys vs individual functions

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+        self._reg_write(_REG_FIFO_RX_BASE_ADDR, 0)
+        self._set_mode(
+            _OPMODE_MODE_RX_CONTINUOUS if timeout_ms is None else _OPMODE_MODE_RX_SINGLE
+        )
+
+        # will_irq if DIO0 and DIO1 both hooked up, or DIO0 and no timeout
+        return self._dio0 and (self._dio1 or timeout_ms is None)
+
+    def is_receive_completed(self):
+        # assumption here that _REG_IRQ_FLAGS_MASK is set correctly, so we don't have to check exactly which flags are set
```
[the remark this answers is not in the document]
projectgus: yes, this is my mistake I did not notice that the modem will happily deliver you an invalid packet! Thanks

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+        self._reg_write(_REG_FIFO_ADDR_PTR, 0)
+        self._reg_write(_REG_FIFO_RX_BASE_ADDR, 0)
+        self._set_mode(
+            _OPMODE_MODE_RX_CONTINUOUS if timeout_ms is None else _OPMODE_MODE_RX_SINGLE
+        )
+
+        # will_irq if DIO0 and DIO1 both hooked up, or DIO0 and no timeout
+        return self._dio0 and (self._dio1 or timeout_ms is None)
+
+    def is_receive_completed(self):
```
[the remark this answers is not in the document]
projectgus: As noted elsewhere, this function shouldn't have  been indicated as part of a public API. The `receive()` function/coro both call it, but they should know the state when they do that.

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+        self._set_mode(_OPMODE_MODE_TX)
+        return self._dio0 is not None  # will_irq if dio0 is set
+
+    # Use the same function, as setting _REG_IRQ_FLAGS_MASK ensures we only get relevanti nterrupts
+    is_transmit_completed = is_receive_completed
+
+    def end_transmit(self):
+        self._irq_en = ()
+        flags = self._reg_update(_REG_IRQ_FLAGS, _IRQ_TX_DONE, _IRQ_TX_DONE)  # clear flags
+        assert flags & _IRQ_TX_DONE
```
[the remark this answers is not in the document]
projectgus: Consistency check that we don't get here unless the `TX_DONE` interrupt flag was set.

Probably makes more sense in light of the other comments about begin_xxx/end_xxx/etc not being intended as public API functions. Sorry about mis-communicating this in the code!

micropython/lora/lora/async_modem.py
```diff
@@ -0,0 +1,42 @@
...
+    def configure(self, lora_cfg, rx_calibrate=False):
+        return self._modem.configure(lora_cfg, rx_calibrate)
+
+    async def receive(self, timeout_ms=None):
+        # TODO: check for any pending TX
+        self._flag.clear()
+        will_cb = self._modem.start_receive(timeout_ms)
+        if will_cb:
+            await self._flag.wait()
+            assert self._modem.is_receive_completed()
```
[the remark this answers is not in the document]
projectgus: Should always be the case:

1. start_receive() gets called, sets interrupt mask and clears flags.
2. If `will_cb` is set then we know an interrupt is expected.
3. Interrupt occurs. The only interrupt causes that are masked in should be RX_DONE or RX_TIMEOUT.
4. ISR sets the ThreadSafeFlag, so this task wakes up.
5. Calling `is_receive_completed()` here is just "belts and braces" checking "one of the interrupt flags we masked in is set", to catch the possibility the flag has gotten out of sync with the task.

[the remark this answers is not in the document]
projectgus: (As noted above though, I missed that with checking on in the modem you can still receive an invalid packet, need to handle this case.)

micropython/lora/lora/sx1276.py
projectgus: I think this can be sleep_us(250), will change. Somehow I missed that API call was available.

(My understanding of TS_OSC is that the modem won't accept any further SPI commands from the host for minimum of this long after coming out of sleep mode. Is that wrong?)

tve:
Re: not breaking config into several smaller functions: I can work with that. Your code does 5-6 unnecessary SPI read-modify-write cycles if given an empty config map, though. That's not so cool. Also, I wonder how long triggering all these try-except blocks takes, plus getting the chip out of sleep mode again.

> If you have a different use case with more complex transmit/receive windows and timing that you'd like this to support the please post more details - I'd like to properly understand it.

Two use-cases:
- I have lots of sensors around the property and have several gateways. The gateways are powered and relay between LoRa and WiFi, they consist of an esp32 + sx1276. The gateways always have RX enabled. When they receive a packet they briefly turn TX on one second later for an ACK. Also, when they receive an outbound packet over WiFi they briefly switch to TX to forward that packet on LoRa.
- I'm just completing a "buddy tracker" that consists of two nodes, each with GPS, and each displaying an arrow pointing in the direction of the other node. Use-case is paddling in the open ocean where it's tough to keep track of where someone else is when the waves are overhead. These nodes TX every second and self-synchronize so they alternate their TX every half second. I'm hoping to extend this to more than two nodes. RX is always on (except for TX) 'cause managing to get an RX is way more important than battery, which has a hefty load due to the display and GPS anyway. In this use-case I could calculate the RX duration such that it ends right when the node needs to TX, but that complicates the app code.

In general, if the power budget is not an issue then why wouldn't you want to have RX always on?

> > I'm used to clear all flags before switching to TX/RX mode so I'm guaranteed I'm starting from a known state for that
> > phase. 
> Goal here was mostly to keep the code small as possible, but this is a reasonable point - especially if it does turn out that masking is causing some of the spurious interrupt issue.

I believe that just writing `0xff` to the IRQ flags and not dealing with the mask at all actually takes less code.

> I know there are use cases where the desire is to call start_xyz and end_xyz manually and interleave other computation in between. My feeling was that switching to async and `await m.transmit()` or `await m.receive(timeout)` instead will make the whole thing simpler.

Please provide an async code example for running in continuous RX with async events causing a need to TX.

> Do you mean you get significant RF noise injected from the SPI transactions?

Yes! I see a big difference in RX sensitivity once you get close to the noise floor if you're constantly polling the IRQ flags. This was a while ago but I know it was on a PCB, i.e., not breadboard with wires.

> > I did not see high-level code to operate the driver in always-RX mode and I believe it's rather tricky. Specifically because
> > the IRQ handler gets no info about why it got called and there's an issue with calling RX when no RX-complete indication
> > has been provided (e.g. because one wants to TX).
> 
> Agree this is not really possible in this version, see the answer above. If you've got some examples of exactly how you'd like to use the driver, that's probably the best way for me to understand what you'd like it to support internally.

I provided some use-cases. For me, this is a complete deal-breaker, i.e., fork&modify situation. Also, I would caution against assuming too many chip-internal states implicitly, like "if this flag is not set it must be that other one and I don't need to check it". In my experience this comes to bite later when misbehavior is discovered.

As an example, I recently saw https://forum.lora-developers.semtech.com/t/sx1262-reduced-rx-sensitivity-packet-reception-fails/162. The failure hypothesis is that AFC/AGC gets off-track due to a strong nearby interference and then the chip misses weak packets that it would normally RX because it's stuck off-frequency and/or low gain. I don't know whether this is what is happening, but I know that a naive driver for sx1231 (rfm69) or sx1276 in FSK mode looks completely different from a good driver, because a good driver has to "watchdog" the chip and reset it if there's any indication that it's stuck. This entails getting interrupts on detecting a high RSSI level, setting a timer, and restarting RX if the chip is still stuck in trying to decode that non-existent packet at the timeout. It has now been so many years that I (and other people) have fought this issue without any ACK from semtech that I wouldn't be surprised if the LoRa chips had that "signature feature" as well. I had hoped to avoid that in LoRa and my heart sank when I bumped into that thread. I haven't confirmed this behavior myself, so I'm secretly hoping there is some other benign issue...

> > Documentation???
> 
> Yes, definitely. This is a Draft PR and there is a TODO checklist in the first post which includes an item for documentation. I have them half-written now but I wanted to get feedback on the API and approach as early as possible. So thanks for providing good feedback on those things here, it might put the actual docs back a little bit though(!) depending if any changes are made.

Ah, I had missed that checkbox, sorry for that. I actually find that writing the docs first is extremely helpful. I often end up with a paragraph explaining something and then realize that I should really change the API so the explanation turns into a short sentence.

Looking forward to the next iteration!

tve:
Re: [RX getting stuck](https://forum.lora-developers.semtech.com/t/sx1262-reduced-rx-sensitivity-packet-reception-fails/162) . Notice the sentence: "In the GFSK Rx continuous mode on the SX1268 chip and in the LoRa Rx continuous mode on the SX1276 chip, the problem does not appear." in go.valeriy's post.

Some other choice posts in the forum:
- https://forum.lora-developers.semtech.com/t/sx126x-strange-behavior-after-rx-timeout/1080/3
- https://forum.lora-developers.semtech.com/t/new-sx1276-chip-revision-with-reg-version-0x13/588
- https://forum.lora-developers.semtech.com/t/sx1280-settxparams-and-setsleep/535 (that's sx1280, so not _yet_ relevant, but relevant to the way the config() API works)
- https://forum.lora-developers.semtech.com/t/sx1276-misses-received-lora-packets/436 (no second person confirming the issue)

If you want a headache just browse the forum :-O

The upshot of all this, again, just personal experience, is that as these chip features surface the code gets more complicated and nifty code-saving assumptions no longer hold...

micropython/lora/lora/sx1276.py
tve: Where did you see that in the datasheet? I don't think I have used sleep mode on the sx1276, so I don't have any off-datasheet experience.

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
tve: this one caught me too...

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
tve: I do not believe the API in async_modem will be sufficient, i.e., I'm pretty sure I will be calling the various transmit_end/receive_end/etc functions. I provided a good number of examples and maybe you can come up with something that nails it in async_modem...

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
tve: IMHO every assert should at the minimum have a comment that describes what an assertion failure indicates. Specifically here, I don't know whether it would signal invalid use of the end_transmit method by the caller or an unexpected situation WRT the state of the radio.

micropython/lora/lora/async_modem.py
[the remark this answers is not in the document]
tve: Would be nice to add the 5 points under "should always be the case" into the source as comment ;-) (or an updated version as you change things)

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
tve: Where do I find `SyncModem`?
I'm not really understanding the issue with `SyncModem(SX1726(...)) `and `AsyncModem(SX1276(...))`.
I don't have a strong opinion here, nor much experience...

tve:
> I really only anticipated modem users calling transmit()/receive() directly (sync or async). The start_xx/end_xx /is_receive_finished/is_transmit_finished are primarily an "internal" API for each modem chip. Of course in that case they should be prefixed as such, will change that.

> I know there are use cases where the desire is to call start_xyz and end_xyz manually and interleave other computation in between. My feeling was that switching to async and await m.transmit() or await m.receive(timeout) instead will make the whole thing simpler.

As you ponder the API, please consider the following, none of which are likely to be satisfied by AsyncModem I assume, perhaps incorrectly:
- I need to get an as precise as possible timestamp for packet RX so I can synchronize nodes, I'm assuming this means I need to write my own interrupt handler, thus I'll be calling the various transmit and receive functions in SX12xx.py directly and will not use Sync/AsyncModem.
- Without using uasyncio for simple nodes I often overlap TX and RX with other stuff. This is often way simpler than throwing asyncio at the problem. Some examples:
 - wake up, get sensor measurement, start TX, go back to sleep (don't bother waiting for TX-done)
 - wake up, get sensor measurement, start TX, low-power-sleep for the expected TX duration, switch radio to sleep, tri-state everything and go to deep sleep
 - sit in a loop consisting of: get sensor values, start TX, do some other work, wait for TX completion, start RX for ACK, do some more other work, wait for RX done/timeout
 - in my tracker the loop is: TX position and wait for completion, start RX, read GPS, update display, sleep, check RX, based on RX timing sleep a small amt more to hit the correct TX timing
- I will want to collect FEI and perhaps packet CRC error stats as well, unless these are exposed this will require dealing with SX12xx.py directly and writing my own interrupt handler
- I will want to periodically query background "RSSI" to get info on the noise floor, again, mess with SX12xx.py directly and write my own high-level loop
- I may finally get around to querying temperature to gauge the need for recalibration and also to have some warning if the chip is overheating (e.g 20dBm+ operation)

tve:
Yet something else to consider... In the esp32 lightsleep and deepsleep can be interrupted by a pin level but not edge (IIUC). In order to be woken up from sleep when a packet arrives one would use DIO0 and DIO1 as wake-up pins. In order to be able to go back to sleep after handling the interrupt the pin level has to be low again (again, IIUC), this means the interrupt handler needs to clear the IRQ flag.

NB: this is another use-case for a custom interrupt handler: being able to go back to sleep.

micropython/lora/lora/sx1276.py
projectgus: You're right, I don't know why I thought this was necessary. Looks like it's only necessary for FIFO access over SPI, and I don't think this will happen within TS_OSC after coming out of sleep. Will take it out.

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
projectgus: > Where do I find SyncModem?

No, I just meant this was something I considered as an alternative and then decided it made the "simple case" (no async) unnecessarily fiddly to work with. But I don't have a strong preference either way, so maybe I'll try implementing it.

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+        pa_config = 0x0
+        update_mask = 0x0  # only update fields that are in lora_conf
+        try:  # max_power
+            # max_power value is actually ignored if PA_BOOST is set
+            pa_config |= (
+                _clamp(lora_cfg["max_power"], 0x0, _PA_CONFIG_MAXPOWER_MASK)
+                << _PA_CONFIG_MAXPOWER_SHIFT
+            )
+            update_mask |= _PA_CONFIG_MAXPOWER_MASK << _PA_CONFIG_MAXPOWER_SHIFT
+        except KeyError:
```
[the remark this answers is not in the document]
projectgus: A long time ago I learned from some authoritative-sounding source that "Easier to Ask Forgiveness than Permission" (EAFP) was "more Pythonic", and it stuck. This seems to no longer even be a consensus (was it ever?), I see [Guido pushed back against it in 2014](https://mail.python.org/pipermail/python-dev/2014-March/133118.html).

More importantly we can put aside minor style choices to save code size, and you're right we can save code size here:

```
>>> try:
...     x["a"]
... except KeyError:
...     pass
...     
...     
... 
File <stdin>, code block '<module>' (descriptor: 7fba3cddc160, bytecode @7fba3cddc0c0 31 bytes)
Raw bytecode (code_info_size=5, bytecode_size=26):
 2c 06 01 22 53 48 0c 11 03 11 04 10 02 55 34 01
 59 4a 0a 57 11 05 df 44 43 59 4a 01 5d 51 63
arg names:
(N_STATE 6)
(N_EXC_STACK 1)
  bc=0 line=1
  bc=2 line=2
  bc=21 line=4
00 SETUP_EXCEPT 14
02 LOAD_NAME __repl_print__
04 LOAD_NAME x
06 LOAD_CONST_STRING 'a'
08 LOAD_SUBSCR
09 CALL_FUNCTION n=1 nkw=0
11 POP_TOP
12 POP_EXCEPT_JUMP 24
14 DUP_TOP
15 LOAD_NAME KeyError
17 BINARY_OP 8 
18 POP_JUMP_IF_FALSE 23
20 POP_TOP
21 POP_EXCEPT_JUMP 24
23 END_FINALLY
24 LOAD_CONST_NONE
25 RETURN_VALUE
>>> if "a" in x:
...     x["a"]
...     
...     
... 
File <stdin>, code block '<module>' (descriptor: 7fba3cddc3e0, bytecode @7fba3cddc2e0 23 bytes)
Raw bytecode (code_info_size=4, bytecode_size=19):
 10 04 01 27 10 02 11 03 dd 44 4a 11 04 11 03 10
 02 55 34 01 59 51 63
arg names:
(N_STATE 3)
(N_EXC_STACK 0)
  bc=0 line=1
  bc=7 line=2
00 LOAD_CONST_STRING 'a'
02 LOAD_NAME x
04 BINARY_OP 6 
05 POP_JUMP_IF_FALSE 17
07 LOAD_NAME __repl_print__
09 LOAD_NAME x
11 LOAD_CONST_STRING 'a'
13 LOAD_SUBSCR
14 CALL_FUNCTION n=1 nkw=0
16 POP_TOP
17 LOAD_CONST_NONE
18 RETURN_VALUE
```

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
projectgus: Kept the current hierarchy in the end. It's a little asymmetric but it makes the "Simple API" very simple, which I think is worth the trade-off.

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
projectgus: Documented now.

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+            pass
+        try:  # pa_boost
+            pa_boost_bit = (
+                _PA_CONFIG_PASELECT_PA_BOOST_PIN
+                if lora_cfg["tx_ant"].upper() == "PA_BOOST"
+                else _PA_CONFIG_PASELECT_RFO_PIN
+            )
+            pa_config |= pa_boost_bit
+            update_mask |= pa_boost_bit
```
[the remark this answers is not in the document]
projectgus: Yes! Adjusted now.

micropython/lora/lora/sx1276.py
projectgus: The IRQ callback still doesn't get any modem state, but I still think this is OK. What the implementer is supposed to do is now different and I hope clearer, and also documented.

micropython/lora/lora/sx1276.py
projectgus: There is now a "low level API" with documentation and (I hope) less arbitrary behaviour.

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
projectgus: Implemented but not yet documented, there's a TODO for that in the first post.

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+        try:
+            # Assuming F(XOSC)=32MHz (datasheet both implies this value can be different, and specifies it shouldn't be different!)
+            fr_val = int(lora_cfg["freq_khz"] * 16384 // 1000)
+            buf = bytes([fr_val >> 16, (fr_val >> 8) & 0xFF, fr_val & 0xFF])
+            self._reg_write(_REG_FR_MSB, buf)
+        except KeyError:
+            pass
+
+        # RX chain calibration at the specific frequency
+        if rx_calibrate:
```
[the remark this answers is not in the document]
projectgus: Added explicit `calibrate()` and `calibrate_image()` functions to the API. Haven't documented yet, have TODO to document.

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+            raise RuntimeError("Unexpected silicon version {}".format(version))
+
+        if lora_cfg:
+            self.configure(lora_cfg, True)
+
+    def set_irq_callback(self, callback):
+        self._irq_callback = callback
+
+    def configure(self, lora_cfg, rx_calibrate=False):
+        self._set_mode(_OPMODE_MODE_SLEEP)
```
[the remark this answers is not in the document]
projectgus: Good point, it was not (except for switching in and out of LoRa mode).

micropython/lora/lora/async_modem.py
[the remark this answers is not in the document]
projectgus: This particular code has changed a lot now, but there are also more comments now.

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
projectgus: Now there is a new low-level API, and a more complex example (although not as complex as your applications).

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
projectgus: Driver now drops invalid packets by default, can optionally set a field `rx_crc_error`  to have it return invalid ones (with a flag set on them to indicate invalid CRC).

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
projectgus: `output_power` is now set in dBm

micropython/lora/lora/sx1276.py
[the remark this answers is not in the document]
projectgus: Switched!

micropython/lora/lora/sx1276.py
```diff
@@ -0,0 +1,592 @@
...
+            # This field has a reserved non-zero field, so do a read-modify-write update
+            self._reg_update(
+                _REG_DETECT_OPTIMIZE,
+                _DETECT_OPTIMIZE_AUTOMATIC_IF_ON | _DETECT_OPTIMIZE_MASK,
+                _DETECT_OPTIMIZE_SF6 if sf == 6 else _DETECT_OPTIMIZE_OTHER,
+            )
+
+            modem_config2 |= sf << _MODEM_CONFIG2_SF_SHIFT
+            update_mask |= _MODEM_CONFIG2_SF_MAX << _MODEM_CONFIG2_SF_SHIFT
+            self._t_sym_ms = 1000 * (1 << sf) // self._bw_hz
```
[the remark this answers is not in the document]
projectgus: New function added: `modem.get_time_on_air_us(plen)`, with some tests to verify the times match the Semtech-provided calculator.

micropython/lora/lora/async_modem.py
projectgus: Now supported!

projectgus:
Big update, lots of additional functionality! @tve @dpgeorge @jimmo Please take a look and let me know what you think.

Still a few small outstanding jobs, but most I've pushed to new PRs as this one is too big already. See first post for details.

The only real downside of adding a lot of functionality is adding code size:

| file | November version (95204e4) | this version (9c90a5995) |
| --- | --- | --- |
| modem.mpy | 539 | 2542 |
| sx126x.mpy | 3599 | 5603 |
| sx127x.mpy | 3320 | 4641 |
| async_modem.mpy | 634 | 1397 |
| Total | 8092 | 14183 |

I feel most of this functionality are things that anyone doing non-trivial LoRa applications will find useful though, and won't be that simple to implement on top of a simpler driver. So I'm not sure I have a better suggestion to make on reducing size here.

micropython/lora/lora/modem.py
```diff
@@ -0,0 +1,510 @@
...
+        # Part of common low-level modem API, see README.md for usage.
+        #
+        # This function may alter the state of the modem - it will clear
+        # RX interrupts, it may read out a packet from the FIFO, and it
+        # may resume receiving if the modem has gone to standby but receive
+        # should resume.
+
+        if not self._rx:
+            # Not actually receiving...
+            return (False, None)
```
projectgus, line 287: @dpgeorge @jimmo In the receive and transmit control flows I now have this pattern of returning 2-tuples and 3-tuples in a few places. In the previous version of this PR these types of results were constrained to "small" types but it got fiddly, not to mention quite confusing to read.

What's your feeling on these additional small heap allocations? I'm thinking I could rework it to allocate a set of list objects and keep them around for reuse, like what's done for the modem command buffers already, but I'm not certain it's worth the complexity vs RAM use tradeoff.

FWIW I think these are the only places that heap allocations can't be avoided during the normal transmit or receive control flow.

projectgus:
@tve regarding a couple of these points:

> I will want to collect FEI and perhaps packet CRC error stats as well, unless these are exposed this will require dealing with SX12xx.py directly and writing my own interrupt handler
> I will want to periodically query background "RSSI" to get info on the noise floor, again, mess with SX12xx.py directly and write my own high-level loop
> I may finally get around to querying temperature to gauge the need for recalibration and also to have some warning if the chip is overheating (e.g 20dBm+ operation)

There is some CRC awareness now, the driver by default drops invalid CRC packets and increments a counter `crc_errors` on the modem object that can be read or cleared by the programmer as needed (this needs documenting still). There is also a field option `rx_crc_error` that can be set to True in order for the modem to return messages with invalid CRCs rather than drop them (this part is documented already.)

I want to add an instantaneous RSSI function as well, but this isn't done yet (I'll put it in a future PR, or contribution welcome!).

FEI seems to only be supported on SX1276 for LoRa, which is a bit surprising. On the topic of frequency accuracy, all I see in the SX126x documentation is a guide to keep the bandwidth within the error limit of the clock source and an equation for that. So probably won't make this part of the common modem API, if it's not on any of the newer modems. However, for cases like this then it should always be possible to call `modem._read_reg()` to read out particular register fields and do "advanced" modem-specific stuff. In particular, the packet statistic registers seem to stay valid until the next packet is received so can easily call them after a receive ends and before going back to sleep. Ditto for manually checking the temperature registers and determining if re-calibraiton is needed.

projectgus:
>  If you want a headache just browse the forum :-O

Regarding silicon bugs, this is indeed sobering reading.  From what I can see, the only fully confirmed bug in that list is SX126x AGC getting stuck on "min gain" when the receiver gets swamped temporarily. I think probably the best thing to do with these is to document "known" (if not confirmed) silicon bugs in the README and give some workaround tips (like stopping a continuous receive periodically if nothing is received).

I did find what seems like it might be a very nasty silicon bug though, workaround applied in b6eaf74. I still feel like it's too nasty to be that simple, but logic analyzer traces confirm the commands on the wire look correct and the modem otherwise works fine... Might report this one to Semtech as well.

projectgus:
Added some more updates and cleanups.

Someone also emailed me to report this driver works well on SX1278 as well (yay!) Looking at the datasheet I don't see any reason not to support all four modems, so it's been updated to do this.

micropython/lora/lora/modem.py
projectgus: Following some discussion offline, I've removed these tuples and the normal send/receive code paths are heap allocation free (provided rx_packet instance is passed into receive() and the new packet has the same length.)

micropython/lora/lora/lora/modem.py
```diff
@@ -0,0 +1,537 @@
...
+
+        return ticks_ms
+
+    # Simple synchronous modem API functions
+    #
+    # These are intended for simple applications. They block the caller until
+    # the modem operation is complete, and don't support interleaving transmit
+    # and receive.
+
+    def transmit(self, packet, tx_at_ms=None):
```
dpgeorge, line 447: IMO (and @jimmo agrees), this driver (and all its components) should use `send` and `recv` naming, like a `socket.socket` does.  (It's not a stream so not read/write.)

That would mean this method is called `send`, and the method below `recv`.

Then it would also have `prepare_send`, `start_send`, `poll_send`, `start_recv`, `poll_recv`, `recv_continuous` etc.

dpgeorge: Some of the reasons for this are:
1. more Pythonic, matches naming elsewhere
2. reduces code size because the names already exist as qstrs

micropython/lora/lora-async/lora/async_modem.py
```diff
@@ -0,0 +1,149 @@
...
+        # to wake both for the case of a "soft" interrupt triggered by sleep() or standby(), where
+        # both tasks need to unblock and return.
+        #
+        # The poll_receive() method is a no-op if _tx is set, so waking up both
+        # blocked tasks doesn't have much overhead (no second polling of modem
+        # IRQ flags, for example).
+        for f in self._flags:
+            f.set()
+
+    # Wrappers for common API calls on the inner modem object
```
dpgeorge, line 110: Unless the below methods are overridden by the SX12xx class, I think it's possible (and it's simpler, less code) to do:
```py
configure = BaseModem.configure
```

dpgeorge:
One thing @jimmo and I just discussed was splitting out the sync-modem methods (namely `BaseModem.transmit`, `BaseModem.receive` and `BaseModem._sync_wait`) into a separate class called `SyncModem`.  That would take a particular SX12xx instance in its constructor, essentially mirroring `AsyncModem`.  It would live in `lora-sync/lora/sync_modem.py`.

Benefits:
- a clear separation of layers (BaseModem -> SX12xx -> SyncModem/AsyncModem)
- no need to have any of the sync modem code if you only use AsyncModem

Drawbacks:
- users must now always install (at least) two packages: one for their hardware, eg `sx126x`, and one for the top-level driver (sync or async)

@projectgus what are your thoughts on that?

micropython/lora/lora-async/lora/async_modem.py
projectgus: Unfortunately these all call through to `self.modem.configure()` not `self.configure`, so I think no easier way to bind it.

(Unless this class was implemented with multiple inheritance instead, see comment below.)

projectgus:
@dpgeorge 
> One thing @jimmo and I just discussed was splitting out the sync-modem methods (namely `BaseModem.transmit`, `BaseModem.receive` and `BaseModem._sync_wait`) into a separate class called `SyncModem`. That would take a particular SX12xx instance in its constructor, essentially mirroring `AsyncModem`. It would live in `lora-sync/lora/sync_modem.py`.

@tve also asked about the possibility of a change like this, so it's clearly a logical layout for folks.

I agree about the benefits, it's a more symmetrical set of relationships. The reasons I avoided it until now was a general idea of "keep the simple case - Sync - as simple as possible", but also that I don't know what to logically describe these classes if we split them into two types of "thing":

* SyncModem  & AsyncModem are one type of thing (a "modem interface"?)
* SX1276, SX1262, BaseModem, etc are a different type of a thing (a "modem implementation"?)

... not knowing what names to give them makes me think it's a code smell.

(I guess the current code has this problem as well regarding AsyncModem, and maybe if keeping the current structure then AsyncModem should be called AsyncAdapter or something to disambiguate... this is all starting to feel very Java-y!)

The other way around this would be to use multiple inheritance, so each "public" class in the package inherits from either SyncModem or AsyncModem and also from a chip-specific modem base (SX1276, SX1262, etc.), meaning you can do something like `from lora.sync import SX1276`. I generally avoid multiple inheritance (and I'm not even sure how well MP supports it) but actually this might be the most intuitive class structure to go with?

projectgus:
Method renames have been implemented!

> One thing @jimmo and I just discussed was splitting out the sync-modem methods (namely BaseModem.transmit, BaseModem.receive and BaseModem._sync_wait) into a separate class called SyncModem. That would take a particular SX12xx instance in its constructor, essentially mirroring AsyncModem. It would live in lora-sync/lora/sync_modem.py.

@dpgeorge @jimmo Please take a look at the head commit in this branch. I'm not 100% sure this is a good idea, but I think it's still preferable to having SyncModem/AsyncModem wrappers that take chip-specific objects as arguments.

I put some reasons in the previous comment on this PR, but the final challenge was telling the user which object to call which method on. For example, do you call `modem.configure(...)` or `modem.modem.configure(...)`? The old AsyncModem wrapper had some hacky workarounds for this like `def configure(self, ...): self.modem.configure(...)`, but this is not a great solution either.

If you think the "mixin-like class" approach in that commit is too fiddly then I fully understand though, I'm not 100% convinced on it myself - the best I could say for it is that it might be the "least bad" way. Am happy to drop thjs commit and try something else instead if you're not convinced.

projectgus:
(Rebased, tweaked some wording in the README a tiny bit.)

projectgus:
Rebased, pushed a bugfix with CRC as reported by Barry on Discord and another bug fix with async wait timeout on send (only relevant if IRQ doesn't trigger as expected).

dpgeorge:
Please make it configure the pins in the constructor (using `hasattr(pin, init)`).  See `sdcard.py` for example.

Also rename u-module to module.

projectgus:
@dpgeorge @jimmo Thanks for the review of a large and complex PR! Have applied the requested changes (see [30214ee](https://github.com/micropython/micropython-lib/pull/501/commits/30214eeb78711555e7b8b135a753c42f649f1fd0), [247597b](https://github.com/micropython/micropython-lib/pull/501/commits/247597b95499b93b4c29d26edb49d744f01e348d)), re-tested the two major modem models, squashed commits and rebased on master.

dpgeorge:
I tested the `simple_rxtx` code, both sync and async versions, on a pair of SX1276's.  It works!

(I had to fix the `simple_rxtx_async.py` to rename `uasyncio` to `asyncio`, and I pushed that here.)

dpgeorge:
Thanks for the huge effort on this @projectgus !

mattytrentini:
Congrats and thanks to everyone involved!

PR #502, opened 2022-07-04
micropython/monotonic: Add a wrapper replacement for time.monotonic().

andrewleech:
MicroPython replacement for time.monotonic suitable for timeouts / comparisons

CPython time.monotonic() → float returns the value (in fractional seconds) of a monotonic clock,
i.e. a clock that cannot go backwards. The clock is not affected by system clock updates.
The reference point of the returned value is undefined, so that only the difference between the
results of two calls is valid.

Most micropython ports have single-precision float for size / efficiency reasons, and some do not have
float support at all in hardware (so are very slow).
To support measurements of difference between two time points, time.ticks_ms() and time.ticks.diff()
are generally recommended, however this can complicate efforts to port existing libraries using
time.monotonic.

This library is intended to support being used as a drop-in replacement for many/most use cases of
time.monotonic. It will wrap the ticks functions and handle/hide the 32-bit rollover handling.

Note however if you convert the output of monotonic to int or float, eg `float(monotonic())` then
comparisions between these value are not always valid becasuse they will wrap around back to zero
after a certain length of time. In other words, always do comparisons against the object returned
by monotonic() without type conversion.

See the test_monotonic.py unit test for examples of usage.


micropython/monotonic/monotonic.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __gt__(self, other):
+        return int(self.__sub__(other)) > 0
+
+    def __lt__(self, other):
+        return int(self.__sub__(other)) < 0
+
+
+def monotonic(impl=int):
+    return TimeMS(ticks_ms(), impl=int)
```
thyrrestrup, line 76: should this be `TimeMS(ticks_ms(), impl)` with `=int` removed?

micropython/monotonic/monotonic.py
```diff
@@ -0,0 +1,76 @@
...
+from time import ticks_ms, ticks_diff
+from time import *
+
+
+class TimeMS:
+    def __init__(self, val_ms, impl=int) -> None:
+        self.val_ms = val_ms
+        self.impl = impl
+
+    def __int__(self):
```
thyrrestrup, line 41: Is the `__int__` function ever used?

micropython/monotonic/monotonic.py
andrewleech: Ah yes it should thanks!

PR #503, opened 2022-07-06
python-stdlib/threading: Add Lock, Condition and Event to threading.

andrewleech:
Requires https://github.com/micropython/micropython-lib/pull/440


dpgeorge:
PR #440 was closed without merging.  The code is now in upstream, but the `clear()` method is not implemented.

Can you change this code here so it doesn't need `deque.clear()`?

Also, what's the licensing of the code here, is it copied from CPython?

andrewleech:
Thanks yes I'll update that. 

These classes were copied from cpython yes, with very minimal updates for compatibility if I recall correctly. 

> The licencing declaration is a good point, thanks for bringing it up. I'm not sure if it's possible/appropriate to keep these in the same file as the existing (MIT presumably) Thread class? I could comment each of these classes with licence / link to original cpython implementation, or perhaps they need to move to a separate file with suitable comment at the top?

PR #505, opened 2022-07-14
keyword: Add pristine from Python-3.3.3 tarball.

hejaali84:
h


andrewleech:
I've just taken a look at the [python 3.5 version of this](https://github.com/python/cpython/blob/3.5/Lib/keyword.py) and the [3.10 version](https://github.com/python/cpython/blob/3.10/Lib/keyword.py) and it's clear that everything in the `main()` function is simply there to auto-generate the contents of the lists from CPython source - as such it's not going to be useful here on micropython.

If anything, the 3.10 source above would be a better starting point as all the generation code was refactored out into a separate module, leaving just the useful lists / attributes here.

Could you explain a little of your use case for this module? According to [the docs](https://docs.python.org/3/library/keyword.html) for it, it's supposed to be a list of keywords that match the interpreter. The list here currently matches the CPython 3.3 interpreter which will certainly be slightly different to the current micropython vm.

PR #508, opened 2022-07-16
time: Add time module to provide strftime.

iabdalkader:


python-stdlib/time/time.py
```diff
@@ -0,0 +1,37 @@
...
+_TS_WDAY=6
+_TS_YDAY=7
+_TS_ISDST=8
+
+_WDAY = ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday",
+     "Saturday", "Sunday"]
+_MDAY = ["January", "February", "March", "April", "May", "June",
+     "July", "August", "September", "October", "November", "December"]
+
+_time_regex = re.compile(r"%(-?[a-zA-Z])")
```
andrewleech, line 18: If be inclined to have this defined as None here, then calculated / set on first use in the function below to avoid the time / memory use during import.

python-stdlib/time/time.py
```diff
@@ -0,0 +1,37 @@
...
+_TS_YEAR=0
+_TS_MON=1
+_TS_MDAY=2
+_TS_HOUR=3
+_TS_MIN=4
+_TS_SEC=5
+_TS_WDAY=6
+_TS_YDAY=7
+_TS_ISDST=8
+
```
andrewleech, line 12: These probably should all be `const()`

In the latest mpy the lists below could be too, though I'm not sure how important back compatibility is here.

andrewleech:
FYI I've also got a recent open PR to add a `time` module with `monotonic` in https://github.com/micropython/micropython-lib/pull/502

It'd certainly be trivial to merge them eventually.

It's probably worth you adding `from time import *` in `__init__.py` so this new module transparently includes everything from there too?

iabdalkader:
> It's probably worth you adding `from time import *` in `__init__.py` so this new module transparently includes everything from there too?

I don't think it should import everything in there.

andrewleech:
In micropython, the C implementation of builtins are in the `u` modules, eg `uos`, `ucollections`. These are automatically shadowed to the non-u names if there's no other non-u implementation on path, eg `os`, `collections` for compatibility. 

If there's a python / frozen version of the non-u version provided however that overrides the shadow; the intention is to allow python modules to extend the built-ins exactly like this PR is doing.

As-is, once this time.py is added to path, `import time` should no longer include the rest of the built in functionality? 
In my understanding the python built-in packages should wrap / override the built-in C `u` version, see https://github.com/micropython/micropython-lib/blob/70e422dc2e885bbaafe6eb7e3d81118e17d4b555/python-stdlib/collections/collections/__init__.py#L5

That way`import time` will give access to everything in C `utime` as well as the new python functionality.

iabdalkader:
Ah okay, but you said :

> It's probably worth you adding from time import * in __init__.py so this new module transparently includes everything from there too?

I think you meant from `utime import *` now that makes sense yes.

andrewleech:
> I think you meant from `utime import *` now that makes sense yes.

Ah geez sorry for the confusion, I missed the imperfect autocorrect!

python-stdlib/time/setup.py
```diff
@@ -0,0 +1,24 @@
+import sys
```
dpgeorge, line 1: This file (and metadata.txt) are no longer needed, only `manifest.py` is.

python-stdlib/time/time.py
```diff
@@ -0,0 +1,54 @@
...
+_time_regex = None
+
+
+def strftime(datefmt, ts):
+    global _time_regex
+    if _time_regex is None:
+        _time_regex = re.compile(r"%(-?[a-zA-Z])")
+    return _time_regex.sub(r"%(\1)s", datefmt) % {
+        "a": _WDAY[ts[_TS_WDAY] - 1][0:3],
```
dpgeorge, line 39: I think wday is 0-6 for Mon-Sun, so there's no need to do the `-1` here and below?

python-stdlib/time/time.py
```diff
@@ -0,0 +1,54 @@
...
+    )
+)
+
+_time_regex = None
+
+
+def strftime(datefmt, ts):
+    global _time_regex
+    if _time_regex is None:
+        _time_regex = re.compile(r"%(-?[a-zA-Z])")
```
dpgeorge, line 37: What is the `-?` used for?

python-stdlib/time/time.py
```diff
@@ -0,0 +1,54 @@
...
+        "I": f"{ts[_TS_HOUR]%12:02d}",
+        "j": f"{ts[_TS_YDAY]:03d}",
+        "m": f"{ts[_TS_MON]:02d}",
+        "M": f"{ts[_TS_MIN]:02d}",
+        "P": "AM" if ts[_TS_HOUR] < 12 else "PM",
+        "S": f"{ts[_TS_SEC]:02d}",
+        "w": f"{ts[_TS_WDAY]}",
+        "y": f"{ts[_TS_YEAR]%100:02d}",
+        "Y": f"{ts[_TS_YEAR]}",
+    }
```
dpgeorge, line 54: This is a neat way to implement strftime (using the dict and `%` formatting), but it will contribute to a lot of heap allocations, and most of the strings created here will never be used by the caller of `strftime`.

For example, if logging happens frequently (eg a few times per second) then this is going to put a lot of pressure on the GC heap.  Each call to `strftime` creates about 12 strings, only one of which is the actual formatted string.

I don't really have a good alternative, except to implement this as a string parser that looks for the `%` patterns and substitutes them on demand.  Ie `for c in datefmt: ...`.

python-stdlib/time/time.py
```diff
@@ -0,0 +1,54 @@
...
+def strftime(datefmt, ts):
+    global _time_regex
+    if _time_regex is None:
+        _time_regex = re.compile(r"%(-?[a-zA-Z])")
+    return _time_regex.sub(r"%(\1)s", datefmt) % {
+        "a": _WDAY[ts[_TS_WDAY] - 1][0:3],
+        "A": _WDAY[ts[_TS_WDAY] - 1],
+        "b": _MDAY[ts[_TS_MON] - 1][0:3],
+        "B": _MDAY[ts[_TS_MON] - 1],
+        "d": f"{ts[_TS_MDAY]:02d}",
```
dpgeorge, line 43: I think it's better to stick with `%` formatting here, ie `"%02d" % ts[_TS_MDAY]`.  Because this function already uses `%` formatting to do the final formatting step, and it'll mean it can work on ports that don't have f-strings enabled.

python-stdlib/time/time.py
iabdalkader: A `-` in the format specifier should remove leading zeros, but since it's not really supported, I've removed it.

python-stdlib/time/time.py
dpgeorge: I do prefer the new `strftime2` version, simply because it will perform less allocations (and may be faster overall due to less heap allocations).

But I suggest using `io.String()` to build the output string, so it doesn't need to create a new string for each `+=`.  Also please do `import io` within the function so that ports that don't have it can still load this code (and just not use `strftime`).

Furthermore, I think the code can be a little simpler with this logic:
```py
for k in datefmt:
    if fmtsp:
        if k == "a":
            ..
        else:
           ftime += k
        fmtsp = False
    elif k == "%":
        fmtsp = True
    else:
        ftime += k
```

iabdalkader: > within the function so that ports that don't have it can still load this code (and just not use strftime).

@dpgeorge Can we use a bytearray maybe so it's available on most (all?) ports ?

iabdalkader: Or a list and then `"".join(ftime))`

dpgeorge: A bytearray might have problems with unicode characters (or it may be hard to handle them properly).  Using list and join will use a bit of extra memory for the list itself.

I think `io.StringIO()` is the best way to do it.  And from `py/mpconfig.h`:
```c
#define MICROPY_PY_BUILTINS_STR_OP_MODULO (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_CORE_FEATURES)
#define MICROPY_PY_BUILTINS_BYTEARRAY (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_CORE_FEATURES)
#define MICROPY_PY_IO (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_CORE_FEATURES)
```
That means that, %-formatting, bytearray and the `io` module are all available at the same feature level.  So `io.StringIO()` will most likely be available on ports that use this function (as long as they use the feature level config, which we are trying to make sure all ports do).

iabdalkader: > Using list and join will use a bit of extra memory for the list itself.

Changed to StringIO.

dpgeorge:
Thanks for updating.  Merged, and I removed the `import re` and associated global variable because it's no longer used.

I also added a unit test (taking out the test from `time.py`).

Issue #369, opened 2020-02-03
hmac code not compatible with hashlib/uhashlib implementations


tmmorin:
I noticed that micropython's `hmac` implementation [relies on the existence of `.copy()` on instances of shaXX](https://github.com/micropython/micropython-lib/blob/master/hmac/hmac.py#L118), which works fine with the implementation provided in `hashlib._sha224`/`256`/`384`/`512`, but does *not* work with the implementations that `hashlib` auto-loads from `uhashlib` when `uhashlib` is present (since [this change](https://github.com/micropython/micropython-lib/issues/235)).

I tested recent versions micropython-lib master and micropython 1.12).

```
$ cat test2.py 
import hmac
from hashlib import sha256
print(hmac.new(b'test', msg=b'test', digestmod=sha256).hexdigest())
```
```
$ micropython test2.py 
Warning: No block_size attribute on given digest object; Assuming 64.
Traceback (most recent call last):
  File "test2.py", line 4, in <module>
  File "/home/treizh/.micropython/lib/hmac.py", line 137, in hexdigest
  File "/home/treizh/.micropython/lib/hmac.py", line 120, in _current
AttributeError: 'sha256' object has no attribute 'copy'
```

Forcing the use of the implementation under `hashlib._sha256` works though:

```
$ cat test.py 
import hmac
from hashlib._sha256 import sha256
```
```
print(hmac.new(b'test', msg=b'test', digestmod=sha256).hexdigest())
$ micropython test.py 
88cd2108b5347d973cf39cdf9053d7dd42704876d8c9a9bd8e2d168259d3ddf7
```

(this workaround unfortunately does not work for SHA1 since SHA1 has no implementation in `hashlib`)

I presume that whether `hmac` should be implemented without using relying on `copy()`, or at least should not autoload `hashlib` (this last solution would leave HMAC SHA1 unimplemented).

tmmorin:
@peter-conalgo: perhaps are you interested in this issue ?

peter-conalgo:
Somewhat. I used this code as an example, and made my own when I needed HMAC recently. Pared down features I didn't need so that in the end, I didn't need to copy state.

My version:

<https://github.com/Coldcard/firmware/blob/119d7f79d82d98d35e53a941bbc065fa81182e50/shared/hmac.py>

tmmorin:
Thanks @peter-conalgo. I'll try your code !

cositotito:
hello the code error digi

[...]

PR #511, opened 2022-07-26
python-ecosys/ujwt: Add ujwt module.

### LINKED ISSUE #510: Interest in a ujwt library?
I have written a tiny version of pyjwt, which I plan to use to handle web sessions in my Microdot library. Would there be any interest in adding it to python-ecosys here? Code is [here](https://gist.github.com/miguelgrinberg/400955ef03169d0a523fb1a2bd133a77) for those interested.

miguelgrinberg:
Fixes #510 

Here is my ujwt module. Note that I had to fix a small issue in the hashlib library, which prefers to import the built-in `sha256` class even when micropython-hashlib is installed. Unfortunately the hmac library does not work with the built-in hashlib.

Let me know if you need anything else. I have added the packaging files, but I have no idea if there is anything I need to include to push the package to pypi.

### DIFF
--- a/python-ecosys/ujwt/metadata.txt
+++ b/python-ecosys/ujwt/metadata.txt
@@ -1,0 +1,3 @@
+srctype = micropython-lib
+type = module
+version = 0.1.0
--- a/python-ecosys/ujwt/setup.py
+++ b/python-ecosys/ujwt/setup.py
@@ -1,0 +1,25 @@
+import sys
+
+# Remove current dir from sys.path, otherwise setuptools will peek up our
+# module instead of system's.
+sys.path.pop(0)
+from setuptools import setup
+
+sys.path.append("..")
+import sdist_upip
+
+setup(
+    name="micropython-ujwt",
+    version="0.1",
+    description="ujwt module for MicroPython",
+    long_description="This is a module reimplemented specifically for MicroPython standard library,\nwith efficient and lean design in mind. Note that this module is likely work\nin progress and likely supports just a subset of CPython's corresponding\nmodule. Please help with the development if you are interested in this\nmodule.",
+    url="https://github.com/micropython/micropython-lib",
+    author="micropython-lib Developers",
+    author_email="micro-python@googlegroups.com",
+    maintainer="micropython-lib Developers",
+    maintainer_email="micro-python@googlegroups.com",
+    license="MIT",
+    cmdclass={"sdist": sdist_upip.sdist},
+    py_modules=["ujwt"],
+    install_requires=["micropython-hmac"],
+)
--- a/python-ecosys/ujwt/test_ujwt.py
+++ b/python-ecosys/ujwt/test_ujwt.py
@@ -1,0 +1,26 @@
+import ujwt
+from time import time
+
+secret_key = "top-secret!"
+
+jwt = ujwt.encode({"user": "joe"}, secret_key)
+decoded = ujwt.decode(jwt, secret_key)
+if decoded != {"user": "joe"}:
+    raise Exception("Invalid decoded JWT")
+else:
+    print("Encode/decode test: OK")
+
+try:
+    decoded = ujwt.decode(jwt, "wrong-secret")
+except ujwt.exceptions.InvalidSignatureError:
+    print("Invalid signature test: OK")
+else:
+    raise Exception("Invalid JWT should have failed decoding")
+
+jwt = ujwt.encode({"user": "joe", "exp": time() - 1}, secret_key)
+try:
+    decoded = ujwt.decode(jwt, secret_key)
+except ujwt.exceptions.ExpiredTokenError:
+    print("Expired token test: OK")
+else:
+    raise Exception("Expired JWT should have failed decoding")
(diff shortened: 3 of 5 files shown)


jimmo:
Thanks @miguelgrinberg  Code looks good and keen to merge it, I just wanted to bring up two broader points.

Just quickly first on the hashlib change though (see also #369 & #435) I think this makes sense for now, the current behaviour is clearly not right. Longer term, I would prefer to see this fixed by adding the missing methods to the built-in implementations.

1. Should it be called "ujwt"?
a) It matches the implementation of PyJWT, so should it be "upyjwt" ? I know that when you install PyJWT, you get the "jwt" module though. Frustratingly the "jwt" library on PyPi is a different thing.  (It's pyserial all over again...)
b) Should we be using the "u" prefix? I know "urequests" is that way, but that's kind of a historical accident. The way I want this to work is MicroPython users should just be able to write Python code and use the module names they're familiar with, and reserve the "u" prefix for modules that are micropython-specific and different to anything you'd see in CPython. I've been pushing on this for the built-in modules (https://github.com/micropython/micropython/pull/5377, https://github.com/micropython/micropython/pull/5229, and more recently https://github.com/micropython/micropython/issues/9018).

2. Packaging. See https://github.com/micropython/micropython-lib/pull/506 (and https://github.com/micropython/micropython/pull/8914) for the background here, but we're currently in the middle of trying to figure out how packaging works going forward, and in particular trying to unify the different ways that packages can be installed/deployed to devices.
The summary version is that for MicroPython, pypi adds a lot of complexity and potentially doesn't need to be part of how we install packages though any of the various mechanisms (on-device, via usb/uart, mass storage, or freezing).
So, I'd be curious to hear your thoughts on this, especially in terms of your intentions with this package and pypi and how it's used with microdot, and in particular whether you plan to use it from CPython.

miguelgrinberg:
1) a) the problem with `upyjwt` is that this will also become the import name, since this is a one file module. For that reason I think the name should be `ujwt`.
b) I really have no race in this horse, but I always thought it was a good idea to distinguish these minimalistic libraries from their CPython ones by adding the "u" prefix. In my view the "u" means you are using a simplified version. But I'm okay to rename it if that is what you prefer.
2) I made sure the code runs on CPython, but I do not intend for it to be required nor used on that platform, since it makes more sense to go with the standard PyJWT. Would you prefer that I remove the packaging files and add a manifest.txt file instead? It's fine by me, but this is now implemented yet. It may make more sense to commit as is, and then add the refactor to this other PR, maybe?

andrewleech:
The naming issue is annoying but I think important.

In this case, as-is, users might think it's a port of the jwt library which is incorrect. Would it be better to have the folder called pyjwt but the python file / installed module left as jwt for consistency with upstream?

The `u`/no-`u` question is a complicated one, in the past I too have thought it's "helpful" to differentiate the cut down module, however it complicates compatibility.

For example if someone else wants to release a package that works on both cpython and micropython and uses the jwt library, they need to sprinkler try/except everywhere in imports to check for both names.
It also just adds work for anyone porting examples / libraries that use `jwt`, meaning you'll need to fork them even just to change the import lines.

I'm all in favour of dropping the `u` when the implemented library is a compatible subset of the upstream library API.

One minor suggestion... while it's not (yet) common in this repo, a small readme in the folder might be worthwhile even just to link to the upstream module and their docs page?  If motivated you could add a note about the limitations of the module, what is/isn't covered - but yeah I haven't done this yet on packages I've worked on so definitely a case of glass houses...

jimmo:
OK thank you @miguelgrinberg & @andrewleech 

> I really have no horse in this race, but I always thought it was a good idea to distinguish these minimalistic libraries from their CPython ones by adding the "u" prefix. In my view the "u" means you are using a simplified version. But I'm okay to rename it if that is what you prefer.

I think what we're trying to get towards is the "u" indicates "this is a micropython-specific library and works differently to an existing Python library of the same name". Whereas when possible, we try and provide an actually compatible version of the Python library, even if it's a subset.

(i.e. compare old uasyncio which was inspired by but largely incompatible with CPython's vs the "new" v3 version, which we plan to rename to "asyncio").

> In this case, as-is, users might think it's a port of the jwt library which is incorrect. Would it be better to have the folder called pyjwt but the python file / installed module left as jwt for consistency with upstream?

Yes I was going to suggest exactly this. So we'd have `python-ecosys/pyjwt/jwt.py`

So on micropython you'd do `>>> upip.install('pyjwt')` or (soon) `$ mpremote install pyjwt`, and then code would `import jwt`, matching the Python experience as closely as possible.

@miguelgrinberg if you're happy with the above, then no need to update the PR, I can take care of it all in the merge (and sort out the manifest.py when I rebase #506).

miguelgrinberg:
@jimmo Yes, I really have no strong opinions on this one way or another. Whatever makes more sense with your current thinking is fine by me. Thanks!

jimmo:
Added in [9fdf046](https://github.com/micropython/micropython-lib/commit/9fdf046e2707356a009a732c53aaca316d5899bf) with the renaming as discussed plus some minor updates to the test to make it run on CPython with the "real" pyjwt.

I didn't include the hashlib change, instead let's solve this properly (see https://github.com/micropython/micropython-lib/pull/515). This means that the new pyjwt module won't work until this is merged, but we'll get that done ASAP.

Thanks @miguelgrinberg !

PR #513, opened 2022-08-02
Fix out of memory error in ESP 32

jaejeong1:
Delete the wrap_socket causing the problem

### DIFF
--- a/micropython/urllib.urequest/urllib/urequest.py
+++ b/micropython/urllib.urequest/urllib/urequest.py
@@ -28,8 +28,6 @@ def urlopen(url, data=None, method="GET"):
     s = usocket.socket(ai[0], ai[1], ai[2])
     try:
         s.connect(ai[-1])
-        if proto == "https:":
-            s = ussl.wrap_socket(s, server_hostname=host)
 
         s.write(method)
         s.write(b" /")


andrewleech:
I would expect these lines are needed for https support though?

sosi-deadeye:
This won't fix anything, introduces a bug and makes `https` for all unusable.

[Redirects are not supported yet](https://github.com/micropython/micropython-lib/blob/master/micropython/urllib.urequest/urllib/urequest.py#L62), so you can't be redirected accidentally to `https` if you're using `http`. If you're using `https` it will trigger the `OSError: [Errno 12] ENOMEM` exception on Boards with less memory.

mattytrentini:
Redirects have been implemented in `[urequests](https://github.com/micropython/micropython-lib/tree/master/python-ecosys/urequests)` (as well as a whole host of other fixes) and we should probably update `urlib.urequest` to use the same code.

jimmo:
@jaejeong1 thanks but this is not the right solution as the other comments have pointed out.

We are working to fix this issue in the ESP32 SSL implementation, see https://github.com/micropython/micropython/issues/8940 for the main issue tracking this.

PR #515, opened 2022-08-08
python-stdlib/hmac: Update to work with built-in hash functions.

jimmo:
This library was non-functional unless used with the micropython-lib
pure-Python implementation of hashlib, even if the device provides
sha1 and sha256.

This updates hmac to be significantly more RAM efficient (removes the
512-byte table), and functional with the built-in hash functions.

The only unsupported function is "copy", but this is non-critical, and now
fails with a NotSupportedError.


jimmo:
Fixes #369

dpgeorge:
Thanks!  This was on my list for a long time....

Merged in f95260d7e3c123a8b1907ded095e4c25e4251085 with a minor fix, `_binascii` -> `binascii`.

PR #521, opened 2022-08-31
Feature/calendar

pi-myt:


andrewleech:
Nice work pulling this together!

On a structural note, the CI "Check code formatting / build (pull_request)" should normally be resolved by running `python3 ./tools/codeformat.py ./python-stdlib/calendar/*.py` etc. You'll need `pip install black` for this.

Similarly eventually could you split the work into 3 commits - one for each package? There's a particular git commit format used here, hopefully it'll be easy enough to see from the git log: https://github.com/micropython/micropython-lib/commits/master

I'll take a look at the code properly when I've got a little more time :-)

pi-myt:
> On a structural note, the CI "Check code formatting / build (pull_request)" should normally be resolved by running python3 ./tools/codeformat.py ./python-stdlib/calendar/*.py etc. You'll need pip install black for this.

> Similarly eventually could you split the work into 3 commits - one for each package? There's a particular git commit format used here, hopefully it'll be easy enough to see from the git log: https://github.com/micropython/micropython-lib/commits/master

Thanks Andrew for the extra details. Aiming to get back to this soon to correct the formatting. I was looking at how to break it into 3 commits as there are dependencies between the packages. I could update unittest first, then datetime, then calendar can go in.

andrewleech:
Yeah no worries. To be clear they don't need to be separate PR's, separate commits on this branch is fine!

python-stdlib/calendar/calendar.py
```diff
@@ -0,0 +1,685 @@
...
+
+class IllegalWeekdayError(ValueError):
+    def __init__(self, weekday):
+        self.weekday = weekday
+    def __str__(self):
+        return "bad weekday number %r; must be 0 (Monday) to 6 (Sunday)" % self.weekday
+
+
+# Constants for months referenced later
+January = 1
```
mattytrentini, line 40: It's a good idea to apply [`micropython.const()`](https://docs.micropython.org/en/latest/library/micropython.html#micropython.const) to const values

python-stdlib/calendar/calendar.py
```diff
@@ -0,0 +1,685 @@
...
+    print(formatstring(cols, colwidth, spacing))
+
+
+def formatstring(cols, colwidth=_colwidth, spacing=_spacing):
+    """Returns a string formatted from n strings, centered within n columns."""
+    spacing = spacing * ' '
+    return spacing.join(c.center(colwidth) for c in cols)
+
+
+EPOCH = 1970
```
mattytrentini, line 586: This could be a problem; on some ports the epoch is 1970, others it's 2000. 
Read [Damien's summary](https://github.com/micropython/micropython/issues/5969#issuecomment-623053484) and micropython/micropython#6873 for background.

Should be able to use something like `time.gmtime(0)[0]` to retrieve the epoch year.

pi-myt: At the moment,. the EPOCH is only used for calendar.timegm(tuple) to calculate the Unix timestamp. I believe this needs to stay as 1970 so the calculated seconds match expectations. (do we have data size issues?)

I could move this the EPOCH values into the function, so it isn't visible to users of Calendar and can't be misused for systems with an EPOCH of 2000. If timegm() is not a required function, we could remove it from Calendar.

PR #523, opened 2022-09-03
urequests: Detect specific header field names caselessly

Hagb:
HTTP header field names are case-insensitive ([RFC 1945 section 4.2](https://www.rfc-editor.org/rfc/rfc1945#section-4.2)).

### DIFF
--- a/python-ecosys/urequests/urequests.py
+++ b/python-ecosys/urequests/urequests.py
@@ -137,10 +137,11 @@ def request(
             if not l or l == b"\r\n":
                 break
             # print(l)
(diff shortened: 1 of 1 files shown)


pablogventura:
Thanks for this. The `urequests` path has since moved to `requests`, and the header loop also changed with chunked support (#1135), so this branch no longer applies cleanly.

I opened #1147 with the same `lowerl` once-per-line approach on current `requests`, intended to supersede this PR.

dpgeorge:
Closing in favour of #1147.

PR #524, opened 2022-09-06
micropython/aiorepl: Add an asynchronous REPL.

jimmo:
This provides an async REPL with the following features:
- Run interactive REPL in the background.
- Execute statements using await.
- Simple history.

See documentation and example here https://github.com/jimmo/micropython-lib/tree/aiorepl/micropython/aiorepl

### DIFF
--- a/micropython/aiorepl/README.md
+++ b/micropython/aiorepl/README.md
@@ -1,0 +1,95 @@
+# aiorepl
+
+This library provides "asyncio REPL", a simple REPL that can be used even
+while your program is running, allowing you to inspect program state, create
+tasks, and await asynchronous functions.
+
+This is inspired by Python's `asyncio` module when run via `python -m asyncio`.
+
+## Background
+
+The MicroPython REPL is unavailable while your program is running. This
+library runs a background REPL using the asyncio scheduler.
+
+Furthermore, it is not possible to `await` at the main REPL because it does
+not know about the asyncio scheduler.
+
+## Usage
+
+To use this library, you need to import the library and then start the REPL task.
+
+For example, in main.py:
+
+```py
+import uasyncio as asyncio
+import aiorepl
+
+async def demo():
+    await asyncio.sleep_ms(1000)
+    print("async demo")
+
+state = 20
+
+async def task1():
+    while state:
+        #print("task 1")
+        await asyncio.sleep_ms(500)
+    print("done")
+
+async def main():
+    print("Starting tasks...")
+
+    # Start other program tasks.
+    t1 = asyncio.create_task(task1())
+
+    # Start the aiorepl task.
+    repl = asyncio.create_task(aiorepl.task())
+
+    await asyncio.gather(t1, repl)
+
+asyncio.run(main())
+```
+
+The optional globals passed to `task([globals])` allows you to specify what
+will be in scope for the REPL. By default it uses `__main__`, which is the
+same scope as the regular REPL (and `main.py`). In the example above, the
+REPL will be able to call the `demo()` function as well as get/set the
+`state` variable.
+
+Instead of the regular `>>> ` prompt, the asyncio REPL will show `--> `.
+
+```
+--> 1+1
+2
+--> await demo()
+async demo
+--> state
+20
+--> import myapp.core
+--> state = await myapp.core.query_state()
+--> 1/0
+ZeroDivisionError: divide by zero
+--> def foo(x): return x + 1
+--> await asyncio.sleep(foo(3))
+--> 
+```
+
+History is supported via the up/down arrow keys.
+
+## Cancellation
+
+During command editing (the "R" phase), pressing Ctrl-C will cancel the current command and display a new prompt, like the regular REPL.
+
+While a command is being executed, Ctrl-C will cancel the task that is executing the command. This will have no effect on blocking code (e.g. `time.sleep()`), but this should be rare in an asyncio-based program.
+
+Ctrl-D at the asyncio REPL command prompt will terminate the current event loop, which will stop the running program and return to the regular REPL.
+
+## Limitations
+
+The following features are unsupported:
+
+* Tab completion is not supported (also unsupported in `python -m asyncio`).
+* Multi-line continuation. However you can do single-line definitions of functions, see demo above.
+* Exception tracebacks. Only the exception type and message is shown, see demo above.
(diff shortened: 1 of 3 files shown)


ThinkTransit:
This is brilliant, very nice, testing on ESP32-S3 currently.

andrewleech:
My initial test of this was effortlessly delightful and incredibly powerful. Thanks @jimmo

jimmo:
Thanks @ThinkTransit & @andrewleech !

I've raised a tiny PR to the main repo (micropython/micropython#9295) which makes the REPL run by default in the context of `__main__` (i.e. the same as the real REPL), rather than having to awkwardly pass `globals()` to `aiorepl.task()`.

There's a couple of little issues to iron out (Ctrl-D is a bit flakey) but as soon as the PR above is merged I think this is probably good to go for version one.

ThinkTransit:
@jimmo  Might be a stupid question but how feasible/practical would it be to use a socket instead of sys.stdin/sys.stdout? Could this be used remotely over a network?

jimmo:
> @jimmo Might be a stupid question but how feasible/practical would it be to use a socket instead of sys.stdin/sys.stdout? Could this be used remotely over a network?

@ThinkTransit 

`aiorepl` just uses stdin/stdout, but MicroPython supports redirecting (or duplicating) stdin/stdout to any stream you like. So for example, WebREPL and the BLE REPL will work automatically with `aiorepl` (just as most boards run the REPL on both the USB CDC and a UART concurrently).

If instead you mean that you want the aiorepl to be distinct to the regular REPL and have its own input/output stream, then this is also feasible to implement (we could make it so `aiorepl.task()` takes an optional input and output stream), but it's a bit weird because for example `print()` would still go out to the regular stdout.

ThinkTransit:
> `aiorepl` just uses stdin/stdout, but MicroPython supports redirecting (or duplicating) stdin/stdout to any stream you like. So for example, WebREPL and the BLE REPL will work automatically with `aiorepl` (just as most boards run the REPL on both the USB CDC and a UART concurrently).
> 
> If instead you mean that you want the aiorepl to be distinct to the regular REPL and have its own input/output stream, then this is also feasible to implement (we could make it so `aiorepl.task()` takes an optional input and output stream), but it's a bit weird because for example `print()` would still go out to the regular stdout.

Thanks @jimmo 

I was asking about point 2 however after reading your response I now understand that this doesn't really make sense. Because aiorepl is on the standard REPL then I should just use the existing mechanisms WebREPL/BLE REPL to remotely access aiorepl!

I have been using aiorepl all day now, it is a massive help for debugging/monitoring.

dpgeorge:
@jimmo do you want to update this now that core supports `__dict__`?

jimmo:
> @jimmo do you want to update this now that core supports `__dict__`?

Done.

I've pushed an update that improves handling of Ctrl-C and Ctrl-D (based on conversation with @peterhinch).
- Ctrl-D now terminates the async REPL (and the event loop it was running in), taking you back to the regular REPL.
- Ctrl-C now cancels the executing task (via `task.cancel()`). This works really well for well-behaved async code, but will have no effect on blocking code. Hopefully nobody is executing blocking code in an async context, but perhaps at the REPL this is still useful. So I've implemented a very simple heuristic to detect whether the expression is async or not, and Ctrl-C will do a regular KeyboardInterrupt for non-async expressions.

In other words, you can Ctrl-C both `time.sleep(3)` and `await asyncio.sleep(3)` and the former will be `kbd_intr`, and the latter will be `task.cancel()`.

The way Ctrl-C works also prevents mpremote (and similar tools) from working because they expect to see the regular REPL after two successive Ctrl-C's. So there's now also a heuristic to detect this and terminate the async REPL.

peterhinch:
This version now lets me issue one-liners at the `-->` prompt and stop the "background" application with `ctrl-d`, returning me to the REPL.

So I got ambitious and ran this script at the `-->` prompt:
```python
import uasyncio as asyncio
import gc

async def main():
    while True:
        gc.collect()
        print(gc.mem_free())
        await asyncio.sleep(2)

asyncio.run(main())
```
This ran, with the "background" application also running, but I couldn't return to either prompt. On occasion `ctrl-c` briefly issued a `-->` but the foreground script kept running. Here is a typical session with comments added:
```
$ mpr
Connected to MicroPython at /dev/ttyACM0
Use Ctrl-] to exit this shell

>>> import gui.demos.aclock
Using encoder.
Analog clock demo.
Orientation: Horizontal. Reversal: False. Width: 320. Height: 240.
Start row = 0 col = 0
Starting asyncio REPL...
--> Screen.do_gc = False   # Clock demo is running and responds to this
--> import test
84368
84352   # Clock demo and test script both running
84352
--> 85296  # Issued ctrl-c Clock has stopped, test script still running
85296  # Only exit was to quit mpremote
85296
```

jimmo:
> So I got ambitious and ran this script at the --> prompt:

Does this mean that there's two instances of `asyncio.run()` running?

jimmo:
Merged in [7602843](https://github.com/micropython/micropython-lib/commit/7602843209ee4a23a7375dc483f58f8e1cce6ef0)

peterhinch:
> Does this mean that there's two instances of asyncio.run() running?

Yes, good point. I adapted my script to:
```python
import uasyncio as asyncio
import gc

async def main():
    while True:
        gc.collect()
        print(gc.mem_free())
        await asyncio.sleep(2)

asyncio.create_task(main())
```
Now `ctrl-c` still doesn't stop the script but `ctrl-d` does stop the aclock demo. Annotated session:
```
MicroPython v1.19.1 on 2022-09-25; Raspberry Pi Pico with RP2040
>>> 
>>> import gui.demos.aclock
Using encoder.
Analog clock demo.
Orientation: Horizontal. Reversal: False. Width: 320. Height: 240.
Start row = 0 col = 0
Starting asyncio REPL...
--> Free RAM 85712
Screen.do_gc = False  # Turn off Free RAM reports from aclock.py
--> import test
--> 85296  # test.py running
85296
85296
85296
  # ctrl-c issued
--> 85264
85264
85264
85264
85264
85264
  # ctrl-c issued
--> 85264
  # ctrl-d issued
>>>  
```
I also tried commenting out the `create_task` statement, importing test, and issuing `create_task` at the `-->` prompt, but the outcome was the same: I couldn't interrupt the script other than the nuclear `ctrl-d`.

PR #525, opened 2022-09-06
micropython: Migrate drivers/libraries from main repo to micropython-lib.

jimmo:
This adds all Python code from drivers/ and extmod/ in the main repo.

The idea is to make these libraries all available via `require("foo")` in a manifest, and also (soon) installable via upip/mpremote/etc.

### DIFF
--- a/micropython/drivers/bus/onewire/manifest.py
+++ b/micropython/drivers/bus/onewire/manifest.py
@@ -1,0 +1,1 @@
+module("onewire.py", opt=3)
--- a/micropython/drivers/bus/onewire/onewire.py
+++ b/micropython/drivers/bus/onewire/onewire.py
@@ -1,0 +1,92 @@
+# 1-Wire driver for MicroPython
+# MIT license; Copyright (c) 2016 Damien P. George
+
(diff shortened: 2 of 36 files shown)


jimmo:
Note: asyncio is left out of this PR because I want to move the docs over too, and first we need to set up how documentation from micropython-lib will be integrated into the main docs build.

dpgeorge:
We need to think of a standard for commit messages.  Eg if making changes to `lps22h` does the commit message start with `micropython:`, `micropython/drivers:`, `micropython/drivers/sensor/lps22h:` or `lps22h:`, or something else?

jimmo:
> We need to think of a standard for commit messages. Eg if making changes to `lps22h` does the commit message start with `micropython:`, `micropython/drivers:`, `micropython/drivers/sensor/lps22h:` or `lps22h:`, or something else?

`lps22h:`

Rationale: All the leaf directory names are unique.

(Except for unix-ffi, but then it can be e.g. `unix-ffi/re`)

dpgeorge:
Sounds good!

PR #526, opened 2022-09-06
Unittest regression fixes

stinos:
I finally came around to getting my micropython-lib fork in sync, turns out it broke all my tests :) Strange no-one noticed this.

### DIFF
--- a/unix-ffi/unittest/unittest.py
+++ b/unix-ffi/unittest/unittest.py
@@ -448,7 +448,7 @@ def _test_cases(mod):
             yield c
 
 
-def run_module(runner, module, path, top):
+def run_module(runner, module, *extra_paths):
     if not module:
         raise ValueError("Empty module name")
 
@@ -457,13 +457,12 @@ def run_module(runner, module, path, top):
     sys.modules.update(__modules__)
 
     sys_path_initial = sys.path[:]
-    # Add script dir and top dir to import path
-    sys.path.insert(0, str(path))
-    if top:
-        sys.path.insert(1, top)
+    for path in (path for path in reversed(extra_paths) if path):
+        sys.path.insert(0, str(path))
+
     try:
-        suite = TestSuite(module)
         m = __import__(module) if isinstance(module, str) else module
+        suite = TestSuite(m.__name__)
         for c in _test_cases(m):
             suite.addTest(c)
         result = runner.run(suite)
@@ -490,29 +489,37 @@ def main(module="__main__", testRunner=None):
     else:
         runner = TestRunner()
 
-    if len(sys.argv) <= 1:
-        result = discover(runner)
-    elif sys.argv[0].split(".")[0] == "unittest" and sys.argv[1] == "discover":
-        result = discover(runner)
+    if module is not None:
+        result = run_module(runner, module)
     else:
-        for test_spec in sys.argv[1:]:
-            try:
-                uos.stat(test_spec)
-                # test_spec is a local file, run it directly
-                if "/" in test_spec:
-                    path, fname = test_spec.rsplit("/", 1)
-                else:
-                    path, fname = ".", test_spec
-                modname = fname.rsplit(".", 1)[0]
-                result = run_module(runner, modname, path, None)
+        argn = len(sys.argv)
+        # See https://docs.python.org/3/library/unittest.html#test-discovery:
+        # python -m unittest is the equivalent of python -m unittest discover.
+        if not argn or (
+            sys.argv[0].split(".")[0] == "unittest"
+            and (argn == 1 or (argn >= 1 and sys.argv[1] == "discover"))
+        ):
+            result = discover(runner)
+        # Remaining arguments should be test specifiers.
+        else:
+            for test_spec in sys.argv[1:]:
+                try:
+                    uos.stat(test_spec)
+                    # test_spec is a local file, run it directly
+                    if "/" in test_spec:
+                        path, fname = test_spec.rsplit("/", 1)
+                    else:
+                        path, fname = ".", test_spec
+                    modname = fname.rsplit(".", 1)[0]
+                    result = run_module(runner, modname, path)
 
-            except OSError:
-                # Not a file, treat as import name
-                result = run_module(runner, test_spec, ".", None)
+                except OSError:
+                    # Not a file, treat as import name
+                    result = run_module(runner, test_spec, ".")
 
     # Terminate with non zero return code in case of failures
     sys.exit(result.failuresNum or result.errorsNum)
 
 
 if __name__ == "__main__":
-    main()
+    main(None)
--- a/unix-ffi/unittest/unittest.py
+++ b/unix-ffi/unittest/unittest.py
@@ -488,7 +488,8 @@ def split_path(path):
         return ".", path
 
 
-def main(module="__main__", testRunner=None):
+def main(module="__main__", **kwargs):
+    testRunner = kwargs.get("testRunner", None)
     if testRunner:
         if isinstance(testRunner, type):
             runner = testRunner()
--- a/unix-ffi/unittest/unittest.py
+++ b/unix-ffi/unittest/unittest.py
@@ -418,23 +418,23 @@ def run_suite(c, test_result: TestResult, suite_name=""):
 
     set_up_class()
 
-    if hasattr(o, "runTest"):
-        name = str(o)
-        run_one(o.runTest)
-        return
+    try:
+        if hasattr(o, "runTest"):
+            name = str(o)
+            run_one(o.runTest)
+        else:
+            for name in dir(o):
+                if name.startswith("test"):
+                    m = getattr(o, name)
+                    if not callable(m):
+                        continue
+                    run_one(m)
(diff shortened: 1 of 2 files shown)


andrewleech:
Thanks for the updates - when I originally tried to run "all tests in the repo" with the discover updates there were so many failures from a mixture of:
* dependency hell
* non-unittest based test scripts
* tests that have clearly been broken for a long time
I basically stopped bothering trying to work out which tests should/shouldn't pass and made it work with the test usages I had / could find locally.

I'm curious about the first commit here:
> Fix regression introduced by a7b2f6311 which broke the typical case of
> running a test script which has

> 
> if __name__ == "__main__":
>     unittest.main()

This usage seems to work fine for me as-is on current master?
```
STEP: ~/micropython-lib/unix-ffi/unittest $ micropython ./test_unittest.py 
test_NotChangedByOtherTest (test_unittest_isolated.TestUnittestIsolated) ... ok

Ran 1 tests

OK
testFail (test_unittest.TestUnittestAssertions) ... ok
testEqual (test_unittest.TestUnittestAssertions) ... ok
test_AlmostEqual (test_unittest.TestUnittestAssertions) ... ok
test_AlmostEqualWithDelta (test_unittest.TestUnittestAssertions) ... ok
testNotEqual (test_unittest.TestUnittestAssertions) ... ok
testIs (test_unittest.TestUnittestAssertions) ... ok
testIsNot (test_unittest.TestUnittestAssertions) ... ok
testIsNone (test_unittest.TestUnittestAssertions) ... ok
testIsNotNone (test_unittest.TestUnittestAssertions) ... ok
testTrue (test_unittest.TestUnittestAssertions) ... ok
testFalse (test_unittest.TestUnittestAssertions) ... ok
testIn (test_unittest.TestUnittestAssertions) ... ok
testIsInstance (test_unittest.TestUnittestAssertions) ... ok
testRaises (test_unittest.TestUnittestAssertions) ... ok
testSkip (test_unittest.TestUnittestAssertions) ... skipped: test of skipping
testAssert (test_unittest.TestUnittestAssertions) ... ok
testExpectedFailure (test_unittest.TestUnittestAssertions) ... ok
testExpectedFailureNot (test_unittest.TestUnittestAssertions) ... ok
test_NotChangedByOtherTest (test_unittest.TestUnittestAssertions) ... FAIL
test_subtest_even (test_unittest.TestUnittestAssertions) ... ok
testSetUpTearDownClass_1 (test_unittest.TestUnittestSetup) ... ok
testSetUpTearDownClass_2 (test_unittest.TestUnittestSetup) ... ok

FAIL: test_NotChangedByOtherTest <class 'TestUnittestAssertions'>
  File "/home/anl/micropython-lib/unix-ffi/unittest/unittest.py", line 393, in run_one
  File "<string>", line 1, in <module>
  File "./test_unittest.py", line 148, in test_NotChangedByOtherTest
AssertionError: 

Ran 22 tests

FAILED (failures=1, errors=0)
```

Can you suggest a reproduction of the issue you ran into? 
Your change looks clean, I'm just not following the usage of it.

unix-ffi/unittest/unittest.py
```diff
@@ -444,7 +444,7 @@ def _test_cases(mod):
         c = getattr(mod, tn)
         if isinstance(c, object) and isinstance(c, type) and issubclass(c, TestCase):
             yield c
-        elif tn.startswith("test_") and callable(c):
+        elif tn.startswith("test") and callable(c):
```
andrewleech, line 447: Sorry for breaking some tests with this change... I don't remember clearly now but I think I found some existing test scripts in the repo where there were testFoo functions being run that were not intended to be unittest functions, so thought the `_` made more sense. I believe this is non-standard functionality that cpython doesn't support anyway (finding & running bare test* functions)?

unix-ffi/unittest/unittest.py
```diff
@@ -490,19 +498,22 @@ def main(module="__main__", testRunner=None):
...
+    # See https://docs.python.org/3/library/unittest.html#test-discovery:
+    # python -m unittest is the equivalent of python -m unittest discover.
+    if not argn or (
+        sys.argv[0].split(".")[0] == "unittest"
+        and (argn == 1 or (argn >= 1 and sys.argv[1] == "discover"))
+    ):
         result = discover(runner)
+    # If there's one argument assume it's the path to the script which called main(), so run its tests.
+    elif argn == 1:
+        result = run_module(runner, module, split_path(sys.argv[0])[0], None)
```
andrewleech, line 511: `split_path(sys.argv[0])`
This is to split out just the filename isn't it... I would think this will fail if you're running a unittest from a different folder?

stinos: Yeah that's basically `os.path.dirname` . Not sure why that would fail, the point of the `path` argument is to add the script directory to `sys.path` right? So `python /path/to/test.py` is going to add `path/to`. I just realize this is not actually needed though, the directory of the executing script gets added to `sys.path` anyway. Moreover, as the comment says this is an assumption but thinking about it: this is just wrong and I'm going to remove it. If anything `argv[0]` could be used as prog name for argparse but that's it.

unix-ffi/unittest/unittest.py
stinos: > I believe this is non-standard functionality that cpython doesn't support anyway (finding & running bare test* functions)?

I think so yes, at least I didn't find anything which does that. I'd personally refrain from doing that because it's non-standard, and because if you have a file like that then decide it runs under CPython as well you might think its tests are all fine but they are just not being ran. CPython does support a single `runTest` function though like

```
class Test(unittest.TestCase):
  @staticmethod
  def runTest():
    test = unittest.TestCase()
    test.assertEqual(1, 1)
```

But anyway `test` is the standard prefix; e.g. [`By default these are the method names beginning with test`](https://docs.python.org/3/library/unittest.html#loading-and-running-tests) and `testMethodPrefix = 'test'` in `unittest/loader.py`.

stinos:
> Can you suggest a reproduction of the issue you ran into?

The output from your test actually shows the problem: it runs not just test_unittest, but both test_unittest and test_unittest_isolation, because it's just running running test discovery in the current directory. E.g. running `micropython ./unittest.py` will also run both tests (instead of nothing). Whereas going one directory up and running `micropython ./unittest/test_unittest.py` will discover all tests in unix_ffi. Or still another way: create a file

```
import unittest
import test_unittest

unittest.main(test_unittest)
```

in the unittest directory and it will simply ignore the argument to `main()` and instead run all tests in the directory.

Btw any idea why `test_NotChangedByOtherTest` fails on your machine?

Also: ping for my [other comment](https://github.com/micropython/micropython-lib/pull/488/files#r963787097)

stinos:
I went over the code again and fixed some things:
- basically if `main` is called with a module as argument it shouldn't look at `argv` at all, just run the module, so make it like that
- passing a module to `run_module` would fail because the module was passed to `TestSuit` but that expects a string
- force using keywords with `main` (not sure if that one is really needed)

Note the code still has the equivalent of `if len(sys.argv) == 0: discover()` because the original had that. but I'm not sure what it would mean for sys.argv to be empty?

andrewleech:
> The output from your test actually shows the problem: it runs not just test_unittest, but both test_unittest and test_unittest_isolation, because it's running test discovery in the current directory.

Ah I see now. That'll be why I hadn't noticed the bug - at first glance it looks like it's still working. But it's not doing the right thing certainly, and pulls in the dependency when it wasn't intended to. Nice fix, thanks!

andrewleech:
> Btw any idea why `test_NotChangedByOtherTest` fails on your machine?

That I'm not sure about, didn't get a chance to investigate when I ran that this morning.

unix-ffi/unittest/unittest.py
```diff
@@ -490,29 +498,34 @@ def main(module="__main__", testRunner=None):
     else:
         runner = TestRunner()
 
-    if len(sys.argv) <= 1:
-        result = discover(runner)
```
andrewleech, line 494: > Note the code still has the equivalent of `if len(sys.argv) == 0: discover()` because the original had that. but I'm not sure what it would mean for sys.argv to be empty?

Do you mean this original? This was intended to catch the `len==1` mostly, main case of this is `micropython -m unittest` where `sys.argv == ['unittest']`.

It's true I don't know if there's ever a case where `len(sys.argv) == 0` unless you're manipulating `sys.argv` manually in a wrapper script, in which case all bets are off really. Not sure if I had a specific reason to make this `<= 1` rather than just `== 1` other than maybe not wanting to leave an undefined case statement.

stinos: > Do you mean this original?

Yes exactly.

But without the check it would be unconditionally indexing `sys.argv` which is also not very nice so I'll just leave it as-is.

jimmo:
Thanks @stinos . I think all of this should be addressed by https://github.com/micropython/micropython-lib/pull/527#issuecomment-1242132100

Please let me know if not.

dpgeorge:
Should be fixed in 796a5986cd65e8f2c76d06267b92497ff602eca6

Issue #328, opened 2019-02-07
wait_msg() does not allways return a value


sulimma:
wait_msg() does not return a value when successfull.
I suggest to add a single line at the end of that method:
return op

This would inform the caller whether a message was processed.
This way a caller could repeatedly call this method until no more messages are queued and then wait for a while.

dpgeorge:
Implemented in b50d3462d783e4aab2f10d6b8117641244918f64

PR #528, opened 2022-09-06
wait_msg() return op as indication if there was a message

meirarmon:
As indicated in this issue: https://github.com/micropython/micropython-lib/issues/328, it would be useful if wait_msg() would return a value if a message was received.
A common use case using this library would be:

```
def callback(topic, msg):
    print(topic, msg)


async def check_messages(mqtt_client, interval):
    while True:
        mqtt_client.check_msg()
        await asyncio.sleep(interval)
```

In the above scenario, if more than one message was published to the topic being subscribed to, it will still get consumed every <interval> seconds. If this change is merged, we would be able to do:

```
def callback(topic, msg):
    print(topic, msg)


async def check_messages(mqtt_client, interval):
    while True:
        while mqtt_client.check_msg() is not None:
            continue
        await asyncio.sleep(interval)
```
And any waiting messages will get consumed immediately.

### DIFF
--- a/micropython/umqtt.simple/umqtt/simple.py
+++ b/micropython/umqtt.simple/umqtt/simple.py
@@ -206,6 +206,7 @@ class MQTTClient:
             self.sock.write(pkt)
         elif op & 6 == 4:
             assert 0
+        return op
 
     # Checks whether a pending message from server is available.
     # If not, returns immediately with None. Otherwise, does


dpgeorge:
Thanks for the contribution, this is a nice and simple improvement.  Rebased and merged in b50d3462d783e4aab2f10d6b8117641244918f64

PR #529, opened 2022-09-06
weakref: Add basic stub function to aid in porting.

andrewleech:
This just allows most basic usages of weakref in other modules to "work" unless they test for actual cleanup via gc.collect etc.

Docs: https://docs.python.org/3/library/weakref.html#weakref.ref

### DIFF
--- a/python-stdlib/weakref/weakref.py
+++ b/python-stdlib/weakref/weakref.py
@@ -1,0 +1,11 @@
+"""
+# weakref
+https://docs.python.org/3/library/weakref.html
+
+Micropython does not have support for weakref in the VM, so currently this is a simple stub
+module that directly assigns the object reference to simplify porting of other libraries.
+"""
+
+
+def ref(obj):
+    return lambda: obj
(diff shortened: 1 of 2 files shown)


python-stdlib/weakref/weakref.py
```diff
@@ -0,0 +1,16 @@
+"""
+# weakref
+https://docs.python.org/3/library/weakref.html
+
+The weakref module allows the Python programmer to create weak references to objects.
+In the following, the term referent means the object which is referred to by a weak reference.
+A weak reference to an object is not enough to keep the object alive: when the only remaining references to a referent are weak references, garbage collection is free to destroy the referent and reuse its memory for something else. However, until the object is actually destroyed the weak reference may return the object even if there are no strong references to it.
+A primary use for weak references is to implement caches or mappings holding large objects, where it’s desired that a large object not be kept alive solely because it appears in a cache or mapping.
```
dpgeorge, line 8: Is this copied from CPython?  If so I think just remove it, keep only the link to CPython's weakref docs, and the paragraph below about lack of support in MicroPython.

dpgeorge:
MicroPython now supports the `weakref` module.  So I think this can be closed.

PR #532, opened 2022-09-12
aioble: Split into optional components.

jimmo:
This replaces the options that could be specified previously to include and require.

The `aioble` package provides the same as the default options did previously. Optional dependencies can now be require()'ed explicitly.

### DIFF
--- a/micropython/bluetooth/aioble-central/manifest.py
+++ b/micropython/bluetooth/aioble-central/manifest.py
@@ -1,0 +1,5 @@
+metadata(version="0.1.0")
+
+require("aioble-core")
+
+package("aioble", files=("central.py",), base_path="../aioble")
--- a/micropython/bluetooth/aioble/manifest.py
+++ b/micropython/bluetooth/aioble/manifest.py
@@ -1,27 +1,11 @@
-_files = (
-    "__init__.py",
-    "core.py",
-    "device.py",
-)
+# This directory contains all aioble code, but the manifest itself just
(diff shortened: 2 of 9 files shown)


micropython/bluetooth/aioble-central/manifest.py
```diff
@@ -0,0 +1,7 @@
+metadata(version="0.1.0")
+
+require("aioble-core")
+
+package("aioble", files=(
+    "central.py",
+), base_path="../aioble")
```
dpgeorge, line 7: Black wants this to be on one line...

micropython/bluetooth/aioble/manifest.py
```diff
@@ -1,27 +1,11 @@
...
-    _files += ("server.py",)
-
-if options.l2cap:
-    _files += ("l2cap.py",)
-
-if options.security:
-    _files += ("security.py",)
-
-package("aioble", files=_files)
+require("aioble-core")
```
dpgeorge, line 9: This requirement of `aioble-core` is not strictly needed.

micropython/bluetooth/aioble-central/manifest.py
jimmo: Fixed.

micropython/bluetooth/aioble/manifest.py
jimmo: Replaced with a comment explaining what's going on.

dpgeorge:
Looks good!

PR #533, opened 2022-09-12
python-stdlib: Rename extension modules to name-extension.

jimmo:
This applies to `os.path`, `collections.defaultdict`, and `collections.deque`.


jimmo:
Also fixes a CPython compatibility issue where os automatically includes os.path (so do that in our version if available).

dpgeorge:
Looks good!

andrewleech:
It'd be good to add a note about this extension convention in the readme, currently there's a paragraph:
> Other libraries are packages, in which case you'll need to copy the directory instead. For example, to add collections.defaultdict, copy collections/collections/__init__.py and collections.defaultdict/collections/defaultdict.py to a directory named lib/collections on your device.

That would be a good section to reword to suit?

jimmo:
> It'd be good to add a note about this extension convention in the readme, currently there's a paragraph:

Yes, definitely. I plan a more significant documentation update when this is all done, so will include that then.

jimmo:
Merged in c262628a41a89cc7d417d57cd0070909c95419f6 through ad9309b669cd4474bcd4bc0a67a630173222dbec.

PR #536, opened 2022-09-16
aioble: Ordered handling of char.written().

andrewleech:
On a ble peripheral system with a "significant" number of background asyncio tasks, reception & handling concurrent characteristic writes (eg. a looping task with `await characteristic.written()` can occur out of order.

For some services this can severly impact function, where writes to multiple characteristics to perform functions is critical. For example, in the Object Transfer Service it's common to "set a list filter" on one char, then "send the next list item" on a separate char. If these are handled out of order incorrect results are sent.

Essentially, if the vm / asyncio loop is slowed down by a background task (in this case a `time.sleep()`) while two ble write operations are received by the stack, the asyncio.Event for both characteristics can be set before either aioble characteristic event handler services them.

In this case, the `characteristic.written()` event of which ever task is checked first (whichever one was originally registered first) will be serviced first, regardless of which one came in first.

The first commit in this PR adds a multitest that demonstrates this behavior, failing as some of the messages arrive out of order.

The second commit addresses this issue by re-working the existing `Characteristic(capture=True)` functionality to work through a single shared `deque` which is handled by a background async task. This task pops individual incoming write events from the shared queue and forwards them to the respective characteristic event sequentially. Only once that `characteristic.written()` handles the event does the manager task continue to any other events in the shared queue.

Characteristics configured with `capture=False` (default) still behave the same as previously.

---

### DIFF
--- a/micropython/bluetooth/aioble/aioble/server.py
+++ b/micropython/bluetooth/aioble/aioble/server.py
@@ -97,6 +97,33 @@ class BaseCharacteristic:
         else:
             ble.gatts_write(self._value_handle, data, send_update)
 
+    async def _init_shared_write_task(self):
+        # Should be run on first call of an async task waiting on .written()
+        # Will create the needed events and start the background queue monitor task
+        if not hasattr(BaseCharacteristic, "_shared_write_event"):
+            BaseCharacteristic._shared_write_event = asyncio.ThreadSafeFlag()
+            BaseCharacteristic._shared_written_event = asyncio.ThreadSafeFlag()
+
+        asyncio.create_task(self._shared_write_task())
+
+    async def _shared_write_task(self):
+        # Monitor the shared queue for incoming characteristic writes and forward
+        # them sequentially to the individual characteristic events.
+        ie = BaseCharacteristic._shared_write_event
+        oe = BaseCharacteristic._shared_written_event
+        q = BaseCharacteristic._shared_write_queue
+
+        while True:
+            if len(q):
+                conn, data, characteristic = q.popleft()
+                characteristic._write_detail = (conn, data)
+                characteristic._write_event.set()
+                # Wait for the characteristic event to be handled before continuing
+                await oe.wait()
+
+            if not len(q):
+                await ie.wait()
+
     # Wait for a write on this characteristic. Returns the connection that did
     # the write, or a tuple of (connection, value) if capture is enabled for
     # this characteristics.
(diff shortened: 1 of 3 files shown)


dpgeorge:
Just to be clear, by "out of order characteristic writes" you mean separate (independent!) characteristics appear written in a different order on the server, compared to the order they were written to on the client?  Eg client writes to A then B, but server sees B update first then A?

It seems to me like there is no guarantee of the ordering of writes to independent characteristics, precisely because they are independent.

Or is the issue that the stack (nimble) does get the writes in the "correct" order (eg A then B) but aioble swaps them around?

andrewleech:
Yes client writes to charA then charB, server nimble stack gets them in order, but once they go via asyncio events they can be handled out of order by tasks waiting on their respective events.

The stack itself guarantees the order of the writes, and some standard services rely on the order of these writes - indeed OTS requires it in a number of situations.

andrewleech:
While the fix presented in this MR is very specific to this particular user case, the question on event order might be worth thinking about more broadly for asyncio in general? Probably more of a potential issue if/when deeper asyncio hardware support is added, eg multiple tasks async waiting on pin /switch press, needing patterns of usage so they get handled in order etc.

jimmo:
Thanks @andrewleech -- what you're describing makes sense. GATT itself will preserve the event order, and aioble doesn't make any attempt to preserve it across characteristics. I hadn't considered that there would be ordering dependencies across capture=True characteristics.

I will need to spend some time reviewing this, the capture stuff is pretty subtle. I'm glad you were able to keep the same API.

I'm a little bit confused why there needs to be both a global queue and a per-characteristic queue. Will the per-characteristic queue ever have more than one item in it -- it seems to me that the monitor task will always wait for the characteristic to dispatch before being able to append a new item?

Another thing to consider (although doesn't have to be this PR), is that we also do the same capture queue thing for notifications. Should we apply the same fix there?

andrewleech:
Thanks @jimmo yeah I wasn't 100% convinced I should hijack the existing capture arg however thought that'd be safer that maintaing / documenting two separate queueing mechanisms in the library.

> I'm a little bit confused why there needs to be both a global queue and a per-characteristic queue

I initially kept this to reduce the size of this patch, don't change what's not broke etc - particularly when I saw other usages of `deque(1)` in other places. However I've now stripped it back to just an single attribute on the Char class and it does simplify things a fair bit.

> Another thing to consider (although doesn't have to be this PR), is that we also do the same capture queue thing for notifications. Should we apply the same fix there?

Ah, yes it would certainly make sense to have the same pattern there, I can easily foresee situations where the order of notifications is important.

micropython/bluetooth/aioble/aioble/server.py
```diff
@@ -97,6 +102,33 @@ def write(self, data, send_update=False):
         else:
             ble.gatts_write(self._value_handle, data, send_update)
 
+    async def _init_shared_write_task(self):
+        # Should be run on first call of an async task waiting on .written()
+        # Will create the needed events and start the background queue monitor task
+        if BaseCharacteristic._shared_write_event is None:
```
dpgeorge, line 108: Might be possible to change this to `hasattr(...)` and then remove the `..=None` settings in the global class scope above.

andrewleech: I often struggle with questions like this when it's a bit of a ram vs runtime cost, however in this case yeah - the current ram cost of having that attr "preallocated" is paid by people not using the feature, whereas a hasattr call vs "attr lookup and compare to none" would be nigh on identical runtime cost. I'll make the change thanks!

jimmo:
Thanks @andrewleech this is a really good simplification and much easier to reason about.

Merged with some minor changes in https://github.com/micropython/micropython-lib/commit/eba897420d96650a372b4b65831c24741d667890

PR #541, opened 2022-09-27
python-ecosys/kpn-senml: Add KPN SenML library.

iabdalkader:


iabdalkader:
@jimmo If this is merged, and the `require`'d in a `manifest.py`, can I `import kpn_senml` or need to move it up a level ?

dpgeorge:
How does this relate to https://github.com/kpn-iot/senml-python-library and https://github.com/kpn-iot/senml-micropython-library, is it just a copy of the latter?  If so, then I don't think adding it verbatim here is the right way to go.  Instead one can just use the github URL to reference this library.

(All the examples have a boot.py and pymakr.conf which are really not necessary.)

dpgeorge:
Out of interest, what do you need this library for?  Are there other libraries that could be used instead?

> Both libraries seem unmaintained and they don't respond to PRs/issues, and as of now we don't have a way to freeze remote modules in `manifest.py`. If that's implemented soon, we can fork that library instead and freeze it.

I think that's the best solution at this point in time: to fork that repo, apply your patches as new commits, then implement freezing of remote URLs in manifest.py.

iabdalkader:
> Out of interest, what do you need this library for? Are there other libraries that could be used instead?

It's used by cloud clients to parse/create CBOR/JSON. And not aware of another library, but this one works really well.

> I think that's the best solution at this point in time: to fork that repo, apply your patches as new commits, then implement freezing of remote URLs in manifest.py.

I don't mind that. My end goal is just to make devices capable of connecting to the cloud out of the box, and for that I need this library, #508 for time and improved logging in #507 and the cloud client library which won't be in `micropython-lib` anyway, but we don't have remote freezing yet. Any ETA on that @jimmo ?

I think #508 and #507 are generic enough to be in `micropython-lib` though.

dpgeorge:
We need to consider this as two separate components: cbor and senml.  And for both of them we need to consider whether the library is trying to be compatible with an existing library, or if it's a custom MicroPython library.

For cbor: there is an existing `cbor` library on PyPI but it looks abandoned.  Instead there is `cbor2` which had a commit 8 days ago (see https://github.com/agronholm/cbor2).  It looks like the cbor code here (from KPN) is trying to be compatible with `cbor2`.  So if we decide to keep that compatibility then our version should be `python-ecosys/cbor2`.  And I think that's what we should do.

For senml: aside from `kpn_senml`, there doesn't seem to be any other maintained senml library.  If we want to maintain compatibility with `kpn_senml` then our library should be `python-ecosys/kpn_senml`.  Otherwise it should be `micropython/senml` and we are free to change the API.  I would suggest the latter (call it `micropython/senml`) because `kpn_senml` has not been updated in a year, and the PyPI package has not been updated in about 4 years.  Also I think we can remove the special KPN names, we don't need them and they take up space.

jimmo:
Agree to both -- cbor2 and rename to senml, remove all kpn references.

iabdalkader:
> Also I think we can remove the special KPN names, we don't need them and they take up space.
> remove all kpn references

@dpgeorge @jimmo By remove them, you mean rename `KPN_SENML_*` to maybe `SENML_NAME*` (to match `SENML_UNIT_*` ? Because they're used in the examples, if I remove them will have to also remove all the examples that reference them.

iabdalkader:
Hi, what other changes are needed here to move forward with this ?

iabdalkader:
@dpgeorge @jimmo Hi! We need this hosted somewhere so we can link to it and update the imports (`kpn_senml` to `senml` etc..) before publishing our library which needs to be out very soon, what's blocking this PR ?

python-ecosys/cbor2/manifest.py
```diff
@@ -0,0 +1,3 @@
+metadata(version="0.0.1")
```
dpgeorge, line 1: I think we can start at version 0.1.0

python-ecosys/cbor2/cbor2/decoder.py
```diff
@@ -0,0 +1,252 @@
...
+break_marker = object()
+
+
+class CBORSimpleValue(object):
+    """
+    Represents a CBOR "simple value".
+    :param int value: the value (0-255)
+    """
+
+    __slots__ = "value"
```
dpgeorge, line 30: MicroPython doesn't have slots, so this line can be deleted.

python-ecosys/cbor2/cbor2/decoder.py
```diff
@@ -0,0 +1,252 @@
...
+    31: lambda self: break_marker,
+}
+
+
+class CBORDecoder(object):
+    """
+    Deserializes a CBOR encoded byte stream.
+    """
+
+    __slots__ = "fp"
```
dpgeorge, line 190: This line can be deleted.

python-ecosys/cbor2/cbor2/encoder.py
```diff
@@ -0,0 +1,172 @@
...
+    dict: encode_map,
+}
+
+
+class CBOREncoder(object):
+    """
+    Serializes objects to a byte stream using Concise Binary Object Representation.
+    """
+
+    __slots__ = "fp"  # keep the ram usage down by specifying the allowed fields and nothing more for this class
```
dpgeorge, line 126: This line can be deleted.

micropython/senml/README.md
```diff
@@ -0,0 +1,15 @@
+# Introduction
+
+The KPN SenML library helps you create and parse [senml documents](https://tools.ietf.org/html/draft-ietf-core-senml-13)
```
dpgeorge, line 3: There shouldn't be any references to KPN in the code or docs.  Eg reword this as "The SenML library ...".

micropython/senml/README.md
```diff
@@ -0,0 +1,15 @@
...
+in both json and cbor format.
+
+# key features
+
+- Object oriented design.
+- built in support for [senml's unit registry](https://tools.ietf.org/html/draft-ietf-core-senml-12#section-12.1)
+- extensible for new data types
+- makes use of (but doesn't restrict to) KPN's predefined list of record names.
```
dpgeorge, line 11: I don't think we should include KPN's predefined list of record names.  As per the discussion in the main thread of this PR, this takes up space and is very specificy to KPN.  Users can define and use their own record names.

micropython/senml/README.md
```diff
@@ -0,0 +1,15 @@
...
+- direct support to read/write in json and cbor format.
+- automatically adjusts record data with respect to base time, base value & base sum.
+
+Please visit our [docs site](https://kpn-iot.github.io/senml-micropython-library) for more info.
```
dpgeorge, line 15: This link is not relevant, so probably just remove this line.

micropython/senml/docs/index.md
```diff
@@ -0,0 +1,14 @@
+Welcome to the API documet site for the micro-python version of KPN's SenML library.
```
dpgeorge, line 1: Need to reword this to remove KPN reference.

micropython/senml/docs/index.md
```diff
@@ -0,0 +1,14 @@
...
+
+Check out the [detailed docs](https://kpn-iot.github.io/senml-library/) for an in-depth explanation of all the features in the library.
```
dpgeorge, line 3: Need to remove this line.

micropython/senml/senml/senml_kpn_names.py
```diff
@@ -0,0 +1,46 @@
...
+# License: MIT license.
+# Author: Jan Bogaerts
+#
+# sensor names
+
+from senml.senml_unit import enum
+
+
+SenmlNames = enum(
+    KPN_SENML_PRESSURE="pressure",
```
dpgeorge, line 17: As mentioned elsewhere, IMO this file can be removed.  Users can define their own names simply as strings.

micropython/senml/manifest.py
```diff
@@ -0,0 +1,5 @@
+metadata(version="0.0.1")
```
dpgeorge, line 1: I think this can start at 0.1.0

micropython/senml/examples/base.py
```diff
@@ -0,0 +1,35 @@
...
+#
+# base value example
+
+from senml import *
+import utime as time
+
+
+pack = SenmlPack("device_name")
+temp = SenmlRecord(
+    SenmlNames.KPN_SENML_TEMPERATURE, unit=SenmlUnits.SENML_UNIT_DEGREES_CELSIUS, value=23.5
```
dpgeorge, line 19: This `KPN_SENML_TEMPERATURE` constant can be replaced simply with the string `"temperature"`.  Then there's no need for the special KPN names (which don't really have any official meaning, as far as I know).

dpgeorge:
> By remove them, you mean rename KPN_SENML_* to maybe SENML_NAME* (to match SENML_UNIT_* ? Because they're used in the examples, if I remove them will have to also remove all the examples that reference them.

I suggest to just remove all `KPN_SENML_*` names.  The examples can then simply use strings with the appropriate string value.  I don't think there's anything special about the string (the units on the other hand seem to be "official", so can stay as an enum).

micropython/senml/README.md
iabdalkader: What about the license header in `*.py` files ?

iabdalkader:
> I suggest to just remove all KPN_SENML_* names. The examples can then simply use strings with the appropriate string value. 

Removed the file and and replaced all references with strings, I also updated the license headers, how does it look now ?

I just have one concern, do we need the `py` files nested like `cbor2/cbor2/decoder.py` or can we move them up one level ?

micropython/senml/README.md
iabdalkader: I updated all the license headers.

dpgeorge:
> I just have one concern, do we need the `py` files nested like `cbor2/cbor2/decoder.py` or can we move them up one level ?

Yes they need to be nested, because it's a package.  So you have it correct.

micropython/senml/examples/actuator.py
```diff
@@ -0,0 +1,68 @@
...
+
+json_data = (
+    '[{"bn": "device_name", "n":"actuator", "v": 20 }, {"n": "another_actuator", "vs": "a value"}]'
+)
+print(json_data)
+pack.from_json(json_data)
+
+print('[{"bn": "device_name", "n":"temp", "v": 20, "u": "Cel" }]')
+# this represents the cbor json struct: [{-2: "device_name", 0: "temp", 1: "Cel", 2: 20}]
+cbor_data = binascii.unhexlify("81A4216B6465766963655F6E616D65006474656D70016343656C0214")
```
dpgeorge, line 67: For this use of binascii, it would be simpler to use the now-supported `bytes.fromhex()`.

micropython/senml/examples/basic_cbor.py
```diff
@@ -0,0 +1,43 @@
...
+pack = SenmlPack("device_name")
+
+while True:
+    with SenmlRecord(
+        "test", value=10
+    ) as rec:  # use a with statement to automatically remove the item from the list when it goes out of scope, generate a value for the record
+        pack.add(rec)
+        cbor_val = pack.to_cbor()
+        print(cbor_val)
+        print(ubinascii.hexlify(cbor_val))
```
dpgeorge, line 41: Could use `cbor_val.hex()` (unless there is a mistmatch of str/bytes bytes here).

python-ecosys/cbor2/examples/cbor_test.py
```diff
@@ -0,0 +1,40 @@
...
+import ubinascii
+
+input = [
+    {"bn": "urn:dev:ow:10e2073a01080063", "u": "Cel", "t": 1.276020076e09, "v": 23.5},
+    {"u": "Cel", "t": 1.276020091e09, "v": 23.6},
+]
+
+data = encoder.dumps(input)
+print(data)
+print(ubinascii.hexlify(data))
```
dpgeorge, line 38: Could use `data.hex()`.

micropython/senml/LICENSE
```diff
@@ -0,0 +1,22 @@
+MIT License
```
dpgeorge, line 1: Now that this license is copied at the top of each file, this LICENSE file itself can be removed (otherwise it's confusing why it's here, no other packages in this repo have one, I don't think).

python-ecosys/cbor2/LICENSE
```diff
@@ -0,0 +1,22 @@
+MIT License
```
dpgeorge, line 1: As above, can remove this file.

micropython/senml/examples/actuator.py
```diff
@@ -0,0 +1,68 @@
+"""
+The MIT License (MIT)
+
+Copyright (c) 2018 KPN (Jan Bogaerts)
+Copyright (c) 2023 MicroPython
```
dpgeorge, line 5: For this copyright line, please use either your name or Arduino company name (not "MicroPython").

dpgeorge:
For `cbor2/__init__.py`, I think it should contain imports of the decoder and encoder files, so it matches how other packages (eg `senml`) work.

micropython/senml/examples/basic_cbor.py
iabdalkader: @dpgeorge I have no idea, I never ran those examples, but I changed it as you suggested.

iabdalkader:
> For cbor2/__init__.py, I think it should contain imports of the decoder and encoder files, so it matches how other packages (eg senml) work.

Done.

dpgeorge:
Merged!!

PR #542, opened 2022-09-28
micropython/mip: Add a new `mip` library for on-device installation.

jimmo:
This PR adds:
 - A library (`mip`) that can be frozen (or manually installed) on network-capable boards for on-device installation of packages from micropython-lib (or other similar package indices).
 - A tool (`tools/build.py`) to deploy micropython-lib to a static web server suitable for access from `mip` (or `mpremote` when similar functionality is added).

The structure of the static web server output is explained in the comments at the top of deploy.py.

`mip.install()` supports the following use cases:

- `mip.install("foo")` -- this will download the latest version of the foo package (and dependencies) from micropython-lib.
- `mip.install("foo", version="x.y")` -- install foo version x.y, including dependencies at the version when foo x.y was published
- `mip.install("foo", target="path/to/lib")` -- install foo and dependencies in specified path (otherwise defaults to the first "lib" directory in sys.path)
- `mip.install("foo", index="https://example.com/")` -- install foo from example.com's index
- `mip.install("http://example.com/x/y/foo.py")` -- download foo.py directly from the URL
- `mip.install("http://example.com/x/y/foo.json")` -- download a package, including dependencies, described by foo.json (see below)
- `mip.install("http://example.com/x/y")` -- implicitly http://example.com/x/y/package.json
- `mip.install("github:org/user/path/foo.py")` -- shortcut for github-hosted content (fetches user/path/foo.py from the default branch)
- `mip.install("github:org/user/path/mypackage.json")` -- as above
- `mip.install("github:org/user")` -- this will hopefully be the common case for "self-hosted" packages.
- `mip.install("github:org/user", version="devel")` -- as above, but using the "devel" branch
- `mip.install("foo", mpy=False) -- fetch foo but as .py files (i.e. for debugging).

Major differences in `mip` compared to `upip`:
- Packages installed from the index are bytecode compiled to .mpy by default. The request includes the device's supported version.
- pypi is not supported as a source.
- It uses `urequests` rather than having its own HTTP client. (Most boards that freeze `mip` also freeze urequests anyway)

The package.json looks like

```json
{
  "hashes": [
    ["aioble/server.mpy", "e39dbf64"],
    ...
  ],
  "urls": [
    ["target/path.py", "http://url/foo/bar/path.py"],
    ...
  ],
  "deps": [
    ["name", "version"],
    ...
  ]
  "version": "0.1"
}
```

The "hashes" are for referencing .mpy files directly from micropython-lib. This is used by micropython-lib packages, but shouldn't be used for self-hosted ones. The "urls" allows arbritrary URLs to be fetched, and "deps" allows recursive dependencies, e.g. a self-hosted package might want to provide some urls to their own files, and then depend on some micropython-lib packages (or a package json in another github repo).

Note: deploy.py requires https://github.com/micropython/micropython/pull/9437

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,4 +29,3 @@ Future plans (and new contributor ideas)
 
 * Develop a set of example programs using these libraries.
 * Develop more MicroPython libraries for common tasks.
-* Provide a replacement for the previous `upip` tool.
--- a/README.md
+++ b/README.md
@@ -27,7 +27,6 @@ Other libraries are packages, in which case you'll need to copy the directory in
 Future plans (and new contributor ideas)
 ----------------------------------------
 
-* Provide compiled .mpy distributions.
 * Develop a set of example programs using these libraries.
 * Develop more MicroPython libraries for common tasks.
 * Provide a replacement for the previous `upip` tool.
--- a/micropython/mip/manifest.py
+++ b/micropython/mip/manifest.py
@@ -1,0 +1,5 @@
+metadata(version="0.1.0", description="On-device package installer for network-capable boards")
+
+require("urequests")
+
+module("mip.py")
--- a/micropython/mip/mip.py
+++ b/micropython/mip/mip.py
@@ -1,0 +1,169 @@
+# MicroPython package installer
+# MIT license; Copyright (c) 2022 Jim Mussared
+
+import urequests as requests
+import sys
+
+
+_PACKAGE_INDEX = const("https://micropython.org/pi/v2")
+_CHUNK_SIZE = 128
+
+
+# This implements os.makedirs(os.dirname(path))
+def _ensure_path_exists(path):
+    import os
+
+    split = path.split("/")
+
+    # Handle paths starting with "/".
+    if not split[0]:
+        split.pop(0)
+        split[0] = "/" + split[0]
+
+    prefix = ""
+    for i in range(len(split) - 1):
+        prefix += split[i]
+        try:
+            os.stat(prefix)
+        except:
+            os.mkdir(prefix)
+        prefix += "/"
+
+
+# Copy from src (stream) to dest (function-taking-bytes)
+def _chunk(src, dest):
+    buf = memoryview(bytearray(_CHUNK_SIZE))
+    while True:
+        n = src.readinto(buf)
+        if n == 0:
+            break
+        dest(buf if n == _CHUNK_SIZE else buf[:n])
+
+
+# Check if the specified path exists and matches the hash.
+def _check_exists(path, short_hash):
+    import os
+
+    try:
+        import binascii
+        import hashlib
+
+        with open(path, "rb") as f:
+            hs256 = hashlib.sha256()
+            _chunk(f, hs256.update)
+            existing_hash = str(binascii.hexlify(hs256.digest())[: len(short_hash)], "utf-8")
+            return existing_hash == short_hash
+    except:
+        return False
+
+
+def _rewrite_url(url, branch=None):
+    if not branch:
+        branch = "HEAD"
+    if url.startswith("github:"):
+        url = url[7:].split("/")
+        url = (
+            "https://raw.githubusercontent.com/"
+            + url[0]
+            + "/"
+            + url[1]
+            + "/"
+            + branch
+            + "/"
+            + "/".join(url[2:])
+        )
+    return url
+
+
+def _download_file(url, dest):
+    response = requests.get(url)
+    try:
+        if response.status_code != 200:
+            print("Error", response.status_code, "requesting", url)
+            return False
+
+        print("Copying:", dest)
+        _ensure_path_exists(dest)
+        with open(dest, "wb") as f:
+            _chunk(response.raw, f.write)
+
+        return True
+    finally:
+        response.close()
+
+
+def _install_json(package_json_url, index, target, version, mpy):
+    response = requests.get(_rewrite_url(package_json_url, version))
+    try:
+        if response.status_code != 200:
+            print("Package not found:", package_json_url)
+            return False
+
+        package_json = response.json()
+    finally:
+        response.close()
+    for target_path, short_hash in package_json.get("hashes", ()):
+        fs_target_path = target + "/" + target_path
+        if _check_exists(fs_target_path, short_hash):
+            print("Exists:", fs_target_path)
+        else:
+            file_url = "{}/file/{}/{}".format(index, short_hash[:2], short_hash)
+            if not _download_file(file_url, fs_target_path):
+                print("File not found: {} {}".format(target_path, short_hash))
+                return False
+    for target_path, url in package_json.get("urls", ()):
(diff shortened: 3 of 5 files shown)


jimmo:
I've written a small demo of what writing a "self-hosted" (on GitHub) package will look like, including how to specify dependencies: https://github.com/jimmo/micropython-mlx90640

jimmo:
Another artifact of the build is a top-level `index.json` (which will be https://micropython.org/pi/v2/index.json) which contains the full list of available packages from micropython-lib. upip doesn't consume this, but other tools that want to discover available packages can. What's included in this for now is pretty minimal, but unlike the package metadata I don't imagine this being consumed by low-memory devices so this could have more info added (possibly version history etc).

tools/deploy.py
```diff
@@ -0,0 +1,313 @@
...
+def main():
+    import argparse
+
+    cmd_parser = argparse.ArgumentParser(
+        description="Compile micropython-lib for serving to upip."
+    )
+    cmd_parser.add_argument("--output", required=True, help="output directory")
+    cmd_parser.add_argument("--hash-prefix", default=8, type=int, help="hash prefix length")
+    cmd_parser.add_argument("--mpy-cross", default=None, help="optional path to mpy-cross binary")
+    cmd_parser.add_argument("--micropython", default=None, help="path to micropython repo")
```
dpgeorge, line 302: if you specify the micropython repo, could that be used as the path to mpy-cross?  or will that automatically happen because of line 307 below?

dpgeorge:
I checked the frozen version of this new upip, compared to the old one (compiled with mpy-cross v6 using -O0, ie no optimisations).  This is to see how the new one compares to the old one in terms of flash size.  Note that the new upip relies on urequests, so the total size of the new upip needs to include the size of urequests.

Old upip size:
- esp8266: 6056 bytes frozen
- rp2: 5832 bytes frozen

New upip+urequests combined size:
- esp8266: 5536 bytes frozen
- rp2: 5360 bytes frozen

So the new upip combined with urequests is about 500 bytes smaller (around 10% smaller) than the old upip.  Even bigger savings are made when urequests already exists for other reasons.  In that case upip on its own is just 2464 bytes frozen (on rp2).

Furthermore the dependencies on built-in C modules are reduced with the new upip.  The dependencies are:
- Old upip: sys, gc, os, errno, json, zlib, uctypes, ssl, socket
- New upip: sys, os, binascii (optional), hashlib (optional), and then json, ssl, socket via urequests.

So gc, errno, zlib and uctypes are no longer required by the new upip, compared to the old one.

tools/deploy.py
jimmo: Yes. It's a bit complicated because of the two somewhat independent dependencies on mpy_cross and mpy-cross. :scream: 

What I had imagined here is that you'd run this script something like:

- clone micropython repo
- for each bytecode version we want to support:
  - checkout that version of the micropython repo
  - build mpy-cross
  - run deploy.py with --micropython

jimmo:
I've pushed an updated version that adds two features requested via private feedback:

- Ability to install "source only" versions of packages. Alongside the supported bytecode versions, there's now a "py" version. This can be requested using `upip.install(..., mpy=False)`.
- Added version history to the index.json. This allows you to discover the available versions for each bytecode version.

micropython/upip/upip.py
```diff
@@ -0,0 +1,173 @@
...
+
+        package_json = response.json()
+    finally:
+        response.close()
+    for target_path, short_hash in package_json.get("hashes", ()):
+        fs_target_path = target + "/" + target_path
+        if _check_exists(fs_target_path, short_hash):
+            print("Exists:", fs_target_path)
+        else:
+            file_url = "{}/file/{}/{}".format(index, short_hash[:2], short_hash[2:])
```
dpgeorge, line 110: is it better to use the full hash name as the filename, ie `.format(index, short_hash[:2], short_hash)`?

micropython/upip/upip.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+        else:
+            if not package.endswith(".json"):
+                if not package.endswith("/"):
+                    package += "/"
+                package += "package.json"
+            print("Installing {} to {}".format(package, target))
+            return _install_json(package, index, target, version, mpy)
+    else:
+        return _install_lib(package, index, target, version, mpy)
```
dpgeorge, line 157: if you inline `_install_lib` here it'd be smaller code, and you could combine the (tail) call to `_install_json` with the one just above

micropython/upip/upip.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def install(package, index=_PACKAGE_INDEX, target=None, version=None, mpy=True):
+    if not target:
+        for p in sys.path:
+            if p == "lib" or p.endswith("/lib"):
```
dpgeorge, line 163: i think this can be simplified to just `p.endswith("/lib")`, since all existing ports use one of the following:
- /lib
- /flash/lib
- /sd/lib

micropython/upip/upip.py
```diff
@@ -0,0 +1,173 @@
...
+                target = p
+                break
+        else:
+            print("Unable to find lib dir in sys.path.")
```
dpgeorge, line 167: no other message ends in a full-stop in this file, so i don't think this one should (and saves a byte!)

micropython/upip/upip.py
```diff
@@ -0,0 +1,173 @@
...
+            return
+
+    if _install_package(package, index.rstrip("/"), target, version, mpy):
+        print("Done")
+    else:
+        print("Package may be partially installed.")
```
dpgeorge, line 173: no full-stop

dpgeorge:
What about the ability to do, on the unix port, `micropython -m upip install foo`?  That could be done with the old upip.  Do we need to support that?

jimmo:
Addressed comments above.

> Edit: maybe we can have a separate upip-cli that supports command-line for unix?

Yep, I want to do this like the change we made for `unittest` where if you install `unittest-discover` then you also get `micropython -m unittest`. (Separate PR)

tools/deploy.py
dpgeorge: Sounds good.  I tested using just `--micropython` and it all works.

jimmo:
As discussed, renamed `upip` to `mip`, and `deploy.py` to `build.py`.

dpgeorge:
Thank you, this is awesome!

peterhinch:
Are there any ideas on how to support platforms like Pico that aren't network-enabled?

jimmo:
> Are there any ideas on how to support platforms like Pico that aren't network-enabled?

@peterhinch 
Yep! https://github.com/micropython/micropython/pull/9467

jimmo:
And https://github.com/micropython/micropython/pull/9463 included a significant overhaul of https://docs.micropython.org/en/latest/reference/packages.html (which hasn't gone live yet, but will hopefully update soon).

PR #543, opened 2022-09-30
os.path: Remove ffi os.access usage in exists().

andrewleech:
The current `os.path.exists()` function uses `os.access()` which is only provided in the ffi os module, not in `uos`.

```
anl@STEP: ~ $ micropython
MicroPython v1.18-305-g587339022 on 2022-04-11; linux [GCC 9.4.0] version
>>> import os
>>> os.access
AttributeError: 'module' object has no attribute 'access'
>>>
```

Also, `is_dir` requries the stat module from micropython-lib

### DIFF
--- a/python-stdlib/os-path/manifest.py
+++ b/python-stdlib/os-path/manifest.py
@@ -1,4 +1,4 @@
-metadata(version="0.1.3")
+metadata(version="0.1.4")
 
 # Originally written by Paul Sokolovsky.
 
--- a/python-stdlib/os-path/os/path.py
+++ b/python-stdlib/os-path/os/path.py
@@ -47,7 +47,11 @@ def basename(path):
 
 
 def exists(path):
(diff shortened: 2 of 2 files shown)


dpgeorge:
Looks OK to me.

jimmo:
Rather than pulling in `stat` can we just inline `S_ISDIR(mode)` as `(mode & 0o040000)` ?

andrewleech:
> Rather than pulling in `stat` can we just inline `S_ISDIR(mode)` as `(mode & 0o040000)` ?

While I like the readability of `S_ISDIR`, considering this is a one-liner in a stdlib function it does make sense to reduce the deployment size for users.

jimmo:
Thanks, merged in https://github.com/micropython/micropython-lib/commit/459e13921af980cc52861576e5915dcaf8f9197b.

PR #544, opened 2022-10-06
micropython/mip: Add command-line functionality for the Unix port.

jimmo:
Moves mip.py to mip/__init__.py, so that the optional (added in this commit) mip/__main__.py can exist to support:

`micropython -m mip install [--target,--index,--no-mpy] package@version`

"install" works by forwarding the arguments directly to mip.install.

Also makes mip opt=3 by default because it's going to be frozen-by-default on many boards.

@mattytrentini I think this is something you've expressed an interest in previously?

### DIFF
--- a/micropython/mip-cmdline/manifest.py
+++ b/micropython/mip-cmdline/manifest.py
@@ -1,0 +1,6 @@
+metadata(version="0.1.0", description="Optional support for running `micropython -m mip`")
+
+require("argparse")
+require("mip")
+
+package("mip")
--- a/micropython/mip-cmdline/mip/__main__.py
+++ b/micropython/mip-cmdline/mip/__main__.py
@@ -1,0 +1,46 @@
+# MicroPython package installer command line
+# MIT license; Copyright (c) 2022 Jim Mussared
+
+import argparse
+import sys
+
+
+def do_install():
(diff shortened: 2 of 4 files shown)


jimmo:
Updated to remove "venv" -- I'll make that a standalone package instead.

jimmo:
TODO: Can we think of a better than than `mip-cmdline`.  `mip-main` is probably by best idea so far.

(Compare to unittest-discover as another example of adding cmdline functionality to a package.... See also https://github.com/micropython/micropython-lib/pull/552)

dpgeorge:
> Can we think of a better than than `mip-cmdline`. `mip-main` is probably by best idea so far

I agree `mip-main` is pretty good, but at the same time it could be confused as being "the main thing to install".  Like if you look through all available packages and see `foo-main`, you'd think that was the main entry point and the one you need to install (and you'd be right... if you're on unix).

`foo-cmdline` is more explicit: install this if you want a command-line version.  Could also be something like `foo-ui`.

jimmo:
Merged in a363ac6b216acaf046690ff85191113c95946d0d and 143c2937defc29ff24ed6f34c27ea10bab9135e9 (with the package name `mip-cmdline`).

PR #545, opened 2022-10-07
added timezone support in ntptime

ljnath:
This PR is to add standard timezone support without DST in the micropython ntptime module.

The new addition includes a comprehensive list of timezones. The existing method `settime()` is modified to take an optional string parameter via which user can specify the required timezone.


jimmo:
Thanks @ljnath

Unfortunately the `ntptime` module is frozen into many boards by default and needs to be as minimal as possible. We cannot afford to include a full copy of the set of global timezones. Also, to do timezone support properly, the timezone data needs to be dynamic (e.g. daylight saving time).

Any timezone support will need to be provided by an optional library that the user can install separately to ntptime.

ljnath:
@jimmo , thank you for your feedback. I will close this PR and will another one with an alternate solution.

PR #549, opened 2022-10-13
micropython/drivers: Add "sht3x" sensor driver.

wemos:
Add "sht3x" sensor driver.

### DIFF
--- a/micropython/drivers/sensor/sht3x/manifest.py
+++ b/micropython/drivers/sensor/sht3x/manifest.py
@@ -1,0 +1,1 @@
+module("sht3x.py", opt=3)
--- a/micropython/drivers/sensor/sht3x/sht3x.py
+++ b/micropython/drivers/sensor/sht3x/sht3x.py
@@ -1,0 +1,26 @@
+# SHT3x driver for MicroPython on ESP8266/ESP32
+# MIT license; Copyright (c) 2022 WEMOS.CC
+
+import utime
+
+
+class SHT3X:
+    def __init__(self, i2c, address=0x45):
+        self.bus = i2c
+        self.slv_addr = address
+        self.buf = bytearray(6)
+
+    def measure(self):
+        self.bus.writeto(self.slv_addr, b"\x24\x00")
+        utime.sleep(1)
+        self.buf = self.bus.readfrom(self.slv_addr, 6)
+
+    def humidity(self):
+        humidity_raw = self.buf[3] << 8 | self.buf[4]
+        humidity = 100.0 * float(humidity_raw) / 65535.0
+        return humidity
+
+    def temperature(self):
+        temperature_raw = self.buf[0] << 8 | self.buf[1]
+        temperature = (175.0 * float(temperature_raw) / 65535.0) - 45
+        return temperature
--- a/micropython/drivers/sensor/sc7a20/manifest.py
+++ b/micropython/drivers/sensor/sc7a20/manifest.py
@@ -1,0 +1,1 @@
+module("sc7a20.py", opt=3)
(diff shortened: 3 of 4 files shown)


micropython/drivers/sensor/sht3x/manifest.py
```diff
@@ -0,0 +1 @@
+module("sht3x.py", opt=3)
```
jimmo, line 1: Please add a `metadata()` with description and version.

micropython/drivers/sensor/sht3x/sht3x.py
```diff
@@ -0,0 +1,26 @@
+# SHT3x driver for MicroPython on ESP8266/ESP32
+# MIT license; Copyright (c) 2022 WEMOS.CC
+
+import utime
```
jimmo, line 4: Prefer to use `import time`

micropython/drivers/sensor/sht3x/sht3x.py
```diff
@@ -0,0 +1,26 @@
+# SHT3x driver for MicroPython on ESP8266/ESP32
```
jimmo, line 1: Doesn't look like this is specific to ESP as it uses the common i2c API.

Can you also add that it's a temp/humidity sensor, and made by Sensirion.

micropython/drivers/sensor/sht3x/sht3x.py
```diff
@@ -0,0 +1,26 @@
...
+# MIT license; Copyright (c) 2022 WEMOS.CC
+
+import utime
+
+
+class SHT3X:
+    def __init__(self, i2c, address=0x45):
+        self.bus = i2c
+        self.slv_addr = address
```
jimmo, line 10: `_i2c`, `_addr`, and `_buf`

micropython/drivers/sensor/sht3x/sht3x.py
```diff
@@ -0,0 +1,26 @@
...
+        self.buf = bytearray(6)
+
+    def measure(self):
+        self.bus.writeto(self.slv_addr, b"\x24\x00")
+        utime.sleep(1)
+        self.buf = self.bus.readfrom(self.slv_addr, 6)
```
jimmo, line 16: This should be readfrom_into to avoid re-allocating `buf`.

micropython/drivers/sensor/sht3x/sht3x.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def humidity(self):
```
jimmo, line 18: I wonder about adding a non-floating point version of these functions (to avoid float allocations), i.e.return an integer in 1/10ths of a degree / percent. (The point being that if you remove this float allocation and with the readfrom_into above, then the whole driver is non-allocating).

micropython/drivers/sensor/sht3x/sht3x.py
```diff
@@ -0,0 +1,26 @@
...
+        humidity_raw = self.buf[3] << 8 | self.buf[4]
+        humidity = 100.0 * float(humidity_raw) / 65535.0
+        return humidity
+
+    def temperature(self):
```
jimmo, line 23: Assuming this is in degrees Celcius? (Maybe add a comment so someone can easily check this by looking at the code, or perhaps rename the variable to temperature_c)

micropython/drivers/sensor/sht3x/sht3x.py
jimmo: Or perhaps just have only the versions that return tenths and the calling code can simply convert to float by doing `s.humidity() / 10.0' if necessary.

jimmo (commented):
Thanks @wemos !

jimmo:
@wemos Thanks for implementing the `SC7A20` driver too, but could you please add this in a different PR?

We're keen to merge this PR, but please address the review comments first.

wemos:
ok

PR #551, opened 2022-10-14
venv: Add a command-line package for creating virtual environments.

jimmo:
Works like "python -m venv path" and creates a rudimentary virtual environment for the Unix port:
 - sets MICROPYPATH
 - copies the micropython binary to venv/bin/micropython which is in $PATH
 - installs mip & mip-cmdline in the venv

Using the venv is the same as for CPython -- source the activate script to enter, and call the deactivate function to leave.


jimmo:
(Note this depends on mip-cmdline in #544)

jimmo:
Merged in c26d77b52e3bfb82729787a5c30cca27a8cf9f99.

PR #554, opened 2022-10-17
net/ntptime/ntptime.py: Allow overriding default NTP timeout.

iabdalkader:
* The default 1 second timeout is sometimes not enough. Allow overriding the default socket timeout by passing a timeout arg.

### DIFF
--- a/micropython/net/ntptime/ntptime.py
+++ b/micropython/net/ntptime/ntptime.py
@@ -11,6 +11,8 @@ except:
 
 # The NTP host can be configured at runtime by doing: ntptime.host = 'myhost.org'
 host = "pool.ntp.org"
+# The NTP socket timeout can be configured at runtime by doing: ntptime.timeout = 1.0
+timeout = 1.0
 
 
 def time():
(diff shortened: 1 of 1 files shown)


dpgeorge:
There's also `settime()` which calls `time()`, and probably should also allow passing in the timeout.

But then maybe a better way to do it is have a global variable to set the timeout.  This will match how you select the NTP host.

Eg:
```py
host = "pool.ntp.org"
timeout = 1

...

    s.settimeout(timeout)
```

iabdalkader:
I think the `host` should be private and the default ex: `_HOST` and a default timeout `_TIMEOUT=const(1.0)` , and both functions accept kwordargs `settime(timeout=_DEFAULT_TIMEOUT, host=_DEFAULT_HOST)`

iabdalkader:
@dpgeorge How about now ?

dpgeorge:
But this changes the existing public API, where you change `ntptime.host` to configure your desired server.  We had a lot of PRs to change this behaviour already and that's why there's a comment there describing how to use it.

The idea behind having a global variable is that you can change it once -- configure it -- then you don't need to pass it in to `time()` every time you call it.  IMO setting it once to a set value is a more common use case that passing in potentially different hosts at different locations where this is called.

(Note that it's not such bad practice having a global here, because the `ntptime` itself is very small and self contained, you can think of `ntptime` as being a singleton class, and `ntptime.host` as being a property that you get/set.)

iabdalkader:
Okay, I see your point, it's not important as long as we can override it somehow. Added a `timeout` globally just like `host`

dpgeorge:
Thanks for updating.  Merged in 900dd1c61b42c709417e303b706b5e5055214740 (and I change the default from "1.0" to "1" so it doesn't use a float if it doesn't need to).

PR #557, opened 2022-10-25
drivers/sensor/dht: Change the sequence for importing dht_readinto.

robert-hh:
Check the machine module first, then try the port specific places.

### DIFF
--- a/micropython/drivers/sensor/dht/dht.py
+++ b/micropython/drivers/sensor/dht/dht.py
@@ -2,8 +2,11 @@
 # MIT license; Copyright (c) 2016 Damien P. George
 
 import sys
+import machine
 
-if sys.platform.startswith("esp"):
+if hasattr(machine, "dht_readinto"):
+    from machine import dht_readinto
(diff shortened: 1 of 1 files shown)


dpgeorge:
I'm not sure what this achieves?  One thing it does do is make the code bigger with the additional error message :smile: 

Probably best to just move all `dht_readinto` methods on the ports into `machine`, and then make this module just do `from machine import dht_readinto` unconditionally.

robert-hh:
The intention is legacy, such that the 'new' dht.py works with previous versions of the firmwarem, at least for a while.

dpgeorge:
Oh, I see!  Yes that makes sense then.

Probably good to add a comment then, why the older imports are there.

Also, except for the esp case, the port-specific cases can actually be combined to:
```py
dht_readinto = __import__(sys.platform).dht_readinto
```

dpgeorge:
Squashed and merged in 0e25b109c23fe30a3a0f075c0edf2c2564bc5046

PR #558, opened 2022-10-26
usb: Add high-level USB device support packages

projectgus:
High-level support packages for creating USB device functionality in Python code.

These modules rely on the machine.USBDevice() object newly merged for MicroPython 1.23 (https://github.com/micropython/micropython/pull/9497).

**IMPORTANT**: The latest version of this PR requires a low-level change in MicroPython's USBDevice that was merged in https://github.com/micropython/micropython/pull/14253 - nightly builds from before 18 April 2024 do not have this change, and will fail in unexpected ways!!

## How to test/use this PR

This PR requires a nightly build of Micropython 1.23-preview or later, on a port with machine.USBDevice support (currently samd or rp2).

Can either edit manifest.py to freeze usb-device packages into the firmware, or use the following command to install packages directly from this PR:

```
$ mpremote connect /dev/ttyUSB0 mip install --index https://projectgus.github.io/micropython-lib/mip/feature/usbd_python usb-device-NAME
```

See the README in this PR for a list of package names.

## Example Code

See the `usb/examples/device` subdirectory for some simple example programs.

These aren't installed with the package but can be run as (for example) `mpremote run mouse_example.py`.

(Note: because any USB serial connection will close on reenumeration, printed output from this point will only be visible if outputting stdout to traditional UART, semihosting, etc.)

## Debugging Tips

See the micropython PR https://github.com/micropython/micropython/pull/9497

## TODOs

This PR is nearly complete, aside from code review comments I have a few things left to finish:

* [ ] Complete README with some class reference information.
* [x] Re-test on Windows.
* [x] Test all examples on macOS (help wanted as I don't have a Mac on hand).

## Future work (not this PR)

* Add an async USB interface helper. As USB is host-initiated and mostly uses callbacks, there isn't a whole lot more work needed here - however some classes would benefit from async variants.
* See if we can provide a host helper program for generating HID Report Descriptors (or link to one that already exists). I tried implementing this in MicroPython so it builds the Report Descriptor on the device at runtime, but this takes up too much space.

### DIFF
--- a/micropython/usb/README.md
+++ b/micropython/usb/README.md
@@ -1,0 +1,136 @@
+# USB Drivers
+
+These packages allow implementing USB functionality on a MicroPython system
+using pure Python code.
+
+Currently only USB device is implemented, not USB host.
+
+## USB Device support
+
+### Support
+
+USB Device support depends on the low-level
+[machine.USBDevice](https://docs.micropython.org/en/latest/library/machine.USBDevice.html)
+class. This class is new and not supported on all ports, so please check the
+documentation for your MicroPython version. It is possible to implement a USB
+device using only the low-level USBDevice class. However, the packages here are
+higher level and easier to use.
+
+For more information about how to install packages, or "freeze" them into a
+firmware image, consult the [MicroPython documentation on "Package
+management"](https://docs.micropython.org/en/latest/reference/packages.html).
+
+### Examples
+
+The [examples/device](examples/device) directory in this repo has a range of
+examples. After installing necessary packages, you can download an example and
+run it with `mpremote run EXAMPLE_FILENAME.py` ([mpremote
+docs](https://docs.micropython.org/en/latest/reference/mpremote.html#mpremote-command-run)).
+
+#### Unexpected serial disconnects
+
+If you normally connect to your MicroPython device over a USB serial port ("USB
+CDC"), then running a USB example will disconnect mpremote when the new USB
+device configuration activates and the serial port has to temporarily
+disconnect. It is likely that mpremote will print an error. The example should
+still start running, if necessary then you can reconnect with mpremote and type
+Ctrl-B to restore the MicroPython REPL and/or Ctrl-C to stop the running
+example.
+
+If you use `mpremote run` again while a different USB device configuration is
+already active, then the USB serial port may disconnect immediately before the
+example runs. This is because mpremote has to soft-reset MicroPython, and when
+the existing USB device is reset then the entire USB port needs to reset. If
+this happens, run the same `mpremote run` command again.
+
+We plan to add features to `mpremote` so that this limitation is less
+disruptive. Other tools that communicate with MicroPython over the serial port
+will encounter similar issues when runtime USB is in use.
+
+### Initialising runtime USB
+
+The overall pattern for enabling USB devices at runtime is:
+
+1. Instantiate the Interface objects for your desired USB device.
+2. Call `usb.device.get()` to get the singleton object for the high-level USB device.
+3. Call `init(...)` to pass the desired interfaces as arguments, plus any custom
+   keyword arguments to configure the overall device.
+
+An example, similar to [mouse_example.py](examples/device/mouse_example.py):
+
+```py
+    m = usb.device.mouse.MouseInterface()
+    usb.device.get().init(m, builtin_driver=True)
+```
+
+Setting `builtin_driver=True` means that any built-in USB serial port will still
+be available. Otherwise, you may permanently lose access to MicroPython until
+the next time the device resets.
+
+See [Unexpected serial disconnects](#Unexpected-serial-disconnects), above, for
+an explanation of possible errors or disconnects when the runtime USB device
+initialises.
+
+Placing the call to `usb.device.get().init()` into the `boot.py` of the
+MicroPython file system allows the runtime USB device to initialise immediately
+on boot, before any built-in USB. This is a feature (not a bug) and allows you
+full control over the USB device, for example to only enable USB HID and prevent
+REPL access to the system.
+
+However, note that calling this function on boot without `builtin_driver=True`
+will make the MicroPython USB serial interface permanently inaccessible until
+you "safe mode boot" (on supported boards) or completely erase the flash of your
+device.
+
+### Package usb-device
+
+This base package contains the common implementation components for the other
+packages, and can be used to implement new and different USB interface support.
+All of the other `usb-device-<name>` packages depend on this package, and it
+will be automatically installed as needed.
+
+Specicially, this package provides the `usb.device.get()` function for accessing
+the Device singleton object, and the `usb.device.core` module which contains the
+low-level classes and utility functions for implementing new USB interface
+drivers in Python. The best examples of how to use the core classes is the
+source code of the other USB device packages.
+
+### Package usb-device-keyboard
+
+This package provides the `usb.device.keyboard` module. See
+[keyboard_example.py](examples/device/keyboard_example.py) for an example
+program.
+
+### Package usb-device-mouse
+
+This package provides the `usb.device.mouse` module. See
+[mouse_example.py](examples/device/mouse_example.py) for an example program.
+
+### Package usb-device-hid
+
+This package provides the `usb.device.hid` module. USB HID (Human Interface
+Device) class allows creating a wide variety of device types. The most common
+are mouse and keyboard, which have their own packages in micropython-lib.
+However, using the usb-device-hid package directly allows creation of any kind
+of HID device.
+
+See [hid_custom_keypad_example.py](examples/device/hid_custom_keypad_example.py)
+for an example of a Keypad HID device with a custom HID descriptor.
+
+### Package usb-device-cdc
+
+This package provides the `usb.device.cdc` module. USB CDC (Communications
+Device Class) is most commonly used for virtual serial port USB interfaces, and
+that is what is supported here.
+
+The example [cdc_repl_example.py](examples/device/cdc_repl_example.py)
+demonstrates how to add a second USB serial interface and duplicate the
+MicroPython REPL between the two.
+
+### Package usb-device-midi
+
+This package provides the `usb.device.midi` module. This allows implementing
+USB MIDI devices in MicroPython.
+
+The example [midi_example.py](examples/device/midi_example.py) demonstrates how
+to create a simple MIDI device to send MIDI data to and from the USB host.
--- a/micropython/usb/examples/device/cdc_repl_example.py
+++ b/micropython/usb/examples/device/cdc_repl_example.py
@@ -1,0 +1,47 @@
+# MicroPython USB CDC REPL example
+#
+# Example demonstrating how to use os.dupterm() to provide the
+# MicroPython REPL on a dynamic CDCInterface() serial port.
+#
+# To run this example:
+#
+# 1. Make sure `usb-device-cdc` is installed via: mpremote mip install usb-device-cdc
+#
+# 2. Run the example via: mpremote run cdc_repl_example.py
+#
+# 3. mpremote will exit with an error after the previous step, because when the
+#    example runs the existing USB device disconnects and then re-enumerates with
+#    the second serial port. If you check (for example by running mpremote connect
+#    list) then you should now see two USB serial devices.
+#
+# 4. Connect to one of the new ports: mpremote connect PORTNAME
+#
+#    It may be necessary to type Ctrl-B to exit the raw REPL mode and resume the
+#    interactive REPL after mpremote connects.
+#
+# MIT license; Copyright (c) 2023-2024 Angus Gratton
+import os
+import time
+import usb.device
+from usb.device.cdc import CDCInterface
+
+cdc = CDCInterface()
+cdc.init(timeout=0)  # zero timeout makes this non-blocking, suitable for os.dupterm()
+
+# pass builtin_driver=True so that we get the built-in USB-CDC alongside,
+# if it's available.
+usb.device.get().init(cdc, builtin_driver=True)
+
+print("Waiting for USB host to configure the interface...")
+
+# wait for host enumerate as a CDC device...
+while not cdc.is_open():
+    time.sleep_ms(100)
+
+# Note: This example doesn't wait for the host to access the new CDC port,
+# which could be done by polling cdc.dtr, as this will block the REPL
+# from resuming while this code is still executing.
+
+print("CDC port enumerated, duplicating REPL...")
+
+old_term = os.dupterm(cdc)
--- a/micropython/usb/examples/device/hid_custom_keypad_example.py
+++ b/micropython/usb/examples/device/hid_custom_keypad_example.py
@@ -1,0 +1,144 @@
+# MicroPython USB HID custom Keypad example
+#
+# This example demonstrates creating a custom HID device with its own
+# HID descriptor, in this case for a USB number keypad.
+#
+# For higher level examples that require less code to use, see mouse_example.py
+# and keyboard_example.py
+#
+# To run this example:
+#
+# 1. Make sure `usb-device-hid` is installed via: mpremote mip install usb-device-hid
+#
+# 2. Run the example via: mpremote run hid_custom_keypad_example.py
+#
+# 3. mpremote will exit with an error after the previous step, because when the
+#    example runs the existing USB device disconnects and then re-enumerates with
+#    the custom HID interface present. At this point, the example is running.
+#
+# 4. To see output from the example, re-connect: mpremote connect PORTNAME
+#
+# MIT license; Copyright (c) 2023 Dave Wickham, 2023-2024 Angus Gratton
+from micropython import const
+import time
+import usb.device
+from usb.device.hid import HIDInterface
+
+_INTERFACE_PROTOCOL_KEYBOARD = const(0x01)
+
+
+def keypad_example():
+    k = KeypadInterface()
+
+    usb.device.get().init(k, builtin_driver=True)
+
+    while not k.is_open():
+        time.sleep_ms(100)
+
+    while True:
+        time.sleep(2)
+        print("Press NumLock...")
+        k.send_key("<NumLock>")
+        time.sleep_ms(100)
+        k.send_key()
+        time.sleep(1)
+        # continue
+        print("Press ...")
+        for _ in range(3):
+            time.sleep(0.1)
+            k.send_key(".")
+            time.sleep(0.1)
+            k.send_key()
+        print("Starting again...")
+
+
+class KeypadInterface(HIDInterface):
+    # Very basic synchronous USB keypad HID interface
+
+    def __init__(self):
+        super().__init__(
+            _KEYPAD_REPORT_DESC,
+            set_report_buf=bytearray(1),
+            protocol=_INTERFACE_PROTOCOL_KEYBOARD,
+            interface_str="MicroPython Keypad",
+        )
+        self.numlock = False
+
+    def on_set_report(self, report_data, _report_id, _report_type):
+        report = report_data[0]
+        b = bool(report & 1)
+        if b != self.numlock:
+            print("Numlock: ", b)
+            self.numlock = b
+
+    def send_key(self, key=None):
+        if key is None:
+            self.send_report(b"\x00")
+        else:
+            self.send_report(_key_to_id(key).to_bytes(1, "big"))
+
+
+# See HID Usages and Descriptions 1.4, section 10 Keyboard/Keypad Page (0x07)
+#
+# This keypad example has a contiguous series of keys (KEYPAD_KEY_IDS) starting
+# from the NumLock/Clear keypad key (0x53), but you can send any Key IDs from
+# the table in the HID Usages specification.
+_KEYPAD_KEY_OFFS = const(0x53)
+
+_KEYPAD_KEY_IDS = [
+    "<NumLock>",
+    "/",
+    "*",
+    "-",
+    "+",
+    "<Enter>",
+    "1",
+    "2",
+    "3",
+    "4",
+    "5",
+    "6",
+    "7",
+    "8",
+    "9",
+    "0",
+    ".",
+]
+
+
+def _key_to_id(key):
+    # This is a little slower than making a dict for lookup, but uses
+    # less memory and O(n) can be fast enough when n is small.
+    return _KEYPAD_KEY_IDS.index(key) + _KEYPAD_KEY_OFFS
+
+
+# HID Report descriptor for a numeric keypad
+#
+# fmt: off
+_KEYPAD_REPORT_DESC = (
+    b'\x05\x01'  # Usage Page (Generic Desktop)
+        b'\x09\x07'  # Usage (Keypad)
+    b'\xA1\x01'  # Collection (Application)
+        b'\x05\x07'  # Usage Page (Keypad)
+            b'\x19\x00'  # Usage Minimum (0)
+            b'\x29\xFF'  # Usage Maximum (ff)
+            b'\x15\x00'  # Logical Minimum (0)
+            b'\x25\xFF'  # Logical Maximum (ff)
+            b'\x95\x01'  # Report Count (1),
+            b'\x75\x08'  # Report Size (8),
+            b'\x81\x00'  # Input (Data, Array, Absolute)
+        b'\x05\x08'  # Usage page (LEDs)
+            b'\x19\x01'  # Usage Minimum (1)
+            b'\x29\x01'  # Usage Maximum (1),
+            b'\x95\x01'  # Report Count (1),
+            b'\x75\x01'  # Report Size (1),
+            b'\x91\x02'  # Output (Data, Variable, Absolute)
+            b'\x95\x01'  # Report Count (1),
+            b'\x75\x07'  # Report Size (7),
+            b'\x91\x01'  # Output (Constant) - padding bits
+    b'\xC0'  # End Collection
+)
+# fmt: on
+
+
+keypad_example()
--- a/micropython/usb/examples/device/midi_example.py
+++ b/micropython/usb/examples/device/midi_example.py
@@ -1,0 +1,78 @@
+# MicroPython USB MIDI example
+#
+# This example demonstrates creating a custom MIDI device.
+#
+# To run this example:
+#
+# 1. Make sure `usb-device-midi` is installed via: mpremote mip install usb-device-midi
+#
+# 2. Run the example via: mpremote run midi_example.py
+#
+# 3. mpremote will exit with an error after the previous step, because when the
+#    example runs the existing USB device disconnects and then re-enumerates with
+#    the MIDI interface present. At this point, the example is running.
+#
+# 4. To see output from the example, re-connect: mpremote connect PORTNAME
+#
+#
+# MIT license; Copyright (c) 2023-2024 Angus Gratton
+import usb.device
+from usb.device.midi import MIDIInterface
+import time
+
+
+class MIDIExample(MIDIInterface):
+    # Very simple example event handler functions, showing how to receive note
+    # and control change messages sent from the host to the device.
+    #
+    # If you need to send MIDI data to the host, then it's fine to instantiate
+    # MIDIInterface class directly.
+
+    def on_open(self):
+        super().on_open()
+        print("Device opened by host")
+
+    def on_note_on(self, channel, pitch, vel):
+        print(f"RX Note On channel {channel} pitch {pitch} velocity {vel}")
+
+    def on_note_off(self, channel, pitch, vel):
+        print(f"RX Note Off channel {channel} pitch {pitch} velocity {vel}")
+
+    def on_control_change(self, channel, controller, value):
+        print(f"RX Control channel {channel} controller {controller} value {value}")
+
+
+m = MIDIExample()
+# Remove builtin_driver=True if you don't want the MicroPython serial REPL available.
+usb.device.get().init(m, builtin_driver=True)
+
+print("Waiting for USB host to configure the interface...")
+
+while not m.is_open():
+    time.sleep_ms(100)
+
+print("Starting MIDI loop...")
+
+# TX constants
+CHANNEL = 0
+PITCH = 60
+CONTROLLER = 64
+
+control_val = 0
+
+while m.is_open():
+    time.sleep(1)
+    print(f"TX Note On channel {CHANNEL} pitch {PITCH}")
+    m.note_on(CHANNEL, PITCH)  # Velocity is an optional third argument
+    time.sleep(0.5)
+    print(f"TX Note Off channel {CHANNEL} pitch {PITCH}")
+    m.note_off(CHANNEL, PITCH)
+    time.sleep(1)
+    print(f"TX Control channel {CHANNEL} controller {CONTROLLER} value {control_val}")
+    m.control_change(CHANNEL, CONTROLLER, control_val)
+    control_val += 1
+    if control_val == 0x7F:
+        control_val = 0
+    time.sleep(1)
+
+print("USB host has reset device, example done.")
--- a/micropython/usb/usb-device-mouse/usb/device/mouse.py
+++ b/micropython/usb/usb-device-mouse/usb/device/mouse.py
@@ -1,0 +1,100 @@
+# MicroPython USB Mouse module
+#
+# MIT license; Copyright (c) 2023-2024 Angus Gratton
+from micropython import const
+import struct
+import machine
+
+from usb.device.hid import HIDInterface
+
+_INTERFACE_PROTOCOL_MOUSE = const(0x02)
+
+
+class MouseInterface(HIDInterface):
+    # A basic three button USB mouse HID interface
+    def __init__(self, interface_str="MicroPython Mouse"):
+        super().__init__(
+            _MOUSE_REPORT_DESC,
+            protocol=_INTERFACE_PROTOCOL_MOUSE,
+            interface_str=interface_str,
+        )
+        self._l = False  # Left button
+        self._m = False  # Middle button
+        self._r = False  # Right button
+        self._buf = bytearray(3)
+
+    def send_report(self, dx=0, dy=0):
+        b = 0
+        if self._l:
+            b |= 1 << 0
+        if self._r:
+            b |= 1 << 1
+        if self._m:
+            b |= 1 << 2
+
+        # Wait for any pending report to be sent to the host
+        # before updating contents of _buf.
+        #
+        # This loop can be removed if you don't care about possibly missing a
+        # transient report, the final report buffer contents will always be the
+        # last one sent to the host (it just might lose one of the ones in the
+        # middle).
+        while self.busy():
+            machine.idle()
+
+        struct.pack_into("Bbb", self._buf, 0, b, dx, dy)
+
+        return super().send_report(self._buf)
+
+    def click_left(self, down=True):
+        self._l = down
+        return self.send_report()
+
+    def click_middle(self, down=True):
+        self._m = down
+        return self.send_report()
+
+    def click_right(self, down=True):
+        self._r = down
+        return self.send_report()
+
+    def move_by(self, dx, dy):
+        if not -127 <= dx <= 127:
+            raise ValueError("dx")
+        if not -127 <= dy <= 127:
+            raise ValueError("dy")
+        return self.send_report(dx, dy)
+
+
+# Basic 3-button mouse HID Report Descriptor.
+# This is based on Appendix E.10 of the HID v1.11 document.
+# fmt: off
+_MOUSE_REPORT_DESC = (
+    b'\x05\x01'  # Usage Page (Generic Desktop)
+        b'\x09\x02'  # Usage (Mouse)
+    b'\xA1\x01'  # Collection (Application)
+        b'\x09\x01'  # Usage (Pointer)
+        b'\xA1\x00'  # Collection (Physical)
+            b'\x05\x09'  # Usage Page (Buttons)
+                b'\x19\x01'  # Usage Minimum (01),
+                b'\x29\x03'  # Usage Maximun (03),
+                b'\x15\x00'  # Logical Minimum (0),
+                b'\x25\x01'  # Logical Maximum (1),
+                b'\x95\x03'  # Report Count (3),
+                b'\x75\x01'  # Report Size (1),
+                b'\x81\x02'  # Input (Data, Variable, Absolute), ;3 button bits
+                b'\x95\x01'  # Report Count (1),
+                b'\x75\x05'  # Report Size (5),
+                b'\x81\x01'  # Input (Constant), ;5 bit padding
+            b'\x05\x01'  # Usage Page (Generic Desktop),
+                b'\x09\x30'  # Usage (X),
+                b'\x09\x31'  # Usage (Y),
+                b'\x15\x81'  # Logical Minimum (-127),
+                b'\x25\x7F'  # Logical Maximum (127),
+                b'\x75\x08'  # Report Size (8),
+                b'\x95\x02'  # Report Count (2),
+                b'\x81\x06'  # Input (Data, Variable, Relative), ;2 position bytes (X & Y)
+        b'\xC0'  # End Collection
+    b'\xC0'  # End Collection
+)
+# fmt: on
--- a/micropython/usb/usb-device/tests/test_core_buffer.py
+++ b/micropython/usb/usb-device/tests/test_core_buffer.py
@@ -1,0 +1,97 @@
+# Tests for the Buffer class included in usb.device.core
+#
+# The easiest way to run this is using unix port. From the parent usb-device
+# directory, run as:
+#
+# $ micropython -m tests.test_core_buffer
+#
+import micropython
+from usb.device import core
+
+if not hasattr(core.machine, "disable_irq"):
+    # Inject a fake machine module which allows testing on the unix port, and as
+    # a bonus have tests fail if the buffer allocates inside a critical section.
+    class FakeMachine:
+        def disable_irq(self):
+            return micropython.heap_lock()
+
+        def enable_irq(self, was_locked):
+            if not was_locked:
+                micropython.heap_unlock()
+
+    core.machine = FakeMachine()
+
+
+b = core.Buffer(16)
+
+# Check buffer is empty
+assert b.readable() == 0
+assert b.writable() == 16
+
+# Single write then read
+w = b.pend_write()
+assert len(w) == 16
+w[:8] = b"12345678"
+b.finish_write(8)
+
+# Empty again
+assert b.readable() == 8
+assert b.writable() == 8
+
+r = b.pend_read()
+assert len(r) == 8
+assert r == b"12345678"
+b.finish_read(8)
+
+# Empty buffer again
+assert b.readable() == 0
+assert b.writable() == 16
+
+# Single write then split reads
+b.write(b"abcdefghijklmnop")
+assert b.writable() == 0  # full buffer
+
+r = b.pend_read()
+assert r == b"abcdefghijklmnop"
+b.finish_read(2)
+
+r = b.pend_read()
+assert r == b"cdefghijklmnop"
+b.finish_read(3)
+
+# write to end of buffer
+b.write(b"AB")
+
+r = b.pend_read()
+assert r == b"fghijklmnopAB"
+
+# write while a read is pending
+b.write(b"XY")
+
+# previous pend_read() memoryview should be the same
+assert r == b"fghijklmnopAB"
+
+b.finish_read(4)
+r = b.pend_read()
+assert r == b"jklmnopABXY"  # four bytes consumed from head, one new byte at tail
+
+# read while a write is pending
+w = b.pend_write()
+assert len(w) == 5
+r = b.pend_read()
+assert len(r) == 11
+b.finish_read(3)
+w[:2] = b"12"
+b.finish_write(2)
+
+# Expected final state of buffer
+tmp = bytearray(b.readable())
+assert b.readinto(tmp) == len(tmp)
+assert tmp == b"mnopABXY12"
+
+# Now buffer is empty again
+assert b.readable() == 0
+assert b.readinto(tmp) == 0
+assert b.writable() == 16
+
+print("All Buffer tests passed")
--- a/micropython/usb/usb-device/usb/device/core.py
+++ b/micropython/usb/usb-device/usb/device/core.py
@@ -1,0 +1,851 @@
+# MicroPython Library runtime USB device implementation
+#
+# These contain the classes and utilities that are needed to
+# implement a USB device, not any complete USB drivers.
+#
+# MIT license; Copyright (c) 2022-2024 Angus Gratton
+from micropython import const
+import machine
+import struct
+
+_EP_IN_FLAG = const(1 << 7)
+
+# USB descriptor types
+_STD_DESC_DEV_TYPE = const(0x1)
+_STD_DESC_CONFIG_TYPE = const(0x2)
+_STD_DESC_STRING_TYPE = const(0x3)
+_STD_DESC_INTERFACE_TYPE = const(0x4)
+_STD_DESC_ENDPOINT_TYPE = const(0x5)
+_STD_DESC_INTERFACE_ASSOC = const(0xB)
+
+_ITF_ASSOCIATION_DESC_TYPE = const(0xB)  # Interface Association descriptor
+
+# Standard USB descriptor lengths
+_STD_DESC_CONFIG_LEN = const(9)
+_STD_DESC_ENDPOINT_LEN = const(7)
+_STD_DESC_INTERFACE_LEN = const(9)
+
+_DESC_OFFSET_LEN = const(0)
+_DESC_OFFSET_TYPE = const(1)
+
+_DESC_OFFSET_INTERFACE_NUM = const(2)  # for _STD_DESC_INTERFACE_TYPE
+_DESC_OFFSET_ENDPOINT_NUM = const(2)  # for _STD_DESC_ENDPOINT_TYPE
+
+# Standard control request bmRequest fields, can extract by calling split_bmRequestType()
+_REQ_RECIPIENT_DEVICE = const(0x0)
+_REQ_RECIPIENT_INTERFACE = const(0x1)
+_REQ_RECIPIENT_ENDPOINT = const(0x2)
+_REQ_RECIPIENT_OTHER = const(0x3)
+
+# Offsets into the standard configuration descriptor, to fixup
+_OFFS_CONFIG_iConfiguration = const(6)
+
+_INTERFACE_CLASS_VENDOR = const(0xFF)
+_INTERFACE_SUBCLASS_NONE = const(0x00)
+_PROTOCOL_NONE = const(0x00)
+
+# These need to match the constants in tusb_config.h
+_USB_STR_MANUF = const(0x01)
+_USB_STR_PRODUCT = const(0x02)
+_USB_STR_SERIAL = const(0x03)
+
+# Error constant to match mperrno.h
+_MP_EINVAL = const(22)
+
+_dev = None  # Singleton _Device instance
+
+
+def get():
+    # Getter to access the singleton instance of the
+    # MicroPython _Device object
+    #
+    # (note this isn't the low-level machine.USBDevice object, the low-level object is
+    # get()._usbd.)
+    global _dev
+    if not _dev:
+        _dev = _Device()
+    return _dev
+
+
+class _Device:
+    # Class that implements the Python parts of the MicroPython USBDevice.
+    #
+    # This class should only be instantiated by the singleton getter
+    # function usb.device.get(), never directly.
+    def __init__(self):
+        self._itfs = {}  # Mapping from interface number to interface object, set by init()
+        self._eps = {}  # Mapping from endpoint address to interface object, set by _open_cb()
+        self._ep_cbs = {}  # Mapping from endpoint address to Optional[xfer callback]
+        self._usbd = machine.USBDevice()  # low-level API
+
+    def init(self, *itfs, **kwargs):
+        # Helper function to configure the USB device and activate it in a single call
+        self.active(False)
+        self.config(*itfs, **kwargs)
+        self.active(True)
+
+    def config(  # noqa: PLR0913
+        self,
+        *itfs,
+        builtin_driver=False,
+        manufacturer_str=None,
+        product_str=None,
+        serial_str=None,
+        configuration_str=None,
+        id_vendor=None,
+        id_product=None,
+        bcd_device=None,
+        device_class=0,
+        device_subclass=0,
+        device_protocol=0,
+        config_str=None,
+        max_power_ma=None,
+    ):
+        # Configure the USB device with a set of interfaces, and optionally reconfiguring the
+        # device and configuration descriptor fields
+
+        _usbd = self._usbd
+
+        if self.active():
+            raise OSError(_MP_EINVAL)  # Must set active(False) first
+
+        # Convenience: Allow builtin_driver to be True, False or one of
+        # the machine.USBDevice.BUILTIN_ constants
+        if isinstance(builtin_driver, bool):
+            builtin_driver = _usbd.BUILTIN_DEFAULT if builtin_driver else _usbd.BUILTIN_NONE
+        _usbd.builtin_driver = builtin_driver
+
+        # Putting None for any strings that should fall back to the "built-in" value
+        # Indexes in this list depends on _USB_STR_MANUF, _USB_STR_PRODUCT, _USB_STR_SERIAL
+        strs = [None, manufacturer_str, product_str, serial_str]
+
+        # Build the device descriptor
+        FMT = "<BBHBBBBHHHBBBB"
+        # read the static descriptor fields
+        f = struct.unpack(FMT, builtin_driver.desc_dev)
+
+        def maybe_set(value, idx):
+            # Override a numeric descriptor value or keep builtin value f[idx] if 'value' is None
+            if value is not None:
+                return value
+            return f[idx]
+
+        # Either copy each descriptor field directly from the builtin device descriptor, or 'maybe'
+        # set it to the custom value from the object
+        desc_dev = struct.pack(
+            FMT,
+            f[0],  # bLength
+            f[1],  # bDescriptorType
+            f[2],  # bcdUSB
+            device_class,  # bDeviceClass
+            device_subclass,  # bDeviceSubClass
+            device_protocol,  # bDeviceProtocol
+            f[6],  # bMaxPacketSize0, TODO: allow overriding this value?
+            maybe_set(id_vendor, 7),  # idVendor
+            maybe_set(id_product, 8),  # idProduct
+            maybe_set(bcd_device, 9),  # bcdDevice
+            _USB_STR_MANUF,  # iManufacturer
+            _USB_STR_PRODUCT,  # iProduct
+            _USB_STR_SERIAL,  # iSerialNumber
+            1,
+        )  # bNumConfigurations
+
+        # Iterate interfaces to build the configuration descriptor
+
+        # Keep track of the interface and endpoint indexes
+        itf_num = builtin_driver.itf_max
+        ep_num = max(builtin_driver.ep_max, 1)  # Endpoint 0 always reserved for control
+        while len(strs) < builtin_driver.str_max:
+            strs.append(None)  # Reserve other string indexes used by builtin drivers
+        initial_cfg = builtin_driver.desc_cfg or (b"\x00" * _STD_DESC_CONFIG_LEN)
+
+        self._itfs = {}
+
+        # Determine the total length of the configuration descriptor, by making dummy
+        # calls to build the config descriptor
+        desc = Descriptor(None)
+        desc.extend(initial_cfg)
+        for itf in itfs:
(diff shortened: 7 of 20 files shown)


paulhamsh:
Hi - I'm trying to try this out, but I can't import device or hid.
I am building on a rp2, in WSL, and used these instructions - I guess I am missing something here?

```
cd ~
git clone https://github.com/projectgus/micropython -b feature/usbd_python
cd micropython
sudo apt-get install build-essential libffi-dev git pkg-config

cd mpy-cross
make

cd ..
cd port/rp2
make submodules
cmake .
make
```

It creates a .uf2 and Thonny lets me run command, but ```import device``` and ```import hid``` don't work.
Sorry if I'm missing something obvious.

jimmo:
@paulhamsh Have you copied device.py and hid.py from this PR to the filesystem?

paulhamsh:
> @paulhamsh Have you copied device.py and hid.py from this PR to the filesystem?

Oops - thank you. I thought that, as they were part of micropython-lib, somehow they were part of the build. I've copied them manually and it works well. Amazing! Thank you.

projectgus:
@paulhamsh Right, sorry - I haven't probably integrated these with the micropython-lib build yet (and they probably won't be in the default config, to save size). Glad it's working well - it's still at an early stage so any feedback, requests for different APIs, etc are very welcome!

paulhamsh:
So - I've tried to use this to make a basic MIDI device but it doesn't even make the descriptor - the override functions in midi.py don't seem to be called at all - any idea what I've done wrong?

https://github.com/paulhamsh/Micropython-Midi-Device

Thanks!

projectgus:
@paulhamsh Hey, I saw in a notification that you were working on https://github.com/paulhamsh/Micropython-Midi-Device but it's not working yet. Your code looks like it should work to me at a glance, and I had MIDI on my list of things to implement anyway, so I'll try to take a look at it soon as well!

paulhamsh:
Ah – I had a problem in getting the descriptors to look right, but I put in some debugging and fixed that.

So now it looks like a good set of descriptors, but doesn’t really work for midi.

If I put in the first empty Audio interface, nothing works.

If I leave it as just the MIDI Interface, it worked for a few times (a single message made it, or some in bulk, like they got suck somewhere).

 

This is under Windows 10 using MixiOx to look for midi messages.

projectgus:
I think you're the first person to use this driver code on Windows, so there are many potential buts that you might have found! :). I'll hopefully get to have a play with MIDI (maybe on Windows) myself, next week.

paulhamsh:
It works now – I think I send a bad midi message! Not with Midi OX, but it does work with MidiView and Reaper on Win 11.

But also not with the Audio interface in place, just the single Midi Interface  – which is odd because that is part of the spec.

paulhamsh:
Hi - my Midi code works fine now - my Python obviously isn't up to it, too many 'self's missing.
Thanks for your help!
https://github.com/paulhamsh/Micropython-Midi-Device

projectgus:
@paulhamsh That's great! I hope it's useful. Debugging is definitely the most fiddly part of this, especially if you don't have the REPL moved off of the default USB-CDC port! I'll try to include some debugging tips when I write up the docs for this.

I see you've MIT licensed your midi.py file (thanks!) Would you happy for me to use that as a basis for inclusion in this PR?

paulhamsh:
> @paulhamsh That's great! I hope it's useful. Debugging is definitely the most fiddly part of this, especially if you don't have the REPL moved off of the default USB-CDC port! I scratched my head a number of times before I got a robust way to always see errors, and then things got smoother! I'll try to include some debugging tips when I write up the docs for this.
> 
> I see you've MIT licensed your midi.py file (thanks!) Would you happy for me to use that as a basis for inclusi

> @paulhamsh That's great! I hope it's useful. Debugging is definitely the most fiddly part of this, especially if you don't have the REPL moved off of the default USB-CDC port! I scratched my head a number of times before I got a robust way to always see errors, and then things got smoother! I'll try to include some debugging tips when I write up the docs for this.
> 
> I see you've MIT licensed your midi.py file (thanks!) Would you happy for me to use that as a basis for inclusion in this PR?

Yes of course!  Happy for you to share it!!

hoihu:
> I wanted the REPL log to UART0 or 1, so I tried os.dupterm(machine.UART(0)) and os.dupterm(machine.UART(1)) but without success. The first one just freezes the REPL on USB, the second one succeeds but I did not see any output on the serial console on UART1.

never mind, it was too late during that evening I guess, I had the wrong pins connected. REPL works now via UART. Sorry for the noise

barafael:
I am trying to get serial data off of my pico and stumbled upon these changes and they look fantastical - I want to create a second usb serial device, which leaves the existing serial connection intact. Is this possible with these changes, even with reenumeration?

projectgus:
@barafael Sorry I didn't reply to you earlier about this. I've had almost no time to work on MP lately, but I've dedicated time from July so the momentum should pick up on this PR quite soon.

I know @hoihu started working on CDC serial support on their fork, there is a branch here with a commit: https://github.com/hoihu/micropython-lib/commits/feat-cdc-python - I haven't tested it, it looks like it would enumerate but there's still a bunch of work before it can be used the same as the default USB serial port.

So, unfortunately I don't know of an easy way to get >1 USB-CDC serial port for now - either with or without these changes. Best I can offer is: watch this space!

hoihu:
> but there's still a bunch of work before it can be used the same as the default USB serial port.

Had some time left today and I pushed an update with a basic example on how to use it.
https://github.com/hoihu/micropython-lib/blob/feat-cdc-python/micropython/usbd/cdc_example.py

But as @projectgus said, it's currently still far away from anything. Especially the read() function should be using a ringbuffer or similar. At least it's possible to do some basic rx/tx over that interface now..

andrewleech:
@hoihu does this sound like it's going to be needed? https://github.com/micropython/micropython/pull/9458

RodargDev:
So i've been looking through a lot of PR's and can't seem to find a answer so im just gonna ask it here. I have a Pico W with MicroPython on it and all i want to do is simulate a left mouse click on the device its connected with. Can this pull request make that happen or is that just straight up impossible right now for MicroPython?

turmoni:
Just for the benefit of anyone else looking at this and wanting to borrow code; I've got a HID keypad and basic USB mass storage in my fork. I've created a draft PR to make the changes easy to find at projectgus/micropython-lib#1 .

robin7331:
I've successfully built the UF2 but after copying the files and running `mpremote exec "import hid_test.py"` I'm getting this: 

```
  File "hid_test.py", line 2, in <module>
  File "device.py", line 7, in <module>
ImportError: can't perform relative import
```

Which is referring to this line in `device.py`:
`from .utils import split_bmRequestType` 

Am I missing something? 🧐

projectgus:
@robin7331 Maybe the usbd package hasn't been included properly. Can you please give details about what "copying the files" means in this context? Without the package directory and `__init__.py` file, the relative imports won't work.

projectgus:
@turmoni Regarding the [issues you mentioned](https://github.com/micropython/micropython/pull/9497#issuecomment-1612033951):

> The inability to STALL endpoints

Now implemented as `USBInterface.set_ep_stall()` & `USBInterface.get_ep_stall()`. Have tested these work "on the wire" but not in a context where they matter for protocol correctness (like MSC).

>  Control transfer callbacks not firing for anything other than the first interface that's added

Do you remember which transfers you weren't seeing interface callbacks for?

> Not really seeing a way to handle input on the control endpoint

This was possible but fiddly. I've adapted the [keypad example you wrote](https://github.com/projectgus/micropython-lib/pull/1) to use control transfers for `SET_REPORT` and added it here.

Am currently trying to decide if it's worth building a more intuitive interface for control transfers with data, or to leave it like this.

turmoni:
Thanks, I'll have a look at these at some point!

> Do you remember which transfers you weren't seeing interface callbacks for?

It was all of them - on a device using a HID keypad and a mass storage device, whichever interface I registered first was the only one to have the control transfer callbacks invoked, e.g.:

```python
kp = hidkeypad.KeypadInterface()
ms = msc.MSCInterface(<...>)
ud.add_interface(kp)
ud.add_interface(ms)
ud.reenumerate()
```

Would fire the control callback in KeypadInterface but not MSCInterface, whilst: 

```python
kp = hidkeypad.KeypadInterface()
ms = msc.MSCInterface(<...>)
ud.add_interface(ms)
ud.add_interface(kp)
ud.reenumerate()
```

would fire the control callbacks for MSCInterface but not KeypadInterface. Having HID followed by MSC happens to work out somewhat well because OSs will just try SCSI on the OUT/IN endpoints if they don't get a response to the max LUN request.

I did try to track this down in the C code, and couldn't understand _why_ it wasn't firing. It's been a few weeks now so I'm a bit fuzzy, but it seemed like it wasn't being handled on the C side, even though it looked like it should have been.

akshaykg:
I'm new to this can someone upload an uf2 file with this change please?

projectgus:
@akshaykg Sorry, I thought I'd replied to you already about this, my mistake. At this point this PR is pretty rough. It's really only useful if you're planning to develop it directly, rather than use it in something. Once it's a bit further along, I'll look at whether we can distribute some "early access" binary builds somewhere.

projectgus:
Have rebased (to pick up pre-commit, woo!) and committed a USB CDC serial driver here, along with an example (REPL via `os.dupterm()`) and a performance test. Thanks to @hoihu who developed the initial CDC driver, and @linted who contributed some fixes to it.

Performance is... underwhelming. Reads seem to be 33% to 50% of the throughput of Native USB-CDC, "verified" writes are about 50% and "unverified" about 20%... I'd hoped we'd be closer to 50% of native performance all around :disappointed:.

<details>
<summary>Complete performance test logs</summary>

<pre>
❯ python -u ./cdc_rate_benchmark.py /dev/ttyUSB0 /dev/ttyACM1 1 | tee cdc_rate_usbd.log
REPL on /dev/ttyUSB0
data on /dev/ttyACM1
USB VCP 1
READ: bufsize=256, read 32768 bytes in 191.68 msec = 166.95 kibytes/sec = 1.37 MBits/sec
READ: bufsize=512, read 32768 bytes in 193.90 msec = 165.03 kibytes/sec = 1.35 MBits/sec
READ: bufsize=1024, read 65536 bytes in 393.46 msec = 162.66 kibytes/sec = 1.33 MBits/sec
READ: bufsize=2048, read 131072 bytes in 775.69 msec = 165.01 kibytes/sec = 1.35 MBits/sec
READ: bufsize=4096, read 262144 bytes in 1552.44 msec = 164.90 kibytes/sec = 1.35 MBits/sec
READ: bufsize=8192, read 524288 bytes in 3116.02 msec = 164.31 kibytes/sec = 1.35 MBits/sec
READ: bufsize=16384, read 1048576 bytes in 6214.78 msec = 164.77 kibytes/sec = 1.35 MBits/sec
WRITE: verified=1, bufsize=128, wrote 4096 bytes in 518.10 msec = 7.72 kibytes/sec = 0.06 MBits/sec
WRITE: verified=1, bufsize=256, wrote 4096 bytes in 257.01 msec = 15.56 kibytes/sec = 0.13 MBits/sec
WRITE: verified=1, bufsize=512, wrote 8192 bytes in 513.80 msec = 15.57 kibytes/sec = 0.13 MBits/sec
WRITE: verified=1, bufsize=1024, wrote 16384 bytes in 770.36 msec = 20.77 kibytes/sec = 0.17 MBits/sec
WRITE: verified=1, bufsize=2048, wrote 32768 bytes in 1527.01 msec = 20.96 kibytes/sec = 0.17 MBits/sec
WRITE: verified=1, bufsize=4096, wrote 65536 bytes in 2804.35 msec = 22.82 kibytes/sec = 0.19 MBits/sec
WRITE: verified=1, bufsize=8192, wrote 524288 bytes in 21591.40 msec = 23.71 kibytes/sec = 0.19 MBits/sec
WRITE: verified=0, bufsize=128, wrote 4096 bytes in 503.86 msec = 7.94 kibytes/sec = 0.07 MBits/sec
WRITE: verified=0, bufsize=256, wrote 4096 bytes in 258.68 msec = 15.46 kibytes/sec = 0.13 MBits/sec
WRITE: verified=0, bufsize=512, wrote 8192 bytes in 418.61 msec = 19.11 kibytes/sec = 0.16 MBits/sec
WRITE: verified=0, bufsize=1024, wrote 16384 bytes in 519.21 msec = 30.82 kibytes/sec = 0.25 MBits/sec
WRITE: verified=0, bufsize=2048, wrote 32768 bytes in 1011.07 msec = 31.65 kibytes/sec = 0.26 MBits/sec
WRITE: verified=0, bufsize=4096, wrote 65536 bytes in 1780.98 msec = 35.94 kibytes/sec = 0.29 MBits/sec
WRITE: verified=0, bufsize=8192, wrote 524288 bytes in 13202.22 msec = 38.78 kibytes/sec = 0.32 MBits/sec

❯ python -u ./cdc_rate_benchmark.py /dev/ttyACM1 | tee cdc_native.log
REPL and data on /dev/ttyACM1
READ: bufsize=256, read 32768 bytes in 83.63 msec = 382.62 kibytes/sec = 3.13 MBits/sec
READ: bufsize=512, read 32768 bytes in 94.43 msec = 338.88 kibytes/sec = 2.78 MBits/sec
READ: bufsize=1024, read 65536 bytes in 163.04 msec = 392.53 kibytes/sec = 3.22 MBits/sec
READ: bufsize=2048, read 131072 bytes in 301.30 msec = 424.82 kibytes/sec = 3.48 MBits/sec
READ: bufsize=4096, read 262144 bytes in 502.04 msec = 509.92 kibytes/sec = 4.18 MBits/sec
READ: bufsize=8192, read 524288 bytes in 1123.43 msec = 455.75 kibytes/sec = 3.73 MBits/sec
READ: bufsize=16384, read 1048576 bytes in 2211.46 msec = 463.04 kibytes/sec = 3.79 MBits/sec
WRITE: verified=1, bufsize=128, wrote 4096 bytes in 119.59 msec = 33.45 kibytes/sec = 0.27 MBits/sec
WRITE: verified=1, bufsize=256, wrote 4096 bytes in 103.98 msec = 38.47 kibytes/sec = 0.32 MBits/sec
WRITE: verified=1, bufsize=512, wrote 8192 bytes in 194.74 msec = 41.08 kibytes/sec = 0.34 MBits/sec
WRITE: verified=1, bufsize=1024, wrote 16384 bytes in 375.83 msec = 42.57 kibytes/sec = 0.35 MBits/sec
WRITE: verified=1, bufsize=2048, wrote 32768 bytes in 739.20 msec = 43.29 kibytes/sec = 0.35 MBits/sec
WRITE: verified=1, bufsize=4096, wrote 65536 bytes in 1467.83 msec = 43.60 kibytes/sec = 0.36 MBits/sec
WRITE: verified=1, bufsize=8192, wrote 524288 bytes in 11674.63 msec = 43.86 kibytes/sec = 0.36 MBits/sec
WRITE: verified=0, bufsize=128, wrote 4096 bytes in 52.36 msec = 76.39 kibytes/sec = 0.63 MBits/sec
WRITE: verified=0, bufsize=256, wrote 4096 bytes in 38.10 msec = 104.98 kibytes/sec = 0.86 MBits/sec
WRITE: verified=0, bufsize=512, wrote 8192 bytes in 64.13 msec = 124.75 kibytes/sec = 1.02 MBits/sec
WRITE: verified=0, bufsize=1024, wrote 16384 bytes in 113.82 msec = 140.58 kibytes/sec = 1.15 MBits/sec
WRITE: verified=0, bufsize=2048, wrote 32768 bytes in 217.15 msec = 147.37 kibytes/sec = 1.21 MBits/sec
WRITE: verified=0, bufsize=4096, wrote 65536 bytes in 424.91 msec = 150.62 kibytes/sec = 1.23 MBits/sec
WRITE: verified=0, bufsize=8192, wrote 524288 bytes in 3340.08 msec = 153.29 kibytes/sec = 1.26 MBits/sec
WRITE: verified=0, bufsize=9999, wrote 639936 bytes in 4056.14 msec = 154.07 kibytes/sec = 1.26 MBits/sec
</pre>

</details>

I think the slow "verified" performance on Native really tells the story, the less Python code executing for each USB transfer  then the better.

The client currently uses a pure Python "utils.Buffer" class which is designed around TinyUSB's buffer interface requirements. I'm going to re-test incorporating https://github.com/micropython/micropython/pull/12487, and also maybe look at whether "Buffer" should be implemented in C: possibly as a wrapper around TinyUSB's `tu_fifo` struct (which is already compiled into the binary, and tightly integrates with TinyUSB's APIs.)

hoihu:
awesome👍

> Performance is... underwhelming.

IMO that is already pretty good and more than I would have hoped for! Depends obviously on the use-case, but there will be many applications where speed is not the most critical thing.

Looking forward to the buffer optimisations though😉

DrKendall332:
Finding this very useful, but (perhaps showing my inexperience) I am having trouble sending shift, ctrl, etc through the HID interface.
I can, for example send the 'a' key, and I can send the shift key followed by an 'a'; I just get lowercase 'a's typed either way, although I can leave the system thinking shift is down, so other keyboards send 'A'.
I'm sure I need to send another byte with flags, but I can't make that work without just crashing the device.

projectgus:
Have pushed some changes which include **breaking interface changes** to reduce resource consumption:

* Where previously a `USBInterface` subclass implemented `get_itf_descriptor` and `get_endpoint_descriptors`, now it's only necessary to implement one subclass `descriptor_config_cb`. See the bundled USB class drivers for examples of what this needs to do.
* A single `USBInterface` object can now represent more than one USB Interface (this is something of a "principle of least surprise" naming issue, but it makes cases like MIDI easier where you need to implement multiple USB interfaces grouped together.

There are also some small performance improvements, but not significantly different to the numbers posted above.

projectgus:
@DrKendall332 Sounds like it's pretty close to working, might just need a tweak! Are you able to share the code you're using somewhere?

DrKendall332:
> @DrKendall332 Sounds like it's pretty close to working, might just need a tweak! Are you able to share the code you're using somewhere?

Can't easily post my code right now, but until I do:
I modified the HID_keypad example to move the offset byte, allowing me to send the whole keyboard.
I then tried sending shift as a button before sending the key I want modified. THat put other HID devices in odd states, where they knew shift was down, but it didn't seem to allow my HID device to act as if shifted.
I then tried extending the descriptor to include anotehr byte to represent modifier flags, but that is where I hit a limit - if I send bytes one-at-a-time as int eh keypad, the code runs, but I get no outpput (presumably I am sending the key but it is interpreted as modifier flags with no key) and I cannot figure out how to send 2 bytes (e.g. modifiers and key) without causing an exception.

projectgus:
Note for anyone testing this: the Micropython API (in the other PR) has changed so if compiling with the newest version of that branch then make sure to update this branch to the newest as well.

Luckinber:
Hi @projectgus! I think you're doing the lord's work.

I'm working on passing through a Pi Pico W's uart to a browser through the WebUSB protocol and that originally brought me to this PR and the https://github.com/micropython/micropython/pull/11301 PR. I got the WebUSB specific one working pretty quickly (after getting out of my own way) but I believe in the dream of the method you're building here.

So I cloned your branch of micropython and micropython-lib, compiled a uf2 (following the instructions here) and attempted to run the cdc example but I get:

```
  File "<stdin>", line 5, in <module>
  File "usbd/cdc.py", line 99, in __init__
  File "usbd/device.py", line 43, in get_usbdevice
  File "usbd/device.py", line 86, in __init__
TypeError: extra keyword arguments given
```

Can you provide any guidance for what I'm doing wrong?

Thanks for your time!

bobtedbob:
> Hi @projectgus! I think you're doing the lord's work.
> 
> I'm working on passing through a Pi Pico W's uart to a browser through the WebUSB protocol and that originally brought me to this PR and the [micropython/micropython#11301](https://github.com/micropython/micropython/pull/11301) PR. I got the WebUSB specific one working pretty quickly (after getting out of my own way) but I believe in the dream of the method you're building here.
> 
> So I cloned your branch of micropython and micropython-lib, compiled a uf2 (following the instructions here) and attempted to run the cdc example but I get:
> 
> ```
> Traceback (most recent call last):
>   File "<stdin>", line 5, in <module>
>   File "usbd/cdc.py", line 99, in __init__
>   File "usbd/device.py", line 43, in get_usbdevice
>   File "usbd/device.py", line 86, in __init__
> TypeError: extra keyword arguments given
> ```
> 
> Can you provide any guidance for what I'm doing wrong?
> 
> Thanks for your time!

I get the same error. built everything yesterday from scratch. The only code I have is device = usbd.get_usbdevice()

darkyellow:
I have the same issue from a new build today, the only line of code I have is below.
device = usbd.get_usbdevice()

The error is for this line

     usbd.init(
            descriptor_device_cb=self._descriptor_device_cb,
            descriptor_config_cb=self._descriptor_config_cb,
            descriptor_string_cb=self._descriptor_string_cb,
            open_cb=self._open_cb,
            reset_cb=self._reset_cb,
            control_xfer_cb=self._control_xfer_cb,
     ->       xfer_cb=self._xfer_cb,
        )

It seems to match https://github.com/micropython/micropython/pull/9497/files#diff-a0ab2dbaf982e131b4117d8917ac7549a962a44f294396e2516ce2753e2188dd

darkyellow:
I did a bit of digging and it looks like the error is because the named parameter in usbd.init should be open_itf_cb rather than open_cb.  Changing this and rebuilding fixes the above issue.

Thonny will lose connectivity with the pico though, I have tested that the pico still works but i can't see the debug anymore.

projectgus:
Sorry everyone for this broken state. I started some refactoring work earlier this month and accidentally pushed a change to the other `micropython` PR that didn't have a corresponding update here. Worse, that change has some bugs (beyond the incompatible name change).

If you go to your `micropython` git working directory (not micropython-lib) and run `git switch -d 0b5e6cd3` then this should check out the commit-before-last from that PR, and then this PR should work again.

I've been "almost done" with these refactors for over a week now, hoping to push them. Fingers crossed for an update, soon.

projectgus:
@Luckinber Glad to hear!

> Is it possible to change the USB Platform Capability Descriptors in python code this way? 

Not yet, unfortunately. This requires one more thing in the low-level API ([see these comments](https://github.com/micropython/micropython/pull/9497#issuecomment-1893167133)). It'll probably be added in follow-up PRs from these to avoid scope creep (and these PRs are already monsters), but it's definitely on the radar - and it should be a pretty quick follow-up to add.

Luckinber:
Thanks @projectgus!

That makes sense, I might try my hand at making a shoddy implementation.

darkyellow:
Thanks for the all the work in doing this!

I have got this working with a few different usb descriptors and it works fine. I have had to resort to debugging using an lcd screen because Thonny loses its serial connection to the pico once I reenumerate. It would be good if there was a serial interface you could add to ensure this isn't lost (there might be something already, not sure). I would only need it there for debugging so being able to switch it off would be great. This might be for another PR though.

turmoni:
If you have a USB to TTL serial adapter you can enable the hardware UART (see https://github.com/micropython/micropython/pull/9497 under "debugging tips") and use that

Luckinber:
> Thanks for the all the work in doing this!
> 
> I have got this working with a few different usb descriptors and it works fine. I have had to resort to debugging using an lcd screen because Thonny loses its serial connection to the pico once I reenumerate. It would be good if there was a serial interface you could add to ensure this isn't lost (there might be something already, not sure). I would only need it there for debugging so being able to switch it off would be great. This might be for another PR though.

Are you able to run me through how you did it @darkyellow?

projectgus:
Have just pushed a major refactor to match latest changes in https://github.com/micropython/micropython/pull/9497. Both branches will need to be updated and rebuilt to work together.

If you have existing code written against this module then it will need some changes, take a look at the examples in this PR to see how it behaves now.

Any questions, or problems, then please let me kow.

darkyellow:
I have installed the latest version, things seem to be working like they did before, a couple of questions though...

I don't really understand the static_drivers value, setting it on and off doesn't change Thonny losing REPL ( I suspect there was a serial device connected on startup which never comes back here). Can you explain this more?

Previously I could add an remove interfaces, how do I do this now? I used to test things out by switching descriptiors to emulate different devices but I don't see how to do this now. Do I just call init again with a different interface? Will this remove the old interface? Is there a way to remove all interfaces?

Previous : 

        if usbInterface != None:
               device.remove_interface(usbInterface)
        
        device.add_interface(usbInterface)
        device.reenumerate()

        # wait for host to enumerate as a HID device...
        while not usbInterface.is_open():
            time.sleep_ms(100)

projectgus:
> I don't really understand the static_drivers value, setting it on and off doesn't change Thonny losing REPL ( I suspect there was a serial device connected on startup which never comes back here). Can you explain this more?

Sure. If this is unset, any "static" USB interface(i.e. the native USB-CDC serial port) permanently goes away when you initialise a custom USB device. Upside, you have total freedom in what the "runtime" USB device does.

If set, both the "static" and "runtime" interfaces can coexist, although you may have to be careful about how they're implemented.

Either way, when you call `init` the existing serial port has to vanish first. There's no real way around that apart from moving the REPL to a non-USB UART. The only way for the USB host to see the device has changed is to simulate the device disconnecting and reconnecting. I think eventually it will be possible to add support in `mpremote` to detect the disconnect and reconnect (although some data may be lost). Maybe Thonny might consider a similar feature.

> Previously I could add an remove interfaces, how do I do this now? I used to test things out by switching descriptiors to emulate different devices but I don't see how to do this now. Do I just call init again with a different interface? Will this remove the old interface? Is there a way to remove all interfaces?

Exactly! If you call `init` again with new interface(s), they'll replace the old ones and it will simulate a disconnect/reconnect each time. The itfs argument is `*args` so you can pass more than one.

patfrench:
Hi @projectgus 

I'm posting here because I don't know if you have the time to go through the PR and the discuss.

I wanted to do like the example
image

```pyhon
from usbd import CDC
from machine import UART
import usbd.device
import os
import time

uart1 = UART(1, 115200)

cdc = CDC()
cdc.init(timeout=0)  # zero timeout makes this non-blocking, suitable for os.dupterm()

# pass static_drivers=True so that we get the static USB-CDC alongside,
# if it's available.
usbd.device.get().init(cdc, static_drivers=True)

# wait for host enumerate as a CDC device...
while not cdc.is_open():
    time.sleep_ms(100)

# cdc.is_open() returns true after enumeration finishes.
# cdc.dtr is not set until the host opens the port and asserts DTR
while not (cdc.is_open() and cdc.dtr):
    time.sleep_ms(20)

while True:
     data = cdc.read(-1)
     if (data != None):
         uart1.write(data)

     if uart1.any():
         cdc.write(uart1.read(256))
```
I have 2 questions:

Is this the best way to do it?
Why do I have the impression that it's slow?

Thank you.

projectgus:
@patfrench I've replied to you [on the discussion](https://github.com/orgs/micropython/discussions/13784). Thanks!

projectgus:
Updated this PR to match changes in the low-level PR, again. The API in this PR will almost certainly change again before it's merged, but it should still work in its current state.

Luckinber:
Hi @projectgus!

Since you got the low level PR through yesterday I've been messing around with this API more to get a better understanding of it and I have been having a blast.

After making a script to pass communication between a virtual cdc and uart port [here](https://github.com/orgs/micropython/discussions/13784#discussioncomment-8809041): 

```py
from machine import Pin, UART
from usbd import CDC
import usbd.device
import select
import time

uart = UART(0, baudrate=115200, tx=Pin(16), rx=Pin(17))
uart.init(bits=8, parity=None, stop=1)

cdc = CDC()
cdc.init(timeout=0)
usbd.device.get().init(cdc, builtin_drivers=True)
# Waiting for USB host to configure the interface...
while not cdc.is_open():
    time.sleep_ms(100)
# Waiting for CDC port to open...
while not (cdc.is_open() and cdc.dtr):
    time.sleep_ms(20)
# CDC port is open

io = select.poll()
io.register(uart, select.POLLIN)
io.register(cdc, select.POLLIN)

led = Pin("LED", Pin.OUT)
led.off()

while True:
	for socket, event in io.ipoll(0):
		if socket == uart:
			led.on()
			cdc.write(uart.read(256))
			led.off()
		elif socket == cdc:
			led.on()
			uart.write(cdc.read(256))
			led.off()
```

I'm testing removing the wait conditions and just making the virtual CDC on a wing and a prayer and I'm not getting any issues, are these necessary? Will the CDC happily sit there shouting into the void?

After I've had my fun with this tangent I will look into what it takes to make a derivation of the USBInterface class for WebUSB, with any luck I can help you with some example generation.

Thanks again for the awesome work, the more I use this, the more I love it.

Luckinber:
Okay so after a bit more testing, I have tried this:
```py
from machine import Pin, UART
from usbd import CDC
import usbd.device
import select

uart = UART(0, baudrate=115200, tx=Pin(16), rx=Pin(17))
uart.init(bits=8, parity=None, stop=1)

cdc1 = CDC()
cdc1.init(timeout=0)
usbd.device.get().init(cdc1, builtin_drivers=True)

cdc2 = CDC()
cdc2.init(timeout=0)
usbd.device.get().init(cdc2, builtin_drivers=True)

io = select.poll()
io.register(uart, select.POLLIN)
io.register(cdc1, select.POLLIN)
io.register(cdc2, select.POLLIN)

led = Pin("LED", Pin.OUT)
led.off()

while True:
	for socket, event in io.ipoll(0):
		if socket == uart:
			led.on()
			data = uart.read(256)
			uart.write(data)
			cdc1.write(data)
			cdc2.write(data)
			led.off()
		elif socket == cdc1:
			led.on()
			data = cdc1.read(256)
			uart.write(data)
			cdc1.write(data)
			cdc2.write(data)
			led.off()
		elif socket == cdc2:
			led.on()
			data = cdc2.read(256)
			uart.write(data)
			cdc1.write(data)
			cdc2.write(data)
			led.off()
```

The communication between the ports works fine but cdc2 overwrites cdc1 as I only see one extra COM port, am I initialising them wrong or missing something?

Right now I have COM1 (which is something unrelated), COM3 (REPL), COM4 (Virtual CDC2) and COM5 (UART to USB serial)

Luckinber:
Sorry for the spam but after some further messing around I realised that `usbd.device.get().init()` reenumerates and overwrites predefined interfaces (I think) so I moved to `usbd.device.get().init(cdc1, cdc2, builtin_drivers=True)` which seems to be closer. My code is a fair bit more developed now so sorry if it's a little hard to discern:

```py
from machine import Pin, UART
from usbd import CDC
import usbd.device
import select
import sys
import time

# Create a list of all the ports we want to poll
ports = []


# Create a UART object and add it to the list
uart = UART(0, baudrate=115200, tx=Pin(16), rx=Pin(17))
uart.init(bits=8, parity=None, stop=1)
ports.append(uart)

# Create a CDC object and add it to the list
cdc1 = CDC()
cdc1.init(timeout=0)
ports.append(cdc1)

# Create another CDC object and add it to the list
cdc2 = CDC()
cdc2.init(timeout=0)
ports.append(cdc2)

# Add the standard input to the list
ports.append(sys.stdin)

# Initialise the USB driver(?) with the CDC objects
usbd.device.get().init(cdc1, cdc2, builtin_drivers=True)

# Create a poll object and register all the ports
io = select.poll()
for port in ports:
	io.register(port, select.POLLIN)

led = Pin("LED", Pin.OUT)
led.off()

# Wait for the USB to be ready
time.sleep(1)

while True:
	for stream, event in io.ipoll(0):
		if stream in ports:
			led.on()
			data = stream.read(1)
			# If the data is a carriage return, add a newline
			if data == b'\r':
				data = b'\r\n'
			for port in ports:
				# If the port is the standard input, handle it differently
				if port == sys.stdin:
					if isinstance(data, str):
						print(data, end='')
					else:
						print(data.decode(), end='')
					continue
				port.write(data)
			led.off()
```

Everything enumerates correctly and I see a new, pretty COM6, connecting to it works fine but when I go to send data I get:

```
  File "<stdin>", line 60, in <module>
  File "usbd/cdc.py", line 367, in write
  File "usbd/cdc.py", line 332, in _wr_xfer
  File "usbd/device.py", line 493, in xfer_pending
KeyError: 132
```

Looking at the error it's giving leads me to the write function in `usbd/cdc.py`:

```py
def write(self, buf):
        # use a memoryview to track how much of 'buf' we've written so far
        # (unfortunately, this means a 1 block allocation for each write, but it's otherwise allocation free.)
        start = time.ticks_ms()
        mv = memoryview(buf)
        while True:
            # Keep pushing buf into _wb into it's all gone
            nbytes = self._wb.write(mv)
            self._wr_xfer()  # make sure a transfer is running from _wb

            if nbytes == len(mv):
                return len(buf)  # Success

            mv = mv[nbytes:]

            # check for timeout
            if time.ticks_diff(time.ticks_ms(), start) > self._timeout:
                return len(buf) - len(mv)
```

From what I am understanding the `_wr_xfer()` check is failing meaning the transfer didn't happen but I'm really not sure. Hopeful for any help.

Also if this is the wrong place to be doing this, I'm happy to move to an issue if that would be better.

projectgus:
@Luckinber Thanks for the detailed report. There was a bug where endpoints from multiple interfaces weren't being tracked correctly. The error should be fixed now, please try it out when you get a chance.

scruss:
Report of reenumeration causing rp2 device to move from ttyACM0 to ttyACM1 on Linux: [Dynamic USB Devices - Raspberry Pi Forums](https://forums.raspberrypi.com/viewtopic.php?p=2204794&sid=6f96f8aee64ba4862b8364a200bd7282#p2204745)
(user, for employment reasons, is prohibited from using github so can't participate here)

In my own limited testing, USB disconnect is extremely jarring on soft reboot, causing disconnection from any dev system I'm using.

projectgus:
Thanks for the heads up, @scruss . I'll take a look and reply on there.

> In my own limited testing, USB disconnect is extremely jarring on soft reboot, causing disconnection from any dev system I'm using.

This is definitely the case. We talked around alternative approaches but it's hard to find one on the device end: if the Python code that implements the USB device is being cleaned up by the soft reset then the device has to go away.

I think we can make some tooling improvements for mpremote in some circumstances so that if it sees the port go away and reappear immediately then it reconnects, which will help a lot. I haven't verified this yet, though.

projectgus:
Have performed a very major update, including squashing all the commit history and splitting up packages with new packages names. Please check the README in the PR and the updated examples for details.

micropython/usb/examples/device/midi_example.py
```diff
@@ -0,0 +1,50 @@
...
+# This example demonstrates creating a custom MIDI device.
+#
+# This example uses the usb-device-midi package for the MIDIInterface class.
+# This can be installed with:
+#
+# mpremote mip install usb-device-midi
+#
+# MIT license; Copyright (c) 2023-2024 Angus Gratton
+import usb.device
+from usb.device.midi import MIDIInterface
```
projectgus, line 12: I'm unsure whether it's better to have each sub-package install its own module like this, or to have the top-level usb/device/__init__.py expose all of the public interfaces in that package (in which case it would be `usb.device.MIDIInterface`).

* Putting everything in the top `usb.device` namespace may be a little better for discoverability, as can tab-complete all the "public" class names.
* It seems like no matter what the filesystem will expose the `usb.device.midi` module, so the current version clutters the namespace less overall.
* Relying on the filesystem to discover the separate modules makes the binary code size smaller as `__init__.py` is nice and small. (On the other hand, I'm not sure if relying on the filesystem to expose these modules is a good practice.)
* Some modules contain more than one "public" class, for example `usb.device.keyboard` contains `KeyboardInterface`, `KeyCode`, and `LEDCode`. So namespacing these together in their own keyboard module seems sensible.

projectgus:
Thanks @alanvgreen for the very comprehensive bug report, and for beating me to some Windows testing. :grinning: 

>  Unfortunately Windows was not able to recognise a MIDI device.

There was a bug in the descriptors being generated for the USB MIDI interface. It seems Linux didn't mind, but Windows usbaudio.sys saw it as invalid. Have pushed a fix and this now works for me on Windows 10 and a SAMD board, please let me know if you get a chance to try it.

> At this point the RPi has become unresponsive and I cannot reconnect to the REPL - I take it that this is expectedd. 

The previous version of midi_example.py didn't pass `builtin_driver=True` when configuring the runtime USB device, which meant the built-in serial port vanished when the dynamic USB device became active.

This doesn't really make sense as the example prints a bunch of things to the console, and the default console is always the USB serial port. I've changed this in the example now, it'll still disconnect when it runs as the original port has to temporarily go away. However, you should be able to re-connect now and see output from the running example.

> Not sure how significant this is, but one difference in the before and after is that the "Demanded Current" reduced from 250mA to 100mA.

Good catch! There's an optional `max_power_ma` argument to `init()` to set the MaxPower current. I had this defaulting to 100mA if not explicitly set, but it makes more sense for the default to be whatever the built-in driver reports (as it may vary from device to device). Have fixed so it now does this.

projectgus:
I've now run all the examples on Windows 10 and they seem to work as expected (after the fixes mentioned above - now pushed). If anyone sees something not working under Windows, please let me know!

If anyone has run any of the examples on macOS then please let me know, as I don't have easy access to a Mac for testing.

alanvgreen:
Just retried the midi example on Windows 11 and it's sending notes. Woohoo! Looks like the way is clear to build an RPi USB MIDI controller with MicroPython :)

One small issue: while the CDC came back and I was able to connect a terminal to it, it did not show the output of the print statements. Once I pressed Ctrl-C to interrupt the loop, the REPL came back and worked as expected.

hoihu:
Hi @projectgus , sorry for the late reply. I played around on my Macbook M1 Max, Running Sonoma 14.3.

I started off with the CDC example, naturally ;) 

- In general the provided CDC example works. I see 2 comports that I can connect and have a REPL on both, that's pretty cool :)

- I then tried using `builtin_driver=False` and got
```
  File "<stdin>", line 27, in <module>
  File "usb/device/impl.py", line 84, in init
  File "usb/device/impl.py", line 203, in config
IndexError: bytes index out of range
```

- trying with 3 CDC ports that initially worked too, I got 3 com ports!
image

But I could only open 2 REPL interfaces... Here is my test code 
 ```
import os
import time
import usb.device

# Create a CDC object and add it to the list
cdc1 = CDCInterface()
cdc1.init(timeout=0)
# Create another CDC object and add it to the list
cdc2 = CDCInterface()
cdc2.init(timeout=0)

# Initialise the USB driver(?) with the CDC objects
usb.device.get().init(cdc1, cdc2, builtin_driver=True)

old_term = os.dupterm(cdc1, 1 )
old_cdc1 = os.dupterm(cdc2, 2)
```

Interestingly, the non-working REPL com port was always the one "in the middle", in the above case number 4

**Comment by projectgus** [comment]
Thanks @hoihu and @alanvgreen for the updates!

**IMPORTANT**: The latest version of this PR requires a low-level change in MicroPython's USBDevice, see this PR: https://github.com/micropython/micropython/pull/14253 - without a MicroPython firmware built with this change, the host OS *will* report errors with the dynamic USB devices.

> One small issue: while the CDC came back and I was able to connect a terminal to it, it did not show the output of the 
> print statements. Once I pressed Ctrl-C to interrupt the loop, the REPL came back and worked as expected.

I think this is the limitation of raw REPL versus normal REPL, in which case Ctrl-B will work as well (and not interrupt the loop).

> I then tried using builtin_driver=False and got [...] IndexError: bytes index out of range

Yes, my mistake - a quick fix for max_power_ma didn't work when the built-in driver was disabled. Have pushed a fix for this.

> trying with 3 CDC ports that initially worked too, I got 3 com ports!
> But I could only open 2 REPL interfaces... Here is my test code

I haven't had a chance to try your code yet, but the low-level fix mentioned above was to correct some weird behaviour with CDC on Windows ([reported here](https://github.com/orgs/micropython/discussions/13784#discussioncomment-8993905)). With some luck this issue is related and fixed by the same update. Otherwise I'll make time to try and reproduce on Linux or Windows, soon.

**Comment by hoihu** [comment]
@projectgus all good, my bad. `os.dupterm` just replaces the stream (if given the default slot 0). 

The rp2 port sets `MICROPY_PY_OS_DUPTERM` to 1, so it's only possible to have one REPL via `os.dupterm`. I recompiled it with `MICROPY_PY_OS_DUPTERM` to 2 and that works now with this code here:

```
import os
import time
import usb.device

cdc1 = CDCInterface()
cdc1.init(timeout=0)
cdc2 = CDCInterface()
cdc2.init(timeout=0)
usb.device.get().init(cdc1, cdc2, builtin_driver=False)

os.dupterm(cdc1)
os.dupterm(cdc2, 1)

print(f'cdc1={cdc1}')
print(f'cdc2={cdc2}')
```

so, everything looks good from the mac side with CDC from my point of view.

**Comment by hoihu** [comment]
I tested the midi example on macOS. It looks good! I was able to run the example, open garage band and the piano was playing notes 👍

But I also observed that I did not get any print statements after the example starts. 

I'm using a setup with the UART REPL. I'm just copy&pasting the code of the exapmle to that REPL and it shows:

```
Waiting for USB host to configure the interface...
Starting MIDI loop...
```

(from here on, I'm able to listen to the note being played). But no print statements regarding the notes on the REPL... 
That sounds somewhat similar as https://github.com/micropython/micropython-lib/pull/558#issuecomment-2036836934

**Comment by hoihu** [comment]
I digged a little deeper in the non-print-statements and it looks that `def _rx_cb(self, ep, res, num_bytes):` is not called (in `midi.py`). 

But I don't know what that means. Maybe that's depending on the installed GUI tools (garage band in my example)? (sorry I'm an absolute novice here on the midi side... any recommendation on simpler tools than garage band to interact with midi devices?)

**Comment by jzhvymetal** [comment]
I'm making my own custom ABS mouse with both vertical and horizontal wheel descriptors. I placed 'usb.device.get().init' in the 'boot.py' because if it's placed in my main code, which I'm currently debugging, the initialization will reboot the REPL console. Everything works correctly when using the REPL console using the object. Since it's declared in the 'boot.py', how can I utilize it in 'code.py' or program being run from REPL console.  Is there a function simular to CP find_hid_device?

boot.py
```
import usb.device
from KM_MPHID_ACTORS import Mouse as MouseInterface
Mouse= MouseInterface()
usb.device.get().init(Mouse , builtin_driver=True)
```
 In REPL console I can do the following and it postions the mouse cursor to the pixels of the screen
`Mouse.move(500,500)`

**Comment by jzhvymetal** [comment]
I fixed the issues above by creating a shared variable import so that both 'boot.py' and 'code.py' can access it.

Now, the problem I'm encountering is that the mouse performance is very slow and lags. I am working on a project that involves creating a Barrier Client (https://github.com/jzhvymetal/PY_Barrier) which receives mouse and keyboard commands via Barrier Server over WiFi and then relays them to HID. I implemented the same code on CircuitPython, and the response from the mouse has no lag. Additionally, I implemented it in both MP and CP by using a serial CH9329, which also showed no lag. Is there anything I'm missing to improve the MP HID performance?

All of my code is available on the Git repository below. Simply copy all the .py files to the 'lib' directory, rename the correct 'boot.py' for the correct environment, and place it in the root directory.

https://github.com/jzhvymetal/PY_Barrier

**Comment by hoihu** [comment]
just as a follow up to my midi tests. I used  [MIDI Monitor](https://www.snoize.com/MIDIMonitor/) to debug the midi notes and the notes come in reliably, with the control value increasing, as expected:

image

It let that run for several hours without issues.

**Comment by projectgus** [comment]
> Now, the problem I'm encountering is that the mouse performance is very slow and lags.

Neat project, @jzhvymetal! Where you call `super().send_report(self.report)`, try adding `timeout_ms=0` to the arguments.

(If an existing report is pending, send_report will block for up to the timeout - default 100ms - before returning. Because you're queueing the same buffer object each time, you can return immediately - the contents of the buffer are what is sent, so if there's one pending then it should be updated by the time the report is actually sent to the host.)

This part of the base class might need to be changed (at minimum, possibly change the default to 0ms.)

If this isn't it, let me know and I'll debug some more.

**Comment by projectgus** [comment]
> But I also observed that I did not get any print statements after the example starts.
> ...
> I digged a little deeper in the non-print-statements and it looks that `def _rx_cb(self, ep, res, num_bytes):` is not called (in `midi.py`).

@hoihu I think this might be a poorly commented example on my part. The example does two things - it sends MIDI notes in a loop for as long as the code keeps running. This part doesn't print anything.

At the same time, if the host sends any MIDI data to the device then it should print something when it receives that MIDI data.

Each direction (IN/OUT) is usually configured separately in MIDI settings (as most devices only transfer MIDI one way). So if the device is configured in Garage Band as only a MIDI input, the current example doesn't print anything. There is also sometimes an option on the host named "MIDI Thru" (or similar), in this mode the host will echo the received "IN" midi messages back "OUT" to that device.

I'll change the example so it prints when operating in both directions.

> any recommendation on simpler tools than garage band to interact with midi devices?

I'm not a MIDI expert either, so my go-to for testing has been the [Virtual MIDI Piano Keyboard](https://vmpk.sourceforge.io/). When configured with an external device as MIDI Input, the keys are highlighted when notes are received. When configured with an external device as MIDI Output, clicking the keys will send messages.

**Comment by projectgus** [comment]
Hmm OK, it looks like there is something else going on with MIDI Out. In some of my runs it works fine in both directions, but sometimes the host sends a few MIDI OUT packets and then stops (MIDI IN keeps flowing). I don't know why, the USB traffic shows the transfers are successful.

This is on Linux with ALSA but I expect whatever the bug is might be the same on all OSes. Still debugging...

**Comment by jzhvymetal** [comment]
> 
> Neat project, @jzhvymetal! 
Thank...needed to learn Python what better way than have a project


>Where you call `super().send_report(self.report)`, try adding `timeout_ms=0` to the arguments.
Perfect, this fixed the problem... I should have delved deeper down the rabbit hole. Now it seems obvious, with the argument named `timeout_ms`. I also lost 2 hours because the Git version of my code for `KM_MPHID_ACTORS.py` was not the latest. I must have forgotten that I saved it down at the board level and didn't make a local copy. 🤦 I also forgot that another board was running with the same device name, which rejects the second device connecting. 🤦🤦

**Comment by jzhvymetal** [comment]
> Where you call `super().send_report(self.report)`, try adding `timeout_ms=0` to the arguments.

Follow-up question: I updated my code to include a custom keyboard device. I noticed that if I set the value to 0, it would sometimes keep the keys down even though the report was sent again without it. I changed the value to 3ms, and it seems to fix the problem. What is the proper usage, and when should you include a delay?

Reviewing the code from CircuitPython custom descriptors, they include a report ID, for example, `0x85, 0x0B, # Report ID [11 is SET at RUNTIME]`, and you pass the ID to the creation of the HID device. With the MP implementation, I had to comment out the Report ID. Is there any way to keep the Report ID for consistency?

Once again, thanks for the implementation. Without it, I would not have been able to code my project on MP. I know the main focus is on rp2 and SAMD boards, but if you managed to also get ESP32 working. I would love to test it. In any case, great job so far.

File: micropython/usb/usb-device-mouse/usb/device/mouse.py [hunk 2]
```diff
@@ -0,0 +1,118 @@
...
+            raise ValueError("dx")
+        if not -127 <= dy <= 127:
+            raise ValueError("dy")
+        self.send_report(dx, dy)
+
+
+# Basic 3-button mouse HID Report Descriptor.
+# This is cribbed from Appendix E.10 of the HID v1.11 document.
+_MOUSE_REPORT_DESC = bytes(
+    [
```
felixdoerre, line 66: I think this report descriptor will compile to far larger code than the descriptor for `usb/device/keyboard.py`. In particular, the (implicit) concatenation of `b"..."`-literals is stored as a constant next to the code, while this variant of constructing the `bytes` object generated a whole bunch of bytecode.

projectgus: You're totally right Felix. This is something I meant to come back to and clearly forgot, appreciate the reminder!

hoihu:
> please let me know if it's still not working as expected.

yes it does. Thanks for the updates.

I also tested the mouse and keyboard_hid examples. They work too on macOS. So looking really good here 👍

micropython/usb/usb-device-mouse/usb/device/mouse.py
projectgus: Fixed to use same approach as keyboard, thanks again for the reminder!

projectgus:
Hi @jzhvymetal,

> Follow-up question: I updated my code to include a custom keyboard device. I noticed that if I set the value to 0, it would sometimes keep the keys down even though the report was sent again without it. I changed the value to 3ms, and it seems to fix the problem. What is the proper usage, and when should you include a delay?

Yes, in retrospect setting `timeout_ms=0` was papering over the root cause of the lags you were seeing: when a timeout was set, `send_report()` was sleeping for 50ms between tries. This was way too long and probably causing the visible stalls/jerkiness. If you pull the latest version of this PR then you can hopefully go back to the default 100ms without any stalls, or any lost reports.

(There is still potentially a race condition that could miss an event when the caller reuses the same bytearray object as the buffer every time, an interim report might not be sent to the host because its data has been replaced by a later report. I've just updated the PR so `mouse.py` and `keyboard.py` demonstrate two different ways to avoid this - mouse.py checks `busy()` before updating the buffer, and `keyboard.py` uses double buffering. However this also might not be a problem you care about, it shouldn't cause stuck keys or buttons.)

> Reviewing the code from CircuitPython custom descriptors, they include a report ID, for example, 0x85, 0x0B, # Report ID [11 is SET at RUNTIME], and you pass the ID to the creation of the HID device. With the MP implementation, I had to comment out the Report ID. Is there any way to keep the Report ID for consistency?

Yes, report IDs are optional in USB HID but they should work. Include them in the HID descriptors, same as CircuitPython does, and then pass the report ID as as the first byte in the data argument of `send_report()`. (CircuitPython uses the TinyUSB driver, and it passes the report ID as a separate argument but then copies it as the first byte of the buffer that is sent "on the wire".)

> Once again, thanks for the implementation. Without it, I would not have been able to code my project on MP. I know the main focus is on rp2 and SAMD boards, but if you managed to also get ESP32 working. I would love to test it. In any case, great job so far.

Thanks! Supporting ESP32-S series is on the radar. There is some low level work that needs to be done first, because ESP-IDF runs TinyUSB in its own RTOS task. This makes it harder to hook the TinyUSB callbacks into MicroPython, but it will be possible to make it work.

micropython/usb/examples/device/midi_example.py
projectgus: Discussed this with @dpgeorge just now, and looks like we'll keep this approach for USB even though it's a little different to lora & aioble.

projectgus:
Have performed another rebase & squash including a breaking change for anyone who has implemented their own class driver: `usb.device.impl` module is now named `usb.device.core`.

rafaellago:
Hello, not sure I'm missing something, I was trying the KeyboardExample provided, and when it gets to `usb.device.get().init(k, builtin_driver=True)` my rpi pico just reboots. I'm using MP v1.23.0-preview.332.g49ce7a607, on win10 and thonny (alhtough I tested connecting directly to it via puTTY and had the same behaviour sending the code via the REPL). 

Sample sending the lines through the REPL

```
MicroPython v1.23.0-preview.332.g49ce7a607 on 2024-04-22; Raspberry Pi Pico with RP2040

>>> import usb.device
>>> from usb.device.keyboard import KeyboardInterface
>>> kbd = KeyboardInterface()
>>> usb.device.get().init(kbd, builtin_driver=True)
PROBLEM IN THONNY'S BACK-END: Exception while handling 'execute_source' (ConnectionError: EOF).
See Thonny's backend.log for more info.
You may need to press "Stop/Restart" or hard-reset your MicroPython device and try again.


Process ended with exit code 1.
────────────────────

Connection lost -- EOF

Use Stop/Restart to reconnect.

Process ended with exit code 1.
```

Also backend.log from thonny if that helps somehow.

[backend.log](https://github.com/micropython/micropython-lib/files/15082110/backend.log)

Thanks for any help, and sorry if I'm not being smart enough.

projectgus:
> Hello, not sure I'm missing something, I was trying the KeyboardExample provided, and when it gets to `usb.device.get().init(k, builtin_driver=True)` my rpi pico just reboots

Hi @rafaellago,

Thanks for the comprehensive report, and you are certainly being smart enough.

The big limitation with the dynamic USB support is that reconfiguring the USB port requires disconnecting any existing USB devices as well (i.e. the USB serial port holding the REPL in most cases), as we have to simulate an unplug-then-replug of the whole USB device so the host sees any new interfaces. This also has to happen if MicroPython soft resets while a dynamic USB device is currently active.

If the whole program has already been sent to MicroPython then it may actually be running after the port goes away, and you could reconnect with a program like PuTTY and see the output from the running program.

How well any of this works with Thonny is an open question, I haven't looked closely at Thonny's behaviour - it may be that you can also reconnect in Thonny after getting this error and your program will be running with the dynamic USB active, but it also could be that Thonny sends a soft reset each time it reconnects and this is causing the dynamic USB device to immediately disappear again (along with the USB COM port).

I'm planning some work to improve mpremote's support for the USB serial port temporarily going away like this, and I think something similar could be added to Thonny. However, as I said I haven't looked into Thonny specifically and any feature would obviously be up to the Thonny maintainers.

rafaellago:
Hello @projectgus 

As you said, maybe Thonny is restarting the pico on the connection loss, testing through putty and restarting the session actually brought me back to the repl and I managed to send keystrokes to my PC. 

If I manage to make a multimedia keyboard example (just volume up/down/mute) can I send a PR to your repo? Or is it better to send it directly to the micropython one?

Thank you so much for your explanation.

projectgus:
> If I manage to make a multimedia keyboard example (just volume up/down/mute) can I send a PR to your repo? Or is it better to send it directly to the micropython one?

This PR is already pretty huge so I think the best would be to wait until this PR is merged and then submit a new PR against this repo that just adds the example.

If you have any questions as you work on the keyboard program then feel free to ask. Although if they're not directly related to this PR then it might be best to start a Discussion and then make a comment here that links to it (otherwise things tend to pile up a bit.)

projectgus:
> > Whats missing to get it merged?
> 
> @f4grx testing on mac os, I think

Oops, I neglected to update the TODO in the first post until just now, but @hoihu [has tested the examples on macos without issue](https://github.com/micropython/micropython-lib/pull/558#issuecomment-2051208763) :pray: 

The README is still a little bare bones, but the main thing that's still pending is probably code review from some of the other maintainers.

micropython/usb/README.md
```diff
@@ -0,0 +1,85 @@
+# Dynamic USB packages
```
dpgeorge, line 1: I suggest a title like "USB drivers" or "The usb package".

micropython/usb/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+These packages allow implementing USB functionality on a MicroPython device using pure Python code.
```
dpgeorge, line 3: Suggest "MicroPython device" -> "MicroPython system" to avoid confusion with device meaning USB device.

micropython/usb/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Currently only USB device is implemented, not USB host.
+
+## USB Device support
+
+### Support
+
+USB Device support depends on the low-level [machine.USBDevice](https://docs.micropython.org/en/latest/library/machine.USBDevice.html) class. This class is new and not supported on all ports, so please check the documentation for your MicroPython version. It is possible to implement a USB device using only the low-level USBDevice class. These packages are higher level and easier to use.
```
dpgeorge, line 11: Please wrap lines to approx 80 columns (see eg the lora readme).

dpgeorge: Suggest "These packages are higher level and easier to use." -> "The packages here are ...".

micropython/usb/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Setting `builtin_driver=True` means that any built-in USB serial port will still
+be available. Otherwise, you may permanently lose access to MicroPython until
+the next time the device resets.
+
+See [Unexpected serial disconnects](#Unexpected-serial-disconnects), above, for
+an explanation of possible errors or disconnects when the runtime USB device
+initialises.
+
+Placing the call to `usb.device.get().init()` into the `boot.py` of the MicroPython file system allows the runtime USB device to initialise immediately on boot, before any built-in USB. However, note that calling this function on boot without `builtin_driver=True` will make the MicroPython USB serial interface permanently inaccessible until you "safe mode boot" (on supported boards) or completely erase the flash of your device.
```
dpgeorge, line 53: Suggest adding a sentence at the end of this para like "This is a feature (not a bug) and allows you full control over the USB device, for example to only enable USB HID and prevent REPL access to the system."

micropython/usb/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Package usb-device-hid
+
+This package provides the `usb.device.hid` module. USB HID (Human Interface Device) class allows creating a wide variety of device types. The most common are mouse and keyboard, which have their own packages in micropython-lib. However, using the usb-device-hid package directly allows creation of any kind of HID device.
+
+See [hid_custom_keypad_example.py](examples/device/hid_custom_keypad_example.py) for an example of a Keypad HID device with a custom HID descriptor.
+
+### Package usb-device-cdc
+
+This package provides the `usb.device.cdc` module. USB-CDC (Communications Device Class) is most commonly used for virtual serial port USB interfaces, and that is what is supported here.
```
dpgeorge, line 71: Suggest "USB CDC" (no hyphen) to be consistent with "USB HID" above.

micropython/usb/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+The example [cdc_repl_example.py](examples/device/cdc_repl_example.py) demonstrates how to add a second USB serial interface and duplicate the MicroPython REPL between the two.
+
+### Package usb-device-midi
+
+This package provides the `usb.device.midi` module. This allows implementing MIDI devices in MicroPython.
+
+The example [midi_example.py](examples/device/midi_example.py) demonstrates how to create a simple MIDI device to send MIDI data to the USB host.
+
+### Package usb-device
```
dpgeorge, line 81: Suggest putting this section before all the `usb-device-xxx` sections, because it is the most important one and logically should go first.

micropython/usb/examples/device/cdc_repl_example.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# Note that if you run this example on the built-in USB CDC port via 'mpremote
+# run' then you'll have to reconnect after it re-enumerates, and it may be
+# necessary afterward to type Ctrl-B to exit the Raw REPL mode and resume the
+# interactive REPL back.
+#
+# This example uses the usb-device-cdc package for the CDCInterface class.
+# This can be installed with:
+#
+# mpremote mip install usb-device-cdc
```
dpgeorge, line 14: I suggest simple yet full instructions to run this example (I know it's described in the README but I think adding it to each example is useful); eg:
```
To run this example:
1. Make sure `usb-device-cdc` is installed via: mpremote mip install usb-device-cdc
2. Run the example via: mpremote run cdc_repl_example.py
```

micropython/usb/examples/device/hid_custom_keypad_example.py
```diff
@@ -0,0 +1,137 @@
...
+# This example demonstrates creating a custom HID device with its own
+# HID descriptor, in this case for a USB number keypad.
+#
+# For higher level examples that require less code to use, see mouse_example.py
+# and keyboard_example.py
+#
+# This example uses the usb-device-hid package for the HIDInterface class.
+# This can be installed with:
+#
+# mpremote mip install usb-device-hid
```
dpgeorge, line 12: Suggest adding full (but short) instructions to run the example, as above for the CDC example.

micropython/usb/usb-device/tests/test_core_buffer.py
```diff
@@ -0,0 +1,98 @@
+# Tests for the Buffer class included in usb.device.core
+#
+# The easiest way to run this is using unix port. From the top-level usb-device directory,
+# run as:
+#
+# $ micropython -m tests.test_core_buffer
+#
+import micropython
+from usb.device import core
+from core import Buffer
```
dpgeorge, line 10: This line fails.

micropython/usb/usb-device/tests/test_core_buffer.py
```diff
@@ -0,0 +1,98 @@
...
+            return micropython.heap_lock()
+
+        def enable_irq(self, was_locked):
+            if not was_locked:
+                micropython.heap_unlock()
+
+    core.machine = FakeMachine()
+
+
+b = Buffer(16)
```
dpgeorge, line 26: Should be `core.Buffer`.

micropython/usb/usb-device/usb/device/core.py
```diff
@@ -0,0 +1,851 @@
...
+        _usbd = self._usbd
+
+        if self.active():
+            raise OSError(_MP_EINVAL)  # Must set active(False) first
+
+        # Convenience: Allow builtin_driver to be True, False or one of
+        # the machine.USBDevice.BUILTIN_ constants
+        if isinstance(builtin_driver, bool):
+            builtin_driver = _usbd.BUILTIN_DEFAULT if builtin_driver else _usbd.BUILTIN_NONE
+        builtin = _usbd.builtin_driver = builtin_driver
```
dpgeorge, line 116: The local `builtin` variable is never used, it seems.

We should probably enable F841 in `pyproject.toml` to catch these cases.

dpgeorge:
Now merged!!  Thanks to everyone who contributed here, with testing, feedback and coding.

projectgus:
Folks, now this PR is merged some good places to continue discussion are:

* [Discord #usb channel](https://discord.com/channels/574275045187125269/1234766206058369046)
* [Discussions](https://github.com/orgs/micropython/discussions)

... and of course if you find a bug then please open a new Issue on this repo.

PR #560, opened 2022-10-28
aiorepl: Ignore second linefeed immediately after first

sandyscott:
Pressing Enter or Return in a windows client will send two characters. Both will received because aiorepl reads characters one-by-one, both will be a linefeed after universal newline conversion.

This change means the second linefeed is ignored

### DIFF
--- a/micropython/aiorepl/aiorepl.py
+++ b/micropython/aiorepl/aiorepl.py
@@ -104,8 +104,8 @@ async def task(g=None, prompt="--> "):
             cmd = ""
             while True:
                 b = await s.read(1)
-                c = ord(b)
                 pc = c  # save previous character
+                c = ord(b)
                 pt = t  # save previous time
                 t = time.ticks_ms()
                 if c < 0x20 or c > 0x7E:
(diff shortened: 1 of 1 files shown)


dpgeorge:
I really think that it needs to read from `sys.stdin.buffer` because that stream is in binary mode (as opposed to `sys.stdin` which is in text mode and interprets incoming characters as utf-8 encoded).

sandyscott:
I was trying to keep this as minimal as possible, because I'm not sure what other effects that change might have, so I'll leave the PR open, and I have a slightly different one at https://github.com/micropython/micropython-lib/pull/562 which implements the change to `.buffer`

sandyscott:
As mentioned on the other PR - switching to binary mode introduces other issues, so I'll leave this open as a simple fix to resolve 1 windows newline being interpreted as 2

jimmo:
Thanks @sandyscott -- Merged in 81c1408a0763940f9038e104aa38baa16c120acb and d6eb5b6f7e26ed44c828372c4c987970e939716f.

PR #562, opened 2022-10-31
aiorepl: Use stdin in binary mode and catch windows enter presses

sandyscott:
Changed this to use stdin.buffer instead of stdin. Now catches and ignores the consecutive line ending characters within 20ms of the previous.

### DIFF
--- a/micropython/aiorepl/aiorepl.py
+++ b/micropython/aiorepl/aiorepl.py
@@ -101,7 +101,7 @@ async def task(g=None, prompt="--> "):
         while True:
             hist_b = 0  # How far back in the history are we currently.
             sys.stdout.write(prompt)
-            cmd = b''
+            cmd = b""
             while True:
                 b = await s.read(1)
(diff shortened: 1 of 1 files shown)


dlech:
> Changed this to use stdin.buffer instead of stdin.

Is this going to cause problems with multi-byte unicode characters since it now reads one byte instead of one unicode character?

sandyscott:
Having thought about this a bit more, this is going to get messy fast. The aiorepl server would need knowledge of the terminal's character encoding to properly decode the characters. UTF-8 is not a safe assumption: if the terminal is using an 8-bit character set and sends any single-byte character above 0xBF this could cause the server to hang while it waits for the second byte that will never come. The code would also need a fairly significant restructure to separate the decoding from the actions. I think the simpler fix in https://github.com/micropython/micropython-lib/pull/560 is a safer way to fix the specific issue I encountered - that windows line endings are interpreted as two presses of the enter key.

PR #564, opened 2022-11-05
aioble/client.py: Fix default for the `response` arg to char.write().

### LINKED ISSUE #563: Undefined variable....
https://github.com/micropython/micropython-lib/blob/d0f97fc218f07c381c835d9f632904c1ae1c9d6b/micropython/bluetooth/aioble/aioble/client.py#L278

Expect this is meant to be _FLAG_WRITE without a S

jimmo:
- `_FLAG_WRITE` was incorrectly `_FLAGS_WRITE`
- `response` should be defaulted to `None` rather than `False` in order
  to detect that when it is unspecified.

@Enasynco this should fix #563

### DIFF
--- a/micropython/bluetooth/aioble/aioble/client.py
+++ b/micropython/bluetooth/aioble/aioble/client.py
@@ -269,16 +269,13 @@ class BaseClientCharacteristic:
             characteristic._read_status = status
             characteristic._read_event.set()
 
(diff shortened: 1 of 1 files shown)


dpgeorge:
Looks good, thanks.  Rebased and merged in 50b7aca1717291e2b7bb5f423661105545ab9054

PR #565, opened 2022-11-06
shutil: Add disk_usage method

ThinkTransit:
Returns total, used, free size in bytes from provided path.

### DIFF
--- a/python-stdlib/shutil/shutil.py
+++ b/python-stdlib/shutil/shutil.py
@@ -27,3 +27,12 @@ def copyfileobj(src, dest, length=512):
             if not buf:
                 break
             dest.write(buf)
+            
+def disk_usage(path):
+    bit_tuple = os.statvfs(path)
+    blksize = bit_tuple[0] # system block size
+    total = bit_tuple[2] * blksize
+    free = bit_tuple[3] * blksize
+    used = total - free
+
+    return total, used, free


andrewleech:
This is really very handy thanks! 

The cpython version does return a namedtuple which I do think would be nice here too, though certainly it adds a few extra bytes of usage in the field name strings... hard to say which is best.

ThinkTransit:
Thanks Andrew I agree that a named tuple is much more user friendly however I noticed that os.statvfs in micropython doesn't return a named tuple and cpython does. I wonder is it a design principle in micropython to not use namedtuples?

andrewleech:
The built-in modules certainly follow a design principle of smaller-is-better as that's code that everyone has installed regardless of whether they use it or not. 

Libraries in here though can afford to be a bit bigger if the functionality is "worth it" as it's an optional install for people who need it. It's still a balancing act though to decide if the size for any one feature is justified.

jimmo:
Thanks @ThinkTransit !

I think this should return a namedtuple -- especially in micropython-lib we should match CPython.

ThinkTransit:
Thanks @jimmo PR updated to return namedtuple, same as cpython

python-stdlib/shutil/shutil.py
```diff
@@ -27,3 +28,14 @@ def copyfileobj(src, dest, length=512):
             if not buf:
                 break
             dest.write(buf)
+            
+def disk_usage(path):
+    _ntuple_diskusage = namedtuple('usage', 'total used free')
```
dpgeorge, line 33: This will recreate the namedtuple every time this function is called, which is not really efficient.

You could create it at the global scope unconditionally.  Or make `_ntuple_diskusage` global and start out as `None`, then populate it with the namedtuple the first time this function is called.

Also, better to use `namedtuple('usage', ('total', 'used', 'free'))`, it's more compatibly with MicroPython ports (that don't enable `MICROPY_CPYTHON_COMPAT`) and more efficient when the code is frozen.

ThinkTransit: Thanks @dpgeorge the latest commit should resolve both of those issues.

PR #568, opened 2022-11-08
nrf24l01test.py: Add RP2 support, fix ESP32.

peterhinch:
Use explicit pin numbers to instantiate the SPI interface on RP2. On ESP32 use SoftSPI(...) rather than SPI(-1, ...).

Update terminology to initiator/responder.

Tested with two Pico boards.

### DIFF
--- a/micropython/drivers/radio/nrf24l01/nrf24l01test.py
+++ b/micropython/drivers/radio/nrf24l01/nrf24l01test.py
@@ -93,10 +93,7 @@ def initiator():
         # delay then loop
         utime.sleep_ms(250)
 
-    print(
-        "initiator finished sending; successes=%d, failures=%d"
-        % (num_successes, num_failures)
-    )
+    print("initiator finished sending; successes=%d, failures=%d" % (num_successes, num_failures))
 
 
 def responder():
@@ -148,6 +145,4 @@ print("NRF24L01 pinout for test:")
 print("    CE on", cfg["ce"])
 print("    CSN on", cfg["csn"])
 print("    SPI on", cfg["spi"])
-print(
-    "run nrf24l01test.responder() on responder, then nrf24l01test.initiator() on initiator"
-)
+print("run nrf24l01test.responder() on responder, then nrf24l01test.initiator() on initiator")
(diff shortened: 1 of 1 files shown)


micropython/drivers/radio/nrf24l01/nrf24l01test.py
```diff
@@ -7,19 +7,21 @@
...
+_RESPONDER_SEND_DELAY = const(10)
 
 if usys.platform == "pyboard":
     cfg = {"spi": 2, "miso": "Y7", "mosi": "Y8", "sck": "Y6", "csn": "Y5", "ce": "Y4"}
 elif usys.platform == "esp8266":  # Hardware SPI
     cfg = {"spi": 1, "miso": 12, "mosi": 13, "sck": 14, "csn": 4, "ce": 5}
 elif usys.platform == "esp32":  # Software SPI
     cfg = {"spi": -1, "miso": 32, "mosi": 33, "sck": 25, "csn": 26, "ce": 27}
+elif usys.platform == "rp2":  # Hardware SPI with explicit pin definitions
+    cfg = {"spi": -2, "miso": 4, "mosi": 3, "sck": 2, "csn": 5, "ce": 6}
```
dpgeorge, line 24: Would it be simpler to make this a dict that can be passed directly to the SPI constructor, like `SPI(**cfg)`?

Or even simpler, just create the SPI object here, eg `spi = SPI(0, sck=Pin(2), ...)`?

peterhinch: > Or even simpler, just create the SPI object here, eg spi = SPI(0, sck=Pin(2), ...)?

I agree - I was aiming to minimise changes. We would still need `mosi`, `miso` and `sck` in the dict because of the print statements at the end.

Do you want to make this change for each supported platform?

dpgeorge: Yes, please make the change for each platform, if you have the time.

Also, instead of printing the SPI pins at the end, it could just print the `spi` object, which in some cases will show the pins anyway (also baudrate etc).

peterhinch:
Perhaps someone can explain why this fails code formatting. I ran black over the code and the format check wants to revert some of its changes. I thought black was the MP standard?

dpgeorge:
Try `black -l99` (99 chars max line length).

peterhinch:
Thank you, it now passes.

PR #570, opened 2022-11-09
datetime: Add default utc tz for datetime.now().

ThinkTransit:
This is a reauthor of [#561 ](https://github.com/micropython/micropython-lib/pull/561)

The datetime module doesn't support naive datetimes and throws an unimplemented error when calling some of the methods such as datetime.now().

Rather than passing a tz object to these methods on each call I feel it would be best to use a default tz (utc) when a tz isn't provided. This would be more CPython compliant for the end user.

This pr adds a default utc timezone in fromtimestamp() if no tz is supplied.

Signed-off-by: Patrick Joy <patrick@joytech.com.au>


ThinkTransit:
@jimmo @dpgeorge  any chance we can get this one reviewed/merged? I imagine datetime.now() would be one of the more common methods in the datetime library and it would be good to support it and match cpython.

lorcap:
I'm not in favor of this merge. From CPython 3.11.5 documentation, `datetime.now()` is expected to _return the current **local** date and time_.  MicroPhython doesn't have any notion of local timezone, yet. Current library behaviour is correct, IMHO. User shall provide a valid date and time. Or, if/when discussion [12378](https://github.com/orgs/micropython/discussions/12378) leads to a proper timezone support, `datetime.now()` results will be different.

dpgeorge:
This has been superseded by #1045.  `datetime.datetime.now()` is now working.

PR #573, opened 2022-11-10
tempfile module and shutil.rmtree fix

BrianPugh:
`shutil.rmtree` had a reference to `os.walk`, which isn't in micropython. I updated it to use `os.ilistdir` and added unittests for it.

Initial implementation of the `tempfile` module. It currently has the following implemented:

* `mkdtemp`
* `TemporaryDirectory`

I only ever use `TemporaryDirectory`, and I assume others may have similar experience.

This is to make it much easier for me to update my pathlib unittests to be micropython-compatible.

### DIFF
--- a/python-stdlib/shutil/shutil.py
+++ b/python-stdlib/shutil/shutil.py
@@ -5,11 +5,14 @@ from collections import namedtuple
 _ntuple_diskusage = namedtuple("usage", ("total", "used", "free"))
 
 
-def rmtree(top):
-    for path, dirs, files in os.walk(top, False):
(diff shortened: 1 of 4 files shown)


dpgeorge:
Thank you, these are good improvements and additions!

Merged in ee286ed28c24ad163783128a6238f5762cf9cd9b through a99b80186d4ac55b78a380bd66708ba67df63712 with reworded commit messages.

I also added commit 56dc65b6a7535858bf8067cdf0baa16f6cb72834 which adds `manifest.py` to `tempfile.

PR #574, opened 2022-11-14
sdcard: Set MISO high before readblocks/writeblocks.

jimmo:
Originally by @peterhinch.
See https://github.com/micropython/micropython/pull/6007 for discussion.

The summary is that on some cards (especially older Kingston ones) if the bus is shared with other SPI devices, then it seems to require that MISO is high for a few cycles before the transaction is initiated.

Because CS is high, this change should otherwise be a no-op.


dpgeorge:
Thanks!

jimmo:
Thanks @peterhinch!! Sorry this took so long.

PR #582, opened 2022-12-07
sys: Implement intern().

stinos:
Came across this when using the pure python [msgpack implementation](https://github.com/msgpack/msgpack-python/blob/main/msgpack/fallback.py)


dpgeorge:
This is now available in core MicroPython via the option `MICROPY_PY_SYS_INTERN`.

(And the `sys` module is anyway no longer extensible.)

PR #583, opened 2022-12-07
urequests: close socket on delete of resp. object

koroban:
The .close() should be called if response object is not in use / can not be reached anymore. 
This should automatically call the close() method if the object is delected by the garbage collector.

### DIFF
--- a/python-ecosys/urequests/urequests.py
+++ b/python-ecosys/urequests/urequests.py
@@ -12,6 +12,9 @@ class Response:
             self.raw.close()
             self.raw = None
         self._cached = None
+        
+    def __del__(self):
+        self.close()
 
     @property
     def content(self):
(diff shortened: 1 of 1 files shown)


jimmo:
Thanks @koroban but unfortunately `__del__` is not implemented (for non-builtin types) in MicroPython. See https://docs.micropython.org/en/latest/genrst/core_language.html#special-method-del-not-implemented-for-user-defined-classes

I think what might make sense though is adding context manager methods (enter/exit) instead.

koroban:
Oh, I'm afraid I overlooked that __del__ is not available on MicroPython. 

Correct, enter and exit would definitely be a good solution as well. I'll adapt the pull request to this later today.

dpgeorge:
Closing due to inactivity.

PR #585, opened 2022-12-15
stdlib/logging/logging fix bug in isEnabledFor

FlapKap:
Fix incorrect level check in isEnabledFor in logging.py

### DIFF
--- a/python-stdlib/logging/logging.py
+++ b/python-stdlib/logging/logging.py
@@ -115,7 +115,7 @@ class Logger:
         self.level = level
 
     def isEnabledFor(self, level):
-        return level >= (self.level or _level)
+        return level >= self.level or level >= _level
 
     def log(self, level, msg, *args):
         if self.isEnabledFor(level):


python-stdlib/logging/logging.py
```diff
@@ -115,7 +115,7 @@ def setLevel(self, level):
         self.level = level
 
     def isEnabledFor(self, level):
-        return level >= (self.level or _level)
+        return level >= self.level or level >= _level
```
jimmo, line 118: @FlapKap Can you clarify what the issue was with the old code? It's saying "use the default level (INFO) if the logger doesn't have a level set".

If I understand correctly, you're changing it to "log unconditionally if level >= INFO".

Thanks

FlapKap: If only I could. I worked on a project where doing the edit i did fixed a bug where debug records where never printed. My thinking was that int >= (int or None) would never work and the rhs would be a bool which python would cast to 1. However, testing today (which i arguably should have done before submitting this pr) reveals this is not the case, and reverting my change does not produce the original bug i had. I am so confused. 

tl;dr my bad and by extension this pr is bad.

jimmo: @FlapKap  I wonder if perhaps you were seeing the issue that is fixed by this PR: https://github.com/micropython/micropython-lib/pull/589

FlapKap: Oh yeah, thats probably it. Thanks

PR #586, opened 2022-12-16
umqtt.simple and umqtt.robust improvements

zbig-t:
- Keep track of MQTT ping responses (PINGRESP) via MQTTClient.pingresp_pending attribute
- An attempt to send PINGREQ (by calling ping()) when we haven’t yet received a PINGRESP for our previous one will raise an exception
- MQTTClient.ping superclassed in mqtt.robust to make use of the  pingresp_pending attribute. Reconnect will be performed on 'No response for previous ping’ exception
- MQTTClient.connect() will always try to close the previous socket (if any) first. Otherwise the reconnect() function from mqtt.robust might fail on trying to re-instantiate the socket with OSerr=23

### DIFF
--- a/micropython/umqtt.robust/umqtt/robust.py
+++ b/micropython/umqtt.robust/umqtt/robust.py
@@ -52,3 +52,11 @@ class MQTTClient(simple.MQTTClient):
                 self.log(False, e)
             self.reconnect()
             attempts -= 1
+    
+    def ping(self):
+        while 1:
+            try:
+                return super().ping()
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks for the contribution, but I don't think we need the added complexity of this ping response check in `umqtt.simple`.

And the improvement to `.connect()` to close any existing socket was already added in 5ba4b3b39f47561d72a3d0c21236e221972f9043.

So I'll close this PR.

PR #587, opened 2022-12-18
Added timezone support for NTP

JBrO910:


micropython/net/ntptime/ntptime.py
```diff
@@ -41,10 +43,9 @@ def time():
...
 
 
-# There's currently no timezone support in MicroPython, and the RTC is set in UTC time.
 def settime():
     t = time()
     import machine
 
     tm = utime.gmtime(t)
-    machine.RTC().datetime((tm[0], tm[1], tm[2], tm[6] + 1, tm[3], tm[4], tm[5], 0))
+    machine.RTC().datetime((tm[0], tm[1], tm[2], tm[6] + 1, tm[3] + timezone, tm[4], tm[5], 0))
```
andrewleech, line 51: Unfortunately, timezone support is far more complicated to add than this appears.

For instance, with this change, if `timezone` is set to +11 and the current UTC hour (`tm[3]`) is 16 (8 pm), this will now try to set the rtc to an hour setting of 25 - this isn't going to work :-(

The rollover < 0 and > 23 would need to be carefully managed. Which then flows through to possible rollovers of day / date as well.

MrEbbinghaus:
Same issue as in: https://github.com/micropython/micropython-lib/pull/715
See my comment: https://github.com/micropython/micropython-lib/pull/715#issuecomment-1701200071

projectgus:
Thanks for submitting this PR. Everyone seems to agree MicroPython should solve this problem, but the best place is probably not in the `ntptime` module (as NTP itself doesn't know about timezones, and numeric Unix timestamps are generally always in UTC.). 

PR #588, opened 2022-12-19
Add "Content-Type: application/octet-stream" to header

trungdq88:
for when sending binary data

example usage on my ESP32-CAM module:

```python
# capture and upload
import camera
import urequests

camera.init(0, format=camera.JPEG, fb_location=camera.PSRAM)
buf = camera.capture()
response = urequests.post("https://.............", data=buf)
print(response.text)
response.close()
```


dpgeorge:
Commit 1f019f90eaef146960988319166076e12e689e81 made it possible to specify your own headers.

PR #589, opened 2022-12-19
logging: Fall back to root logger level for unset child.

jimmo:
Previously a child logger just uses the global default when unset. Modified to matches the CPython behavior of using the parent's level.

Also implemented CPython's getEffectiveLevel() which provides a convenient way to implement this. In our version, we only ever have one parent (the root), so it only has to recurse one level.

Also set the default level to WARNING to match CPython.

Updated the examples to highlight the differences (but they now match).

### DIFF
--- a/python-stdlib/logging/examples/basic_example.py
+++ b/python-stdlib/logging/examples/basic_example.py
@@ -1,3 +1,7 @@
 import logging
 
-logging.warning("test")
+logging.debug("test - debug")  # ignored by default
+logging.info("test - info")  # ignored by default
+logging.warning("test - warning")
+logging.error("test - error")
+logging.critical("test - critical")
--- a/python-stdlib/logging/examples/root_logger.py
+++ b/python-stdlib/logging/examples/root_logger.py
@@ -5,3 +5,4 @@ for handler in logging.getLogger().handlers:
     handler.setFormatter(logging.Formatter("[%(levelname)s]:%(name)s:%(message)s"))
 logging.info("hello upy")
 logging.getLogger("child").info("hello 2")
+logging.getLogger("child").debug("hello 2")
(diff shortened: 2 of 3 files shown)


jimmo:
Thanks to @chrisovergaauw for reporting and @stinos for investigating.

The initial report was:

```py
>>> logging.basicConfig(level=10, format="%(asctime)s:%(levelname)s:%(name)s:%(message)s", force=True)
>>> logging.getLogger('test2').level
0
>>> logging.getLogger('test2').info('hi')
2022-12-19 10:02:10:INFO:test2:hi
>>> logging.getLogger('test2').debug('hi')
>>>
```

CPython logs on the last line, MicroPython didn't.

jimmo:
Thanks, updated.

Btw, I was wondering about the way `strftime` is used. At the top it has

```py
if hasattr(time, "strftime"):
    from time import strftime
```

then the one place it's used:

```py
    def formatTime(self, datefmt, record):
        if hasattr(time, "strftime"):
            return strftime(datefmt, time.localtime(record.ct))
        return None
```

Why is the top-level check needed? Can the call site just be updated to use `time.strftime`?

```py
    def formatTime(self, datefmt, record):
        if hasattr(time, "strftime"):
            return time.strftime(datefmt, time.localtime(record.ct))
        return None
```

Or equivalently do the trick we do elsewhere with:

```py
if hasattr(time, "strftime"):
    from time import strftime
else:
    strftime = lambda _, _: None
```

dpgeorge:
> Why is the top-level check needed?

It could be because the original implementation had time as a package with `__init__.py` etc...?  But, yeah, that can be simplified now.  I think just do the check at the call site (your first proposal).  If you do it, please do it a as a separate commit here.

jimmo:
> If you do it, please do it a as a separate commit here.

Done.

dpgeorge:
Thanks!  I also made a follow-up commit to bump the version.

PR #591, opened 2022-12-20
umqtt.robust: When reconnecting, close the previous socket.

jefvdb:
This fixes lengthy reconnection runs (on ESP8266) which at some point started throwing OSErrors (including 23), from which it didn't recover even after server was back.


zbig-t:
Hi, good observation but I've already done it in #586

jefvdb:
Will close this one then, cheers! :)

PR #592, opened 2022-12-22
python-ecosys/urequests/urequests.py: Add readinto to Response class.

chrisovergaauw:


chrisovergaauw:
Seconds after creating the PR I realised that Response.raw is not prefixed with an underscore. So it's a public property right?
If we handle it as such then my PR doesn't make much sense because we could do:
```
r = urequests.get('someURL')
r.raw.readinto
```

I'd love to get some feedback on this.

andrewleech:
Hi, it's an interesting question you ask! There's certainly nothing wrong with going straight to `.raw` as you have, however it's also not particularly obvious (to you or other users). 

Also, while it's not a private variable (no underscore as you mention) it's not exactly a "published API" nor is it a standard interface in cpython requests (add far as I can see). As such I wouldn't trust it to not be changed in future, unintentionally breaking use like this.

That being said, `.readinto()` also doesn't appear to be standard on cpython requests so adding it here would arguably need some documentation or examples etc, and adds extra code size for everyone to do something that you can already achieve via accessing `.raw`.

So long story short, I'm not really sure about the best approach here...  

Perhaps documenting `.raw` and/or adding tests or example code might be a more lightweight option?

jimmo:
> Seconds after creating the PR I realised that Response.raw is not prefixed with an underscore. So it's a public property right?

@chrisovergaauw Yes, it matches the same (public) property in the "upstream" `requests` library.

> nor is it a standard interface in cpython requests (add far as I can see). 

@andrewleech See https://requests.readthedocs.io/en/latest/api/#requests.Response.raw

Technically we should be handling the `stream` arg to `request()`.

Either way `resp.raw.readinto` seems like a safe and reasonable thing to use and rely on.

chrisovergaauw:
Alright, then its settled. Thanks for your input Jim.

PR #593, opened 2022-12-30
micropython/drivers: Add new sensor drivers.

iabdalkader:
* Add HS3003 driver.
* Add BMM150 driver.
* Add BMI270 driver.
* Refactor LSM9 driver.
* Refactor LSM6 driver.


iabdalkader:
@dpgeorge Any feedback on this 🙏🏼 ?

micropython/drivers/imu/lsm9ds1/lsm9ds1.py
```diff
@@ -58,129 +58,149 @@
...
+        self.gyro_scale = 32768 / gyro_scale
+        self.accel_scale = 32768 / accel_scale
+        self.scale_factor_magnet = 32768 / ((MAGNET_SCALE.index(magnet_scale) + 1) * 4)
+
+        # Allocate scratch buffer for efficient conversions and memread op's
+        self.scratch_int = array.array("h", [0, 0, 0])
 
     def calibrate_magnet(self, offset):
         """
         offset is a magnet vecor that will be substracted by the magnetometer
```
jimmo, line 157: s/vecor/vector/
s/substracted/subtracted/

micropython/drivers/imu/lsm9ds1/lsm9ds1.py
```diff
@@ -58,129 +58,149 @@
...
         for each measurement. It is written to the magnetometer's offset register
         """
         offset = [int(i * self.scale_factor_magnet) for i in offset]
-        mv = memoryview(self.scratch)
+        mv = memoryview(array.array("B", [0, 0, 0, 0, 0, 0]))
```
jimmo, line 161: This doesn't need a memoryview as it always either subscripts or slices the full length. Also can be `bytearray(6)` instead of array.

On second thought, could this function just do

```
self.bus.writeto_mem(self.address_magnet, _OFFSET_REG_X_M, struct.pack("<H", offset))
```

jimmo (commented):
Thanks @iabdalkader !

micropython/drivers/imu/bmi270/bmi270.py
```diff
@@ -0,0 +1,461 @@
...
+        self.__write_reg(0x41, ACCEL_SCALE.index(accel_scale))
+
+        # Set gyroscope scale and range.
+        self.gyro_scale = 32768 / gyro_scale
+        self.__write_reg(0x43, GYRO_SCALE.index(gyro_scale))
+
+        # Allocate scratch buffer and set scale.
+        self.scratch = memoryview(array.array("h", [0, 0, 0]))
+
+    def __read_reg(self, reg, size=1):
```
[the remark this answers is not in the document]
iabdalkader: Well both, because it might be supported in the future, but also was trying to match the older LSM6DSOX driver, but I've changed them all now.

micropython/drivers/imu/lsm9ds1/lsm9ds1.py
iabdalkader: > On second thought, could this function just do

Probably yes (you mean `<HHH` ?), but would have to import struct and this function is not used from within the driver.

micropython/drivers/imu/bmi270/bmi270.py
```diff
@@ -0,0 +1,461 @@
...
+    def read_magnet(self):
+        """Returns magnetometer vector."""
+        if self.magnet is not None:
+            return self.magnet.read_magnet()
+        return (0.0, 0.0, 0.0)
+
+
+import binascii
+
+_CONFIG_DATA = binascii.unhexlify(
```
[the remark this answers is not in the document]
iabdalkader: Yeah I noticed it's very slow to import, left there for better suggestions, thanks!

micropython/drivers/imu/bmm150/bmm150.py
```diff
@@ -0,0 +1,188 @@
...
+        self.bus.readfrom_mem_into(self.address, reg, buf)
+
+    def __write_reg(self, reg, val):
+        self.bus.writeto_mem(self.address, reg, bytes([val]))
+
+    def __compensate_x(self, raw, hall):
+        """Compensation equation ported from C driver"""
+        x = 0.0
+        if raw != _XYAXES_FLIP:
+            x0 = self.trim_xyz1 * 16384.0 / hall
```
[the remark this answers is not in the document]
iabdalkader: I'm not sure, I just ported the compensation equations from C, I'm reluctant to change/optimize anything to be honest.

iabdalkader:
@jimmo I fixed all of the review comments, I think, and I retested the drivers.

micropython/drivers/imu/lsm9ds1/lsm9ds1.py
jimmo: Yep `<HHH` sorry.
There's very little cost to `import struct`, and you'll end up with shorter code, therefore less bytecode (i.e. less RAM overall if running from .py or .mpy).
Plus you can defer the `import struct` to inside the function that uses it.

micropython/drivers/imu/bmm150/bmm150.py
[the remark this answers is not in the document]
dpgeorge: I think it's OK to leave it as-is.

But, I suggest using integers like `160` instead of floats like `160.0` because it'll be slightly smaller bytecode (.mpy code).  The calculation will still be the same.

Also, it looks like the x and y compensation functions are exactly the same formula, so might be better (smaller bytecode) to create a helper function (or can you even do `__compensate_y` = `__compensate_x`?).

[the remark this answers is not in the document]
dpgeorge: Oh, and what is the licensing of the original C driver that this was copied from?

micropython/drivers/imu/bmm150/manifest.py
```diff
@@ -0,0 +1 @@
+module("bmm150.py", opt=3)
```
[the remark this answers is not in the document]
dpgeorge: Please put the `metadata()` line first in the `manifest.py` file (so it acts like a header and description of what follows).

micropython/drivers/imu/bmm150/bmm150.py
```diff
@@ -0,0 +1,188 @@
...
+        self.trim_z3 = (trim_xy1xy2[7] << 8) | trim_xy1xy2[6]
+        self.trim_z4 = (trim_xyz_data[1] << 8) | trim_xyz_data[0]
+        self.trim_xy1 = trim_xy1xy2[9]
+        self.trim_xy2 = trim_xy1xy2[8]
+        self.trim_xyz1 = ((trim_xy1xy2[5] & 0x7F) << 8) | trim_xy1xy2[4]
+
+        # Allocate scratch buffer.
+        self.scratch = memoryview(array.array("h", [0, 0, 0, 0]))
+
+    def __read_reg(self, reg, size=1):
```
dpgeorge, line 118: Was the intention to change these method names to start with a single underscore?

micropython/drivers/imu/lsm9ds1/lsm9ds1.py
```diff
@@ -58,129 +59,149 @@
...
     def gyro_id(self):
-        return self.i2c.readfrom_mem(self.address_gyro, _WHO_AM_I, 1)
+        return self.bus.readfrom_mem(self.address_imu, _WHO_AM_I, 1)
 
     def magent_id(self):
-        return self.i2c.readfrom_mem(self.address_magnet, _WHO_AM_I, 1)
+        return self.bus.readfrom_mem(self.address_magnet, _WHO_AM_I, 1)
 
-    def magnet(self):
+    def read_magnet(self):
```
dpgeorge, line 177: You've renamed these methods to add `read_` at the start.  But there are existing methods in other drivers like `altitude()` and `pressure()` (also see below in the hs3003 driver).

For consistency they should either all start with `read_`, or none should start with `read_`.

Since we don't want to make breaking changes to existing method names, please keep these names as-is, and make sure the newly added drivers also don't start with `read_` for their methods.

micropython/drivers/imu/bmi270/bmi270.py
```diff
@@ -0,0 +1,202 @@
...
...x030\xe0o\xf3T\x040\x00.\x00.\x01\x89b\x0e\xfa/CB\x110\xfbo\xc0.\x01B\xb0_\xc1J\x00\x00mW\x00\x00w\x8e\x00\x00\xe0\xff\xff\xff\xd3\xff\xff\xff\xe5\xff\xff\xff\xee\xe1\xff\xff|\x13\x00\x00F\xe6\xff\xff\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1\x80.\x00\xc1'
```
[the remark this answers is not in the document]
iabdalkader: It still doesn't look good because the lines are not even, not sure how to make it look pretty, maybe just move this blob into a separate file ?

micropython/drivers/imu/bmm150/bmm150.py
iabdalkader: Yes missed this one.

micropython/drivers/imu/bmm150/bmm150.py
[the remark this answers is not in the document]
iabdalkader: > it looks like the x and y compensation functions are exactly the same formula

No they're not, they use different trim values.

> what is the licensing of the original C driver that this was copied from

It was ported from C reference code here with BSD-3 license.

https://github.com/boschsensortec/BMM150-Sensor-API/blob/a20641f216057f0c54de115fe81b57368e119c01/bmm150.c#L1614

micropython/drivers/imu/lsm9ds1/lsm9ds1.py
iabdalkader: > Since we don't want to make breaking changes to existing method names

The lsm6dsox driver already used `read_` so I followed that one because it's newer,  the lsm9 driver doesn't, so I changed it to match, either way there will be breaking changes.

micropython/drivers/imu/bmi270/bmi270.py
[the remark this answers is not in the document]
iabdalkader: You know what I realized I can just put the hex digits in there escaped, so it should look pretty now.

micropython/drivers/imu/bmm150/bmm150.py
```diff
@@ -0,0 +1,188 @@
...
+_TRIM_Z2_LSB = const(0x68)
+_XYAXES_FLIP = const(-4096)
+_ZHAXES_FLIP = const(-16384)
+
+
+class BMM150:
+    def __init__(
+        self,
+        bus,
+        cs_pin=None,
```
dpgeorge, line 64: I know `lsm6dsox` uses `cs_pin`, but other drivers like `ssd1360`, `nrf25l01` and `sdcard` all use `cs`, and initialising things like `machine.I2C`, `machine.SPI` and `machine.UART` all use the name of the pin (eg `scl`) without a `_pin` suffix.  So I think we should try to match that standard of using the pin name without any suffix, and just call this parameter `cs`.

The more consistent we make this stuff now, the better it will be for users in the long run.

micropython/drivers/imu/lsm9ds1/lsm9ds1.py
```diff
@@ -58,129 +59,144 @@
...
+        self.address_imu = address_imu
         self.address_magnet = address_magnet
-        # check id's of accelerometer/gyro and magnetometer
+
+        ACCEL_SCALE = (2, 16, 4, 8)
+        GYRO_SCALE = (245, 500, 2000)
+        MAGNET_SCALE = (4, 8, 12, 16)
+
+        ODR_IMU = (0, 14.9, 59.5, 119, 238, 476, 952)
+        ODR_MAGNET = (0.625, 1.25, 2.5, 5, 10, 20, 40, 80)
```
dpgeorge, line 97: Is there a reason to have these constants defined within the `__init__` method?  `_SCALE_GYRO` and `_SCALE_ACCEL` were previously constants defined at the top-level of the module, and that makes them efficient (especially in .mpy files).

Can these 5 constant tuples be moved to the top-level scope, and start with underscore to make them true constants?

micropython/drivers/imu/bmm150/bmm150.py
iabdalkader: @dpgeorge Should I change the `lsm6dsox` too, it did go through some refactoring, or just the `bmm150` ?

micropython/drivers/imu/lsm9ds1/lsm9ds1.py
iabdalkader: > Is there a reason to have these constants defined within the __init__ method? 

It's because they're only needed temporarily in the `__init__` method, after that they should be collected, this way they don't take any flash/RAM, after init, right ?

micropython/drivers/imu/bmm150/bmm150.py
dpgeorge: > Should I change the lsm6dsox too

Yes please.  Since some of its methods were renamed, may as well change the `cs` argument as well.

micropython/drivers/imu/lsm9ds1/lsm9ds1.py
dpgeorge: But the bytecode needs to construct these on the fly, and that takes up flash (or RAM if the code is not frozen) due to larger bytecode.  Storing a const tuple in the .mpy file is probably on par in terms of size of the .mpy file.

Also when floats are involved the code is much bigger that if they are just small ints.

dpgeorge: Actually I just checked this and it both cases (global or local) it's pretty efficient!  The compiler is optimising these tuples to a true constant, regardless of whether they are defined globally or in the function.

But the global case where they are defined as a `const` is a little more efficient because no local variable (or global) is made.

So, to reduce bytecode size by a little bit, please make them `_SCALE_GYRO = const(...)` etc.

iabdalkader:
Updated.

PR #594, opened 2023-01-01
urequests support for user-provided http versions

coldenate:
## About
Hi there! :)
This pull request implements a simple update to the urequests's request function. This implementation includes an argument to allow for modifcation of the socket's written http version. 

## Why this was done?
I made this change in my own fork because I had scrapped together an API in FastAPI, and this did not accept my microcontroller's HTTP 1.0 post requests. 

Please let me know if there is a reason the HTTP 1.1 version remains in the code - despite possible need for it to be http 1.0. I am new to learning about HTTP and related topics :)

Cheers!

### DIFF
--- a/python-ecosys/urequests/urequests.py
+++ b/python-ecosys/urequests/urequests.py
@@ -43,9 +43,12 @@ def request(
     auth=None,
     timeout=None,
     parse_headers=True,
+    http_version="1.0",
 ):
     redirect = None  # redirection url, None means no redirection
-    chunked_data = data and getattr(data, "__iter__", None) and not getattr(data, "__len__", None)
+    chunked_data = (
+        data and getattr(data, "__iter__", None) and not getattr(data, "__len__", None)
+    )
 
     if auth is not None:
         import ubinascii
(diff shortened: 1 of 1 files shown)


jimmo:
> Please let me know if there is a reason the HTTP 1.1 version remains in the code - despite possible need for it to be http 1.0. I am new to learning about HTTP and related topics :)

@coldenate The simple answer is that this is only the second time (to my knowledge) that someone has reported a server not supporting HTTP/1.0 (which is a surprising thing for a web server to do), whereas we have no idea how many people are using urequests with servers that _only_ support HTTP/1.0 (odd, but reasonable).

Your suggestion to make it optional is probably reasonable. I would argue that perhaps we are de-facto using HTTP/1.1 anyway (setting Host: and Transfer-Encoding:), so possibly should just make the change rather than just making the code bigger. (And also, seriously, how many web servers are there out there that don't support HTTP/1.1 ???).

I'd be interested to see if anyone else has any comments.

samveen:
> I made this change in my own fork because I had scrapped together an API in FastAPI, and this did not accept my microcontroller's HTTP 1.0 post requests.

@coldenate Nathan, would you confirm that the server implementation is correct? In my rather limited experience, I've found that in most cases of clients failing to handle HTTP/1.0 responses correctly is badly coded servers (at least my implementations of servers had that issue, which has lead to #623 ).

coldenate:
@samveen It most likely is my poorly coded server 🥲, but I will look further into it. I was using FastAPI to make the server.

PR #595, opened 2023-01-03
Added duration_ms to example

ivan-galinskiy:
I've added duration_ms to the example scanning code, since it's not an optional parameter. Also, I've added some more aggressive parameters so that it has a better chance of working for someone just trying it for the first time, like myself :) And I've removed the if statement for the name, since that does not cause runtime problems, and allows to show more data. Hope it's useful!


jimmo:
Thanks @ivan-galinskiy -- merged in https://github.com/micropython/micropython-lib/commit/a5ef231e7d854bab0e6db95f1676173e2281ee3b

PR #597, opened 2023-01-26
added timezone support in ntptime library

ljnath:
ntptime library doesn't support timezone while setting time via NTP server.
This PR is to address that limitation, now the `settime()` takes an optional timezone parameter as string (example asia/kolkata).

Then it uses `wordtimeapi.org` to get the time offset and sets the time.

### DIFF
--- a/micropython/net/ntptime/ntptime.py
+++ b/micropython/net/ntptime/ntptime.py
@@ -44,7 +44,7 @@ def time():
 
 def get_tz_offset(tz: str) -> int:
     tz = tz.strip()
-    if '/' not in tz or len(tz) < 3:
+    if "/" not in tz or len(tz) < 3:
         raise Exception("Unsupported timezone: {}. Example: asia/kolkata".format(tz))
 
     tz_offset = 0
@@ -60,11 +60,11 @@ def get_tz_offset(tz: str) -> int:
         response = response.decode()
 
         from json import loads
-        response = loads(response[response.find('{'):])
+        response = loads(response[response.find("{"):])
 
-        tokens = response['utc_offset'].split(':')
+        tokens = response["utc_offset"].split(":")
         tz_offset = (int(tokens[0][1:]) * 3600) + (int(tokens[1]) * 60)
-        if tokens[0][0] != '+':
+        if tokens[0][0] != "+":
             tz_offset *= -1
     except Exception:
         pass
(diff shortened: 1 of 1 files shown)


jimmo:
Thanks @ljnath -- I think this is a useful idea, but I'm not sure this belongs in `ntptime`.

Some thoughts:
- We freeze `ntptime` on all network-enabled boards, so we have to be conscious of code size. Perhaps this could be its own utility library (or even `ntptime-timezone`). Or could `settime` take the offset as an integer argument, and getting the device timezone could be handled independently.
- All boards that freeze `ntptime` also freeze urequests, so perhaps this could use `urequests.get()` instead of doing the socket handling manually.
- Are there API limits or issues with using `worldtimeapi.org` like this?

ljnath:
@jimmo , thank you for your inputs, please find my response below

- setting of time based on timezones is a much needed functionality in the core `ntptime` , therefore extracting this functionality to a separate library might not be fruitful as it will add extra overhead for user as they will have to import 2 libraries.
Alternately passing the offset to the `settime()` method means user will have to do the heavy lifting of fetching and feeding the offset. This is not a true support of timezone in the `ntptime` library. Similar technique can be used today to set the time in one's timezone, but it increases a lot of overhead for user.
Furthermore this PR increases the size of `ntptime` by `410 bytes` from `1.29 KB` to `1.70 KB`
- Added this change
- `worldtimeapi.org` limits commercial usage of their API. My intension here is to support hobbyist and non-commercial usage.

ljnath:
@jimmo , I have made the changes couple of weeks back; will you please take a look.

projectgus:
Thanks for submitting this PR. Everyone seems to agree MicroPython should solve this problem, but the best place is probably not in the `ntptime` module (as NTP itself doesn't know about timezones, and numeric Unix timestamps are generally always in UTC.). I appreciate the interest approach of using a web service to solve the (otherwise complex) question of timezone, though!

projectgus:
@ljnath If you're still keen on this idea then it could be interesting to look into implementing a module that calls `wordtimeapi.org` independently of NTP. Afterall, their web API can give you the current time accurate to within the bounds of network latency and parsing overhead and this is good enough for a lot of purposes.

(I'm not sure if micropython-lib would accept a new module that's bound to a third party service with API usage restrictions, but it's the kind of thing you could maintain standalone.)

PR #598, opened 2023-01-27
micropython/bundles: Add a bundle-networking meta-package.

jimmo:
This is designed to be a common set of packages that all boards with networking support should include.

Signed-off-by: Jim Mussared <jim.mussared@gmail.com>

### DIFF
--- a/micropython/bundles/README.md
+++ b/micropython/bundles/README.md
@@ -1,0 +1,7 @@
+These are "meta packages" designed to make it easy to provide defined
+bundles of related packages.
+
+For example, all deployments of MicroPython with networking support
+(WiFi/Ethernet) should add `require("bundle-networking")` to their
+`manifest.py` to ensure that the the standard set of networking packages
+(including HTTP requests, WebREPL, NTP, package management) are included.
--- a/micropython/bundles/bundle-networking/manifest.py
+++ b/micropython/bundles/bundle-networking/manifest.py
@@ -1,0 +1,9 @@
+metadata(
+    version="0.1.0",
+    metadata="Common networking packages for all network-capable deployments of MicroPython.",
+)
+
+require("mip")
+require("ntptime")
+require("urequests")
+require("webrepl")


micropython/bundles/README.md
```diff
@@ -0,0 +1,6 @@
+These are "meta packages" designed to make it easy to ensure that boards end
+up with a consistent set of default packages.
```
dpgeorge, line 2: This sentence (and the `metadata` in the `manifest.py`) are all very board specific.  Does it make sense to reword it to be a bit more generic, eg "to ensure that deployments"?

jimmo: Updated

iabdalkader:
Would it be possible to call it `networking-core` instead ? It's easier to remember but also in the future we could add another bundle (`networking-utils`, `networking-tools` or `networking-extras`) to include things like `mqtt`, `uping`, `uftpd` and `utelnet`, and the `-core`, `-utils`, `-tools` and `-extras` can map to other things (for example `sensors-core` etc..).

jimmo:
> Would it be possible to call it `networking-core` instead ?

I thought the `bundle-` prefix was important (to disambiguate it from other packages).

In terms of the suffix, yes I agree that there will likely be more types of networking bundles.

I guess the question is, are the other bundles extensions of the core one, or are they independent? i.e. would you install `bundle-networking-foo` without also installing `bundle-networking`? Equivalently, would `bundle-networking-foo` require() `bundle-networking` ?

If the former, then for consistency with other packages in this repo, the "base" one should have no suffix. If it's the latter then we should rename this one.  @dpgeorge ?

iabdalkader:
> I thought the bundle- prefix was important (to disambiguate it from other packages).

I don't think it's that important to append `bundle-` mainly because a suffix such as `-tools`, `-utils` or `-extras` (and maybe  `-core` too) already implies that there's more than one module in the package, but also maybe `mip` already lists the contents of a package/bundle, or can do that in the future.

> are the other bundles extensions of the core one, or are they independent

Well I think they can have dependencies in some cases, for example a bundle that groups compression modules, may depend on another bundle which provides common optimized compression algorithms, but more commonly a bundle usually groups multiple standalone (but related) modules.

I think we should avoid `-core` for this package, because nothing will depend on it, and maybe use `-utils` instead ? So `-core` ` (or `-base` if you prefer) should be reserved for bundles that will be dependencies of other bundles. Following the compression example, if we have `bzip`, `gzip` and `lzw` packages, a bundle combining them would be called `compression-utils`, if it depends on multiple packages that provide common code, those can be grouped in `compression-core/base`.

dpgeorge:
I think the prefix of `bundle-` is important:
1. so that we don't run out of names / have clashing namespaces; eg `bundle-storage` could get you all storage related modules, but then we might also want an actual `storage` module
2. because you can't actually import the bundle by name, eg if you install `bundle-networking` you don't expect to be able to do `import bundle-networking` (but if it was called simply `networking` then you might want to try and `import networking`

> more commonly a bundle usually groups multiple standalone (but related) modules

Yes, that's the way to think about them, that they group together related modules.

But by definition the related modules are *standalone*; otherwise these related modules would already have dependencies upon one another (well, mip requires urequests but that's kind of an implementation detail).

> So `-core` (or `-base` if you prefer) should be reserved for bundles that will be dependencies of other bundles

By my argument above, no bundle will be a (strict) dependency of another, because they are simply collections of existing libraries.  The bundles exist for convenience only, and not because there is some kind of hierarchy.  Indeed, bundles could have some overlap with each other in what they include.

That said, there could be `bundle-python-stdlib` that gets all stdlib libraries, `bundle-python-ecosys` and `bundle-micropython` (combining their respective dirs), and then `bundle-all` that requires those previous three.

iabdalkader:
> I think the prefix of bundle- is important:

My main argument is not whether there should be a prefix/suffix or not, it's that `-bundle` will be redundant, because it's looking like we're going to need a suffix, so we'll end up with `bundle-networking-tools`, `bundle-networking-servers` and `bundle-networking-utils` etc... but then suffix here already implies that it's a bundle and distinguishes it from `networking`

> By my argument above, no bundle will be a (strict) dependency of another

I tend to agree, after rethinking, so in that case `-core` should be free to use for this package.

dpgeorge:
> it's looking like we're going to need a suffix

Not always.  Eg `bundle-storage` or `bundle-lora` etc.

> the suffix here already implies that it's a bundle and distinguishes it from `networking`

That's also not true: existing packages use a suffix to indicate a subset or extension of a library, eg `aioble-central`.

We need the `bundle-` prefix to indicate that it's not something you can import (as opposed to, eg, `aioble-central` for which you can import `aioble`).

iabdalkader:
> Not always. Eg `bundle-storage` or `bundle-lora` etc.

But what would a `bundle-storage` include for example ? If it groups core storage modules, it can use `storage-base` or `storage-core`, then if we need an extra or a tools bundle for storage, we can use `storage-extras` Vs. `bundle-storage-extras`. The same goes for any bundle, the files that it groups will very likely fall into one the following suffixes: `-base`, `-core`, `-utils`, `-tools`, `-extras`, `-drivers` , while extensions will likely use something different related to the package (ex the `-central` extension you pointed out).

You do have a point about `bundle-` marking the bundle as not importable, is this going to be run-time checked ? If so the above suffixes can also indicate that it's not importable, if it's not going to be checked, then it's just a hint for the user.

dpgeorge:
> But what would a `bundle-storage` include for example ?

It could include for example (some of these are made-up packages for illustration): `upysh`, `sdcard`, `spiflash`, `fsinspect`.

A module named `storage-extras` is already "reserved" by our naming scheme to mean that, if you install it, you'll have additional things in the `storage` module.

Note that installing a package ABC (via `mip install ABC` for example) doesn't always mean you can do `import ABC`.  But our current naming scheme does have the following rules:
- for a package named X without any hyphens you can do `import X`
- a package named X-Y is an extension of the X package (see the last paragraph of the README in this repo for details)

If we don't prefix these new meta packages with `bundle-` (or something else similar) then it breaks that last rule (or at the very least makes it confusing).

> The same goes for any bundle, the files that it groups will very likely fall into one the following suffixes: `-base`, `-core`, `-utils`, `-tools`, `-extras`, `-drivers` , while extensions will likely use something different related to the package

I don't think we can easily distinguish between bundle suffixes (like base, core, utils) and non-bundle suffixes.  Eg `bundle-lora-drivers` can be a collection of LoRa drivers, which don't necessarily have anything to do with the `lora` package.  While `lora-drivers` would install a set of plug-in drivers (eg `sx1280`) for the `lora` module to use (and they must be installed inside the `lora/` subdir so they are truly part of the `lora` import).

> You do have a point about `bundle-` marking the bundle as not importable, is this going to be run-time checked ?

Well, they are implicitly not importable because they have a hyphen in the name; you can't write `import abc-def` in Python (it's a syntax error).  (Note that you can do `__import__("abc-def")`.)

When you install `bundle-xyz` it does not create a package called `bundle-xyz`.  And it does not create a directory `bundle-xyz` either.  It's just a meta name which depends on other packages, and those other packages are the things that get installed and can be imported.

iabdalkader:
> If we don't prefix these new meta packages with bundle- (or something else similar) then it breaks that last rule (or at the very least makes it confusing).

Yes I see the existing pattern, but I guess what I'm missing is what's the point of distinguishing between bundles and extension packages ? They both can't be imported anyway, and when installing them they will either add extensions to an "importable" package, or install a number of related modules that can be imported. So for example `lora-drivers` will add drivers to `lora` package, and `lora-utils` will install some standalone modules that can be imported, why is it important for the user to know any of this ? Assuming `mip` shows a description of the package and/or contents.

iabdalkader:
I guess if it's to avoid any confusion for contributors, and for someone freezing a package, then yeah probably makes sense. But wouldn't `meta-` make more sense than `bundle-` ? Since the bundle doesn't actually include any actual modules. Anyway, don't let me drag this any longer, please fee free to move forward with this.

dpgeorge:
> Yes I see the existing pattern, but I guess what I'm missing is what's the point of distinguishing between bundles and extension packages ? They both can't be imported anyway, and when installing them they will either add extensions to an "importable" package, or install a number of related modules that can be imported.

No part of a bundle's name can be imported, eg if you have `bundle-networking` then you can't do `import networking`.

But for extension packages you can.  For `lora-drivers` you can do `import lora`.

What we are trying to distinguish here is exactly between these two cases.  As you put it:
- bundles "install a number of related modules that can be imported"
- extension packages "add extensions to an "importable" package"

And here's a reason to distinguish them: say you are using a subset of `aioble` (eg for a peripheral) and you like it's functionality and want to know if there is more it can do.  All you have to do is look for packages named `aioble-xyz` and they will be part of the full `aioble` package.

> I guess if it's to avoid any confusion for contributors, and for someone freezing a package, then yeah probably makes sense

This repo and `mip` are intended for use by a whole range of users (developers, contributors, end users etc) so it needs to be as clear as possible.

> But wouldn't `meta-` make more sense than `bundle-` ? Since the bundle doesn't actually include any actual modules.

Yeah I did wonder if "meta" would be better.  But I think "bundle" conveys the idea better, because it does install a set of modules.  If you saw the name `meta-networking` you might think it's just some description and doesn't actually do anything if you install it.  Whereas if you saw the name `bundle-networking` you'd probably guess it installed a bundle of things if you installed it.

PR #599, opened 2023-01-27
ntptime: Use NTPv4

macifell:
Update the NTP client request to version 4 from version 3.

### DIFF
--- a/micropython/net/ntptime/ntptime.py
+++ b/micropython/net/ntptime/ntptime.py
@@ -17,7 +17,7 @@ timeout = 1
 
 def time():
     NTP_QUERY = bytearray(48)
-    NTP_QUERY[0] = 0x1B
+    NTP_QUERY[0] = 0x23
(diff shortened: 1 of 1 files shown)


jimmo:
Thanks @macifell -- This seems fine, but I'm just curious what prompted you to send the PR. Are there NTP servers that no longer accept NTPv3?

macifell:
Hi @jimmo , It just looked weird to me seeing NTPv3 packets on my network as NTPv4 has been out for over 10 years. NTPv4 has some minor improvements for IPv6 compatibility. Having the version already set to 4 might make it easier for future developers to enhance this client. I doubt any NTP servers no longer accept NTPv3, but I guess that will be the case at some point.

dpgeorge (approved):
NTPv4 is backwards compatible with NTPv3 so this should be fine.

I tested on the unix port of MicroPython and it works.

dpgeorge:
Rebased and merged in ee1b12d07372c08364a0cde6e3c1e89683712dd3

PR #600, opened 2023-01-28
urequests: Fixed UTF-8 support for request data

### LINKED ISSUE #251: urequests: miscalculation of Content-Length when dealling with unicode data
code in urequests ver 0.5.1:
`s.write(b"Content-Length: %d\r\n" % len(data))`

"len" calcs a single unicode as 1 length.
to correct it **encode the data before calcs the length**
as:  `s.write(b"Content-Length: %d\r\n" % len(data.encode('')))`
plz fix it.

asid21:
UTF-8 support for request data was added, it will be able to handle special characters correctly.

### DIFF
--- a/python-ecosys/urequests/urequests.py
+++ b/python-ecosys/urequests/urequests.py
@@ -110,7 +110,7 @@ def request(
             if chunked_data:
                 s.write(b"Transfer-Encoding: chunked\r\n")
             else:
-                s.write(b"Content-Length: %d\r\n" % len(data))
+                s.write(b"Content-Length: %d\r\n" % len(data.encode()))
         s.write(b"Connection: close\r\n\r\n")
         if data:
             if chunked_data:


python-ecosys/urequests/urequests.py
```diff
@@ -110,7 +110,7 @@ def request(
             if chunked_data:
                 s.write(b"Transfer-Encoding: chunked\r\n")
             else:
-                s.write(b"Content-Length: %d\r\n" % len(data))
+                s.write(b"Content-Length: %d\r\n" % len(data.encode()))
```
jimmo, line 113: This looks like a tiny change but what it means is that we need to unconditionally make a complete copy of the payload in RAM, only to then compute its length. This will significantly impact the memory overhead of using urequests for posting data.

I note that in the upstream "requests" libraries, it is invalid to pass a string (not bytes) as the `data` argument if that string contains non-`Latin-1` characters.

e.g.

```py
>>> requests.get('http://localhost:8000/', data='⛵')
...
UnicodeEncodeError: 'latin-1' codec can't encode character '\u26f5' in position 0: Body ('⛵') is not valid Latin-1. Use body.encode('utf-8') if you want to send it encoded in UTF-8.
```

I'm not sure if MicroPython has an easy way to detect this case though (without doing something like `len(foo.encode()) != len(foo)` which obviously has the same problem.

But it's definitely not ideal what it does at the moment where it sends the incorrect Content-Length header....

foxt:
+1.
Reporting temperature data to Home Assistant requires the °C to be included.

A quick work around I found to work is just 
```py
urequests.post(
    ...
   data=json.dumps({ ... }).encode()
)
```

python-ecosys/urequests/urequests.py
bixb922: This also affects json payloads with multibyte utf-8 characters. The json gets truncated when sent, because the length is off.

Does encoding a string really allocate new data? I thought encoding just casts the raw string data as bytes?

See also issue #784 for aiohttp in micropython-lib

dpgeorge: @projectgus is correct, doing `bytes(str, "utf-8")` does not copy the payload.  That's great because we only need to convert it if it is a str.

This PR doesn't do it correctly though, because if the user passes in `bytes`/`bytearray`/`memoryview` for `data` then it'll try to encode it -- which will fail.

dpgeorge:
Closing because this breaks the case where the user passes in non-str objects for `data`.

PR #601, opened 2023-01-28
Fix AttributeError when unittests tries to handle exceptions

cb4b1fd915:
without path: 
```python
test__clock_draw__minimal_dimentions__correct_fbuf_values (__main__.SegClockTestCase) ...Traceback (most recent call last):
  File "<stdin>", line 35, in <module>
  File "unittest/__init__.py", line 462, in main
  File "unittest/__init__.py", line 269, in run
  File "unittest/__init__.py", line 254, in run
  File "unittest/__init__.py", line 430, in _run_suite
  File "unittest/__init__.py", line 405, in run_one
AttributeError: 'module' object has no attribute 'exc_info'
```

with patch: 
```python 
test__clock_draw__minimal_dimentions__correct_fbuf_values (__main__.SegClockTestCase) ... ERROR

FAIL: test__clock_draw__minimal_dimentions__correct_fbuf_values <class 'SegClockTestCase'>
  File "unittest.py", line 390, in run_one
  File "<stdin>", line 26, in test__clock_draw__minimal_dimentions__correct_fbuf_values
AttributeError: 'SegClockTestCase' object has no attribute 'assertEquals'
```

### DIFF
--- a/python-stdlib/unittest/unittest/__init__.py
+++ b/python-stdlib/unittest/unittest/__init__.py
@@ -402,9 +402,7 @@ def _run_suite(c, test_result: TestResult, suite_name=""):
             test_result.skipped.append((name, c, reason))
         except Exception as ex:
             _handle_test_exception(
-                current_test=(name, c),
-                test_result=test_result,
-                exc_info=(type(exc), exc, None)
+                current_test=(name, c), test_result=test_result, exc_info=(type(exc), exc, None)
             )
             # Uncomment to investigate failure in detail
             # raise
--- a/python-stdlib/unittest/unittest/__init__.py
+++ b/python-stdlib/unittest/unittest/__init__.py
@@ -332,12 +332,12 @@ class TestResult:
         return self
 
 
-def _capture_exc(exc, traceback):
+def _capture_exc(exc, exc_traceback):
     buf = io.StringIO()
     if hasattr(sys, "print_exception"):
         sys.print_exception(exc, buf)
     elif traceback is not None:
-        traceback.print_exception(None, exc, traceback, file=buf)
+        traceback.print_exception(None, exc, exc_traceback, file=buf)
     return buf.getvalue()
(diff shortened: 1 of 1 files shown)


jimmo:
Thanks @cb4b1fd915 

While I was reviewing this, I noticed a related issue -- the traceback returned as the third element of this tuple ends up in `_capture_exc` which due to https://github.com/micropython/micropython-lib/pull/488/files#diff-9fcb56c6c2b92e63164d1c7110a716dff84de752c52d4bc1c5574016eb21e13fR334 will not work due to the `traceback` argument shadowing the `traceback` global.

I think we should make this PR just do:

```
_handle_test_exception(
                current_test=(name, c), test_result=test_result, exc_info=(type(ex), ex, None,)
            )
```

and never use `sys.exc_info` (it's deprecated anyway and almost never enabled in MicroPython).

Then to fix `_capture_exc` I see two options:
- Make `unittest` require the `traceback` module from micropython-lib (it's tiny). Then we can unconditionally `import traceback` and re-factor `_capture_exc` back into `_handle_test_exception` using `traceback.print_exeception`. Remove `import sys` entirely.
- Just fix the argument name to `_capture_exc` (i.e. rename `traceback` to `tb`).

@andrewleech What do you think? I think I like the first option better

cb4b1fd915:
@jimmo, thank you for the review. 
Decided to follow the second option (just fix the argument name) for now.
 It looks like it's enough to call sys.print_exception in [_capture_exc](https://github.com/micropython/micropython-lib/blob/master/python-stdlib/unittest/unittest/__init__.py#L338).

jimmo:
Thanks @cb4b1fd915 !

Merged in 863a018b892f0bf901fa09844aa3428d5a5a00c1  (added test, fixed crash -- `exc` should be `ex`, bumped package version, squashed, fixed commit message).

Issue #546, opened 2022-10-09
urequests.get ignores content-length header


GM-Script-Writer-62850:
When connecting a server running this php sample
```
<?php
set_time_limit(0);
ob_start();
header('Content-type: text/plain');
header("Content-Encoding: none");
echo "OK";
header('Content-Length: '.ob_get_length());
header('Connection: close');
ob_end_flush();
@ob_flush();
flush();
if(session_id()) session_write_close();
sleep(10)
die("Client should never see this")
?>
```
if you view this page in a web browser or in curl you will only get `OK`, however `urequests.get` ignores the content length and waits for sleep to finish and get the die message before returning content, urequests.head has the same issue except you can't get the status_code

i managed to fix this, but my fix does nothing when i `import` it even though it works if i copy/paste the library inside my script
this was how i fixed it
```
    @property
    def content(self):
        if self._cached is None:
            try:
                if(self.headers["Content-Length"]):
                    self._cached = self.raw.read(int(self.headers["Content-Length"]))
                else:
                    self._cached = self.raw.read()
            finally:
                self.raw.close()
                self.raw = None
        return self._cached
```

jimmo:
I think this probably makes sense to fix!



> i managed to fix this, but my fix does nothing when i `import` it even though it works if i copy/paste the library inside my script
> this was how i fixed it

How are you copying your fixed script to the device. Which directory are you putting it in. Note that `sys.path` by default is `["", ".frozen", "/lib"]` so if you put it in "/lib" the frozen copy will take precedence. Place the file in the root directory instead (alongside your main.py if you want it to override the frozen one).

GM-Script-Writer-62850:
i was putting it in `/lib/urequests.py`
i got it to work by using `/lib/urequest.py`

i guess `urequests` in included on the stock pico firmware, thought it was not but i must have made a typo in the lib name when i tried to import it the 1st time

Now i can just use 
`import urequest as urequests`
or just use
`import urequest`
instead of 
import urequests as urequest

bmidgley:
The problem arises for me with a service that sends a payload without a final newline. But I couldn't use the workaround. I ended up using a terrible hack where I read(1) until I notice the ending } in the payload.

jonnor:
Is this still a problem in the latest versions of requests library?

GM-Script-Writer-62850:
Yes this is present in 
[...]

PR #602, opened 2023-01-30
Use content length to avoid http hang

bmidgley:
response.content() is hanging on responses that don't end with a final newline.

Issue is in https://github.com/micropython/micropython-lib/issues/546

Here's my actual workaround while waiting for the real fix... reach into the impl and read one character at a time. Stop when reaching "}"

```
    response = urequests.get(url)
    response._cached = response.raw.read(1)
    while True:
        response._cached += response.raw.read(1)
        if response._cached[-1] == 125: break
    json = response.json()

### DIFF
--- a/python-ecosys/urequests/urequests.py
+++ b/python-ecosys/urequests/urequests.py
@@ -2,10 +2,11 @@ import usocket
 
 
 class Response:
-    def __init__(self, f):
+    def __init__(self, f, length):
         self.raw = f
         self.encoding = "utf-8"
         self._cached = None
+        self.length = length
 
     def close(self):
         if self.raw:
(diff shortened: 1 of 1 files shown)


bmidgley:
Is there another way to avoid this hang?

andrewleech:
I don't think the .read() function should be waiting for a newline, only .readline() is supposed to work this way.
.read() should just provide all data currently available.
The existing implementation can work for downloading binary files too, which definitely doesn't depend on a trailing newline.

That being said it does seem sensible to me to honour the content length header if it's provided.

dpgeorge:
Closing in favour of #1124, which implements HTTP/1.1.

PR #603, opened 2023-02-01
github/workflows: Add workflow to parse each manifest file.

dpgeorge:


dpgeorge:
The new CI takes about 25s.  Most of that time is building mpy-cross (around 17s).

jimmo:
Looks great, thank you!!

dpgeorge:
Building mpy-cross with -O0 cuts it's build time down to about 10s, and doesn't slow down the actual use of mpy-cross when it runs.  So worth doing.  (Also an interesting test to run mpy-cross with -O0...)

PR #609, opened 2023-02-08
micropython/aiorepl: Add paste mode and raw repl.

Carglglz:
This adds a paste mode to  `aiorepl` and a "raw REPL" mode to enable compatibility with MicroPython development tools .e.g `pyboard.py`/`mpremote` and other tools that use the raw REPL mode to communicate with the device.

### DIFF
--- a/micropython/aiorepl/aiorepl.py
+++ b/micropython/aiorepl/aiorepl.py
@@ -84,6 +84,45 @@ __exec_task = asyncio.create_task(__code())
         print(f"{type(err).__name__}: {err}")
 
 
+async def paste_mode(s):
+    sys.stdout.write("\npaste mode; Ctrl-C to cancel, Ctrl-D to finish\n=== ")
+    buff_paste = ""
+    while True:
+        b = await s.read(1)
+        c = ord(b)
+
+        if c == 0x03:
+            sys.stdout.write("\n")
+            return
+        elif c == 0x04:
+            sys.stdout.write("\n=== \n")
+            return buff_paste
+        else:
+            buff_paste += b
+            sys.stdout.write(b)
+
+
+async def raw_repl(s, g):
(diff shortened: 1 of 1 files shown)


dpgeorge:
Sorry I didn't see this!  It got superseded by #756.

Carglglz:
Don't worry #756 had more features than this one, also I think that it's worth exploring the idea of developing an
 `_aiorepl`  C module to get it working as close as normal REPL/raw REPL (e.g.  autocompletion, `;` chained commands, etc)

andrewleech:
I also just apologize, I didn't know of this work you'd done before I started mine, if nothing else I'd prefer to have built in top of yours rather than writing from scratch!

While working on this space I did look at just exposing the C code rather than re-writing in python, but the structure of how repl works was making this surprisingly difficult!

Maybe it is still worth breaking out smaller chunks to use, but that does at the flash overhead of said python interface code that everyone gets regardless of whether they use aiorepl.
I'd be keen to figure out a way to make the built in repl able to be re-run in async. 

This modded micropython apparently has some async support in the repl, but it's all mashed into the one mega-commit, I haven't looked into exactly what's been done with the repl:
https://github.com/gneverov/micropythonrt

andrewleech:
I also just apologize, I didn't know of this work you'd done before I started mine, if nothing else I'd prefer to have built in top of yours rather than writing from scratch!

While working on this space I did look at just exposing the C code rather than re-writing in python, but the structure of how repl works was making this surprisingly difficult!

Maybe it is still worth breaking out smaller chunks to use, but that does at the flash overhead of said python interface code that everyone gets regardless of whether they use aiorepl.
I'd be keen to figure out a way to make the built in repl able to be re-run in async. 

This modded micropython apparently has some async support in the repl, but it's all mashed into the one mega-commit, I haven't looked into exactly what's been done with the repl:
https://github.com/gneverov/micropythonrt

PR #613, opened 2023-02-16
python-stdlib/unittest: Added subtest usage examples

pi-gjd:
Added examples of how the unittest.subTest context manager can be used, and the features it supports.
---

### DIFF
--- a/python-stdlib/unittest/examples/example_subtest.py
+++ b/python-stdlib/unittest/examples/example_subtest.py
@@ -1,0 +1,241 @@
+# MIT license; Copyright (c) <year>, Planet Innovation
+# SPDX-License-Identifier: MIT
+# 436 Elgar Road, Box Hill, 3128, VIC, Australia
+# Phone: +61 3 9945 7510
+#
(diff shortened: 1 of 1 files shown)


andrewleech:
At PI we've had a few internal requests for guidance on how to write parametised unit tests without pytest. `unittest.subTest` is the "official" way to do this, however it isn't neccesarily obvious how to use it as such. These examples that Gavin have put together should give some really useful guidance to users on how to use this feature effectively!

stinos:
I like the idea of having an example for this, but I think it would be nicer to have it more minimal: it takes quite a bit of reading code now to clearly see what is going on. Could start by just testing an existing function? Also perhaps the license should be like all other licenses used (see MIT section in the LICENSE file in the root of the repository)?

pi-gjd:
Both great suggestions @stinos. I removed the copyright header, and I converted the test for `selection_sort()` to test `sorted()` instead, since for this example, we're not really worried about the implementation details (or even algorithm) of the sorting method.

PR #614, opened 2023-02-16
aioble: add short name support to scan results

FlantasticDan:
I'm working on using BLE to connect to a Blackmagic Camera and have discovered that in it's advertising packets it will never send it's Complete Local Name (denoted by `0x09`), opting to only send it's Shortened Local Name (denoted by `0x08`).  I've made it so a `ScanResult` will prioritize returning the Complete Local Name, then attempt to fall back to the shortened variant before ultimately returning an empty string.

**Bluetooth Specification Reference**
The [Assigned Numbers Document](https://www.bluetooth.com/specifications/assigned-numbers/) identifies the fields/values for both names as Common Data Types in Section 2.3.  The variants of Local Name are described in Part A, Section 1.2 of the [Core Specification Supplement](https://www.bluetooth.com/specifications/specs/core-specification-supplement-11/).

### DIFF
--- a/micropython/bluetooth/aioble-central/manifest.py
+++ b/micropython/bluetooth/aioble-central/manifest.py
@@ -1,4 +1,4 @@
-metadata(version="0.2.0")
+metadata(version="0.2.1")
 
 require("aioble-core")
 
--- a/micropython/bluetooth/aioble/aioble/central.py
+++ b/micropython/bluetooth/aioble/aioble/central.py
@@ -33,6 +33,7 @@ _SCAN_RSP = const(4)
 
 _ADV_TYPE_FLAGS = const(0x01)
 _ADV_TYPE_NAME = const(0x09)
+_ADV_TYPE_SHORT_NAME = const(0x08)
 _ADV_TYPE_UUID16_INCOMPLETE = const(0x2)
 _ADV_TYPE_UUID16_COMPLETE = const(0x3)
 _ADV_TYPE_UUID32_INCOMPLETE = const(0x4)
@@ -187,9 +188,9 @@ class ScanResult:
                     yield payload[i + 2 : i + payload[i] + 1]
                 i += 1 + payload[i]
 
-    # Returns the value of the advertised name, otherwise empty string.
+    # Returns the value of the complete (or shortened) advertised name, if available.
     def name(self):
-        for n in self._decode_field(_ADV_TYPE_NAME):
+        for n in self._decode_field(_ADV_TYPE_NAME, _ADV_TYPE_SHORT_NAME):
             return str(n, "utf-8") if n else ""
 
     # Generator that enumerates the service UUIDs that are advertised.
--- a/micropython/bluetooth/aioble/manifest.py
+++ b/micropython/bluetooth/aioble/manifest.py
@@ -3,7 +3,7 @@
 # code. This allows (for development purposes) all the files to live in the
 # one directory.
 
-metadata(version="0.2.0")
+metadata(version="0.2.1")
 
 # Default installation gives you everything. Install the individual
 # components (or a combination of them) if you want a more minimal install.


micropython/bluetooth/aioble/aioble/central.py
```diff
@@ -187,10 +188,19 @@ def _decode_field(self, *adv_type):
                     yield payload[i + 2 : i + payload[i] + 1]
                 i += 1 + payload[i]
 
-    # Returns the value of the advertised name, otherwise empty string.
+    # Returns the value of the complete (or shortened) advertised name, otherwise empty string.
     def name(self):
         for n in self._decode_field(_ADV_TYPE_NAME):
-            return str(n, "utf-8") if n else ""
+            if n:
```
jimmo, line 194: Thanks @FlantasticDan, I've never seen a device using the short name which is why I never added it, but this looks useful.

The `_decode_field` function allows multiple field IDs to be specified, but they are returned in payload order. I don't see any reason not to re-order the loops in `_decode_field` though to make them in `*arg` order instead though (this wouldn't affect the `services` function because it's returning all services anyway). So that would make this a much smaller change to support short name.

Do you think it's likely that a device would include both short and long names though? Perhaps short in the adv payload and long in the scan response? This would mean that a device would return the short name in the first iteration of the scanner loop, and the long name subsequently. So should we make this a separate function?

FWIW, one thing I've been considering is not returning results for active scans for devices that support scan responses until the scan response arrives (this is what Android and iOS do). This "feature" of aioble to return the partial result is a source of much confusion. My intention had been that an app shouldn't care -- it just keeps scanning until it matches what it needs.

FlantasticDan: @jimmo appreciate the review!  I didn't realize `_decode_field` could take multiple field IDs.  That given with the below excerpt of the spec leads me to believe we should attempt to decode both long and short name and just return whatever comes back first.

> The Local Name data type shall be the same as, or a shortened version of, the local name assigned to the device. The Local Name data type value indicates if the name is complete or shortened. If the name is shortened, the complete name can be read using the remote name request procedure over BR/EDR or by reading the device name characteristic after the connection has been established using GATT.
> 
> A shortened name shall only contain contiguous characters from the beginning of the full name. For example, if the device name is ‘BT_Device_Name’ then the shortened name could be ‘BT_Device’ or ‘BT_Dev’.
>
> __*from the* Supplement to the Bluetooth Core Specification v11, Part A, Section 1.2.1__

I think the spec implies that devices will only send either the complete **or** short name because it explicitly describes the procedure to get a long name when given a short name.  Combine that with the fact that the short name must be contiguous characters from the complete name I think even if in the edge case where the user was returned a short name when a long name was available, it would still be clear which device they were communicating with and thus a non-issue.  What do you think?

I'm happy to modify this PR in this direction.

---

> FWIW, one thing I've been considering is not returning results for active scans for devices that support scan responses until the scan response arrives (this is what Android and iOS do). This "feature" of aioble to return the partial result is a source of much confusion. My intention had been that an app shouldn't care -- it just keeps scanning until it matches what it needs.

Correct me if I'm wrong but I think this explains why an active scan will return the same device (as identified by it's address) and sometimes be able to populate the name field and other times return an empty string...  I've been using the following excerpt to I suspect accommodate the "feature" you described:
```python
devices = []
addrs = set()
async with aioble.scan(duration_ms=5000, interval_us=30000, window_us=30000, active=True) as scanner:
    async for result in scanner:
        if result.device.addr_type == 0 and result.name() != '' and result.device.addr_hex() not in addrs:
            devices.append((result.name().rstrip('\x00'), result.device.addr_hex()))
            addrs.add(result.device.addr_hex())
print(devices)
```

Not that you asked but if you are considering changing this behavior perhaps a non breaking way to do it would be to add a kwarg like `include_partial=True` so that users that want the more iOS/Android-y experience can opt in to it.

jimmo: > I'm happy to modify this PR in this direction.

That would be good thanks! (So I think that means this PR just changes that one line to `for n in self._decode_field(_ADV_TYPE_NAME, _ADV_TYPE_SHORT_NAME):` ?

> Correct me if I'm wrong but I think this explains why an active scan will return the same device (as identified by it's address) and sometimes be able to populate the name field and other times return an empty string...

Yep, that's exactly right. My theory was that if you need the full information (only available once the scan result arrives) then you would just ignore the first result. And if the first result was enough, then that's great you can connect immediately.

> Not that you asked but if you are considering changing this behavior perhaps a non breaking way to do it would be to add a kwarg like include_partial=True so that users that want the more iOS/Android-y experience can opt in to it.

Yes, perhaps. (But maybe I didn't make it clear -- Android and iOS suppress the results _until_ the scan response arrives. So in order to make it backwards-compatible, the Android/iOS behavior would be opt-in).

micropython/bluetooth/aioble/aioble/central.py
```diff
@@ -187,10 +188,11 @@ def _decode_field(self, *adv_type):
                     yield payload[i + 2 : i + payload[i] + 1]
                 i += 1 + payload[i]
 
-    # Returns the value of the advertised name, otherwise empty string.
+    # Returns the value of the complete (or shortened) advertised name, otherwise empty string.
     def name(self):
-        for n in self._decode_field(_ADV_TYPE_NAME):
+        for n in self._decode_field(_ADV_TYPE_NAME, _ADV_TYPE_SHORT_NAME):
             return str(n, "utf-8") if n else ""
+        return ""
```
jimmo, line 195: Hrmm. I realise this is now matching the comment. But I think this should actually be the old behavior (the comment was incorrect). i.e. Returning None means there was no name field, as opposed to empty string which would be the (somewhat odd) case of there being an empty name set.

Sorry! Could you please fix the comment instead.

FlantasticDan: That's definitely more logical, `None` literally meaning there isn't a name available.  I'm on it.

micropython/bluetooth/aioble-central/manifest.py
```diff
@@ -1,4 +1,4 @@
-metadata(version="0.2.0")
+metadata(version="0.3.0")
```
dpgeorge, line 1: @jimmo not sure this is the correct way to bump the version for aioble.  Maybe it should be done to all components?  (And maybe 0.2.1?, it could be considered a bug fix, it doesn't add anything to the API.)

dpgeorge: Let's go for 0.2.1, since it doesn't change any API.

And the super package `aioble` also needs its version number bumped, due to its dependency on `aioble-central`.

PR #617, opened 2023-02-23
ci: Support publishing package & index files to GitHub Pages.

projectgus:
Opt-in feature to make it easier for folks to test packages that are still in development, open in Pull Requests, or even in independent forks.

*EDIT: Instructions are now in the PR contents*

### DIFF
--- a/.github/workflows/build_packages.yml
+++ b/.github/workflows/build_packages.yml
@@ -2,6 +2,9 @@ name: Build all packages
 
 on: [push, pull_request]
 
+env:
+  PACKAGE_INDEX_PATH: /tmp/micropython-lib-deploy
+
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -17,3 +20,8 @@ jobs:
     - name: Publish packages for branch
       if: vars.MICROPY_PUBLISH_MIP_INDEX && github.event_name == 'push' && ! github.event.deleted
       run: source tools/ci.sh && ci_push_package_index
+    - name: Upload packages as artifact
+      uses: actions/upload-artifact@v3
+      with:
+        name: packages-${{ github.sha }}
+        path: ${{ env.PACKAGE_INDEX_PATH }}
--- a/.github/workflows/build_packages.yml
+++ b/.github/workflows/build_packages.yml
@@ -14,3 +14,6 @@ jobs:
       run: source tools/ci.sh && ci_build_packages_check_manifest
     - name: Compile package index
       run: source tools/ci.sh && ci_build_packages_compile_index
+    - name: Publish packages for branch
+      if: vars.MICROPY_PUBLISH_MIP_INDEX && github.event_name == 'push' && ! github.event.deleted
+      run: source tools/ci.sh && ci_push_package_index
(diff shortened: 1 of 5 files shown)


projectgus:
@jimmo @dpgeorge Please take a look. Polishing this from the initial PoC took a little longer than expected (as always) but it seems pretty robust now.

The only downside is that over time the gh-pages commit log will get long on anyone with a very active fork. The actual size in git should stay quite small, as almost all objects will be the same. If you're worried about this then the solution will be rewriting git history, but I recommend leaving it for now and waiting to see if it becomes a problem (not convinced that it will) - or can simply remind folks in the documentation that they can delete the gh-pages branch from their fork whenever they want.

This will need proper documentation added. Do you have a preference between `CONTRIBUTING.md` in micropython-lib, or `docs/reference/packages.rst` in micropython?

projectgus:
> What exactly lead you to implement this? Would you use it yourself?

@dpgeorge Right, sorry, some background missing here!

Someone on Discord asked about how they could install the LoRa packages from my LoRa PR, and the best solution was "copy the files to your board manually". Which is possible but it's not easy (especially with relatively complex packages like that), and the files will be .py not .mpy.

So the motivation was to find an easy way for anyone to make the changes in their micropython-lib PR accessible for others to install and test.

I was originally thinking of doing it in the GitHub Actions `pull_request` event handler and pushing to a single micropython S3 bucket or VPS web server somewhere, but this way the infrastructure is all part of GitHub and the responsibility rests with the fork's owner.

projectgus:
BTW, can see how this looks in my fork here: https://github.com/projectgus/micropython-lib/commits/gh-pages

jimmo:
Thanks @projectgus -- this is really neat, I look forward to using it!

I don't have any feedback on the actual implementation or code, looks great!

> and create a variable named MIP_INDEX with value true (or any "truthy" value).

When I was reviewing this, I noticed that ci.sh has a comment "if the MIP_INDEX repository variable is set". It would be good to update this comment to include this extra bit about what the variable should actually be set to.

I'm not 100% sure about naming the variable `MIP_INDEX`... in other places in mip, the "index" is the URL to the index (i.e. I would expect to set this variable to a URL). Maybe `BUILD_MIP_INDEX` (which is clearly a yes/no thing)?

dpgeorge:
> Maybe `BUILD_MIP_INDEX` (which is clearly a yes/no thing)?

How about prefixing it with `MICROPY_` to create our own namespace that doesn't clash with other things (potentially in the future etc)?

projectgus:
@dpgeorge @jimmo Sorry for all the commit noise, was fiddling with the documentation. Have renamed the opt-in variable to `MICROPY_PUBLISH_MIP_INDEX`.

I also snuck in another commit that attaches the packages as "Artifacts" on all build workflows. These can be downloaded for a short time afterwards (but only as a zip file, so not suitable for `mip` installs.) I thought this might generally be useful.

projectgus:
@Josverl if you don't mind me asking out of the blue, is anything in this PR of interest to you, or would it be with any modifications?  (I saw on Discord & https://github.com/micropython/micropython/pull/10846 that you're doing different but very interesting things with CI deployment of packages, so I thought you might have some valuable perspective.)

dpgeorge:
From my point of view, this looks good to merge!

dpgeorge:
I force pushed to update the commit messages (minor changes) and merged.

Thanks @projectgus !

PR #618, opened 2023-02-23
Add bare-minimum manifest.py files to bisect, json, keyword, and types.

KLimit:
These four modules (bisect, json, keyword, and types) seem to have been missing manifest files; I'm unsureif it's just been an oversight or if there is a reason for it.

I didn't have much success finding what the versions of these modules may be, nor did I put too much effort into finding the details commented in some other manifests (original author, other metadata), but I think that info should also be covered in the code itself. As in the summary, this is the bare minimum necessary to be able to use a custom manifest that requires `keyword`, or example. In fact, I've tested this by successfully compiling Pi Pico firmware with a custom manifest that requires `keyword`.

### DIFF
--- a/python-stdlib/bisect/manifest.py
+++ b/python-stdlib/bisect/manifest.py
@@ -1,0 +1,1 @@
+module("bisect.py")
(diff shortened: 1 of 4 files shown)


KLimit:
Missed the contributing guide; will resubmit with preferred commit summary and some metadata.

PR #619, opened 2023-02-23
bisect, json, keyword, types: Add manifest.py.

KLimit:
Add manifest files for bisect, json, keyword, and types since they appeared to have been overlooked.

I ran into trouble when trying to compile Pi Pico firmware with `keyword` as a frozen module after which I found several other modules without manifests. Adding the manifest for `keyword` resolved my compilation issue, so I have assumed that the other manifests are functional as well.
It wasn't readily apparent what the versions of each of these modules are, so I chose not to add anything about that to their metadata. (It's also a little unclear to me what the version is meant to represent -- version of the Python stdlib that it's ported from or the version of the port itself or is it up to the author?).

### DIFF
--- a/python-stdlib/bisect/manifest.py
+++ b/python-stdlib/bisect/manifest.py
@@ -1,0 +1,3 @@
+metadata(description="Port of bisect for micropython")
+
+module("bisect.py")
--- a/python-stdlib/json/manifest.py
+++ b/python-stdlib/json/manifest.py
@@ -1,0 +1,3 @@
+metadata(description="Port of json module for micropython")
+
+package("json")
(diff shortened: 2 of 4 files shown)


dpgeorge:
Thanks for the contribution, and sorry that this PR got lost/overlooked.

What you did in this PR is correct.

But it was done also by someone else and merged in commits c2b44ea83b1cb930e91ddc1f81ffb426846acdd0, 0f768c9af71b69b5f4e81f0cab8aaa403c820c2e, b95deb31e51c017fd62971fb7c2a050a70be1ba3, 028a369f903fa395e8ed57dcaf8fba98f684cecc

PR #624, opened 2023-03-02
Senml cbor2 modules

iabdalkader:


dpgeorge:
Thanks.  I also updated the senml examples to use `time` instead of `utime`.

dpgeorge:
@iabdalkader when running the senml examples (to test this PR) I noticed that `supported_data_types.py` doesn't work, it complains of undefined `base64`.

iabdalkader:
> noticed that `supported_data_types.py` doesn't work, it complains of undefined `base64`.

Fixed here: https://github.com/micropython/micropython-lib/pull/625

PR #630, opened 2023-03-16
unittest-discover: Print results when no tests are found/run.

andrewleech:
Currently if no tests are found when running `unittest discover` then nothing at all is written to stdout, leading you to think it's not working at all.

cpython unittest does display a `0 tests run` sort of output in this scenareo, this PR ensures our package does the same.

### DIFF
--- a/python-stdlib/unittest-discover/unittest/__main__.py
+++ b/python-stdlib/unittest-discover/unittest/__main__.py
@@ -111,7 +111,6 @@ def _dirname_filename_no_ext(path):
 
 
 def discover_main():
-    failures = 0
     runner = TestRunner()
 
     if len(sys.argv) == 1 or (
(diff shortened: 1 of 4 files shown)


python-stdlib/unittest-discover/unittest/__main__.py
```diff
@@ -121,22 +121,27 @@ def discover_main():
...
-                result = _run_test_module(runner, test_spec)
+                res = _run_test_module(runner, test_spec)
 
-            failures += result.failuresNum or result.errorsNum
+            result += res
+
+    if not result.testsRun:
+        # If tests are run their results are already printed.
+        # Ensure an output is still given if no tests are found.
+        result.printResults()
```
dpgeorge, line 141: Instead of exposing `printResults()` as a separate function (which isn't in CPython...), would it be possible (and simpler) to just call `runner.run(TestSuite())` here?

python-stdlib/unittest-discover/unittest/__main__.py
```diff
@@ -121,22 +121,27 @@ def discover_main():
     ):
         # No args, or `python -m unittest discover ...`.
         result = _discover(runner)
-        failures += result.failuresNum or result.errorsNum
```
dpgeorge, line 124: Line 114 above with `failures = 0` can go.

python-stdlib/unittest-discover/unittest/__main__.py
andrewleech: Nice idea, thanks!
I've made this change and yes it does appear to work correctly.

python-stdlib/unittest-discover/unittest/__main__.py
andrewleech: whoops, I thought I cleaned up the failures variable, good catch.

dpgeorge:
Thanks for updating!

PR #635, opened 2023-03-25
Feature/ntp timezone support

marcy05:
Introduced a lightweight time zone handling to allow hour shift.

### DIFF
--- a/micropython/net/ntptime/ntptime.py
+++ b/micropython/net/ntptime/ntptime.py
@@ -41,10 +41,11 @@ def time():
     return val - NTP_DELTA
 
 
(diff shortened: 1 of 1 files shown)


MrEbbinghaus:
Same issue as in: https://github.com/micropython/micropython-lib/pull/715
See my comment: https://github.com/micropython/micropython-lib/pull/715#issuecomment-1701200071

projectgus:
Thanks for submitting this PR. Everyone seems to agree MicroPython should solve this problem, but the best place is probably not in the `ntptime` module (as NTP itself doesn't know about timezones, and numeric Unix timestamps are generally always in UTC.). 

PR #636, opened 2023-03-31
senml: Allow publishing to PyPI as micropython-senml. 

jimmo:
This PR adds a mechansim to allow micropython-lib packages to be published to PyPI for use by CPython.

The idea is that there are some packages (in the micropython directory) that do not have CPython counterparts -- essentially our version is the "original" or "upstream" for these packages. As opposed to a package in the python-stdlib or python-ecosys directories, where the micropython-lib is our version of those packages -- if you were using CPython you'd use the "real" one instead.

This adds a tool that can parse a package's manifest and generate a Python project directory (with pyproject.toml) suitable for building and uploading to PyPI. (Note this requires https://github.com/micropython/micropython/pull/11162 to be merged first).

This is enabled for the micropython/senml package which has no counterpart in the PyPI. Note that this depends on cbor2, which does exist in PyPI, so the CPython version, micropython-senml will depend on the "real" version instead.

### DIFF
--- a/micropython/senml/manifest.py
+++ b/micropython/senml/manifest.py
@@ -1,4 +1,8 @@
-metadata(version="0.1.0")
+metadata(
+    description="SenML serialisation for MicroPython.",
+    version="0.1.0",
+    pypi_publish="micropython-senml",
+)
 
 require("cbor2")
 
--- a/tools/makepyproject.py
+++ b/tools/makepyproject.py
@@ -1,0 +1,215 @@
+#!/usr/bin/env python3
+#
+# This file is part of the MicroPython project, http://micropython.org/
+#
+# The MIT License (MIT)
+#
+# Copyright (c) 2023 Jim Mussared
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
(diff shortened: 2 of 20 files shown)


micropython/senml/manifest.py
```diff
@@ -1,5 +1,7 @@
-metadata(version="0.1.0")
+metadata(
+    version="0.1.0", description="SenML serialisation for MicroPython.", pypi="micropython-senml"
```
dpgeorge, line 2: Would it be prettier to add a `,` at the end to force Black to put these all on separate lines?

micropython/senml/manifest.py
```diff
@@ -1,5 +1,7 @@
...
+)
 
-require("cbor2")
+require("cbor2", pypi="cbor2")
```
dpgeorge, line 5: This kwarg would need to be repeated each time this package is required.... maybe it would be better to instead have the target package define what it's PyPI package name is, eg in the metadata?  That's more self-documenting ("this package is intended to implement a subset of ...") and has less duplication.

OTOH, maybe it is better to have this dependency written out here, because maybe the package requiring it needs more control over what it depends on.  Eg there may be two packages in micropython-lib:
1. senml, requires cbor2 (micropython-lib) and cbor2 (PyPI)
2. foobar, requires cbor2 (micropython-lib) and foobor2 (PyPI)

??

micropython/senml/manifest.py
jimmo: Done.

micropython/senml/manifest.py
jimmo: > OTOH, maybe it is better to have this dependency written out here, because maybe the package requiring it needs more control over what it depends on.

Yes, this is exactly the intention. I started with the former, with exactly the same idea "it would be neat if cbor2 knew it was actually based on PyPI/cbor2", but I think in general you want the dependee to specify its dependencies explicitly. (This also includes the ability to pin a version).

dpgeorge: OK, very good.

I guess the fact that something is in `python-ecosys` with a given name, means it's supposed to be based on a PyPI package.

Maybe one day we can add a field to `metadata` to make this explicit.  Or just include the PyPI name in the description.

brainelectronics:
This seems to implement kind of inverse "functionality" of https://github.com/brainelectronics/micropython-package-validation which creates a `package.json` from a `setup.py`. Excited to see this!

tools/makepyproject.py
```diff
@@ -0,0 +1,215 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+# This script makes a CPython-compatible package from a micropython-lib package
+# with a pyproject.toml that can be built (via hatch) and deployed to PyPI.
+# Requires that the project sets the pypi= kwarg in its metadata().
```
dpgeorge, line 29: should be "`pypi=` or `pypi_publish=` kwarg"?

dpgeorge: actually should just say `pypi_publish=`

jimmo: Done.

dpgeorge:
The CI here now passes after the changes to `micropython` were merged.

I also updated the manifest changes to put `description` before `version`.  That's the convention that seems to be more widely used elsewhere in this repo.

PR #641, opened 2023-04-07
aioespnow: New module to provide asyncio support for espnow module.

glenn20:
The `aioespnow` module provides asyncio support for the `espnow` module on ESP32 and ESP8266 ports.

`aioespnow` is a supplementary module which extends the micropython `espnow` module to provide
`asyncio` support.

- Asyncio support is available on all ESP32 targets as well as those ESP8266 boards which include the `uasyncio` module (ie. ESP8266 devices with at least 2MB flash storage).

See the documentation in `micropython-lib/micropython/aioespnow/README.md`.

### DIFF
--- a/micropython/aioespnow/README.md
+++ b/micropython/aioespnow/README.md
@@ -1,0 +1,91 @@
+# `aioespnow`
+
+A supplementary module which extends the micropython `espnow` module to provide
+`asyncio` support.
+
+- Asyncio support is available on all ESP32 targets as well as those ESP8266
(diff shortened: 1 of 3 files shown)


dpgeorge:
Thanks!  Merged.

mattytrentini:
Very happy to see this merged, thanks folks! 

A quick note though; there's no mention of aioespnow in the [espnow documentation](https://docs.micropython.org/en/latest/library/espnow.html); should there be?

PR #643, opened 2023-04-11
aioble: Fix descriptor flag handling.

jimmo:
Follow up to https://github.com/micropython/micropython/pull/11239

- Removes the workaround for https://github.com/micropython/micropython/issues/6864.

- Sets the default flags for discovered descriptors to be WRITE, so that d.write() will implicitly set response=True.


dpgeorge:
Thanks, merged.

PR #644, opened 2023-04-13
Mpy lib/pasco2 sensor module

NikhitaR-IFX:
I have added only required things to support PoC and leave scope for further development. 

Also, there is no error handling really here and depends completely on assertions raised by micropython I2C module.


jaenrig-ifx (approved):
Very clean and lean 👍 Just to be consistent with the rest of the convention -> getCO2Value() could be get_co2_value()?

IFX-Anusha (approved):
Looks good!

PR #646, opened 2023-04-19
types: Add manifest file.

andrewleech:
Curiously the stdlib/types module missed getting a manifest file in the big migration from [setup.py](https://github.com/micropython/micropython-lib/commit/ecef7a506ccbf31bbbdc0369d17f1d09543e0442#diff-d230118eba5a373e8601695ff1152c9c938687d82cf42cf214ac14ec78432082) to [manifest.py](https://github.com/micropython/micropython-lib/commit/ce66e701) perhaps because it was missing the metadata.txt and only had [setup.py](https://github.com/micropython/micropython-lib/blob/a18d49cda7a5d615298e8bb65fb24208036befff/python-stdlib/types/setup.py)

### DIFF
--- a/python-stdlib/bisect/manifest.py
+++ b/python-stdlib/bisect/manifest.py
@@ -1,0 +1,3 @@
+metadata(version="0.5")
+
+module("bisect.py")
--- a/python-stdlib/json/manifest.py
+++ b/python-stdlib/json/manifest.py
@@ -1,0 +1,3 @@
+metadata(version="0.1")
+
+package("json")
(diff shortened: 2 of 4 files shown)


andrewleech:
On a related note... 
```
/micropython-lib/python-stdlib $ for F in *; do if [ ! -f $F/manifest.py ]; then ls $F/manifest.py;fi;done
ls: cannot access 'README.md/manifest.py': Not a directory
ls: cannot access 'bisect/manifest.py': No such file or directory
ls: cannot access 'json/manifest.py': No such file or directory
ls: cannot access 'keyword/manifest.py': No such file or directory
```

dpgeorge:
Thanks!

PR #654, opened 2023-05-06
espflash: Reduce the memory fragmentation during flash write.

robert-hh:
espflash used to implicitely allocate four times per write the ~4k buffer for the data. That caused boards with smaller memory to fail due to memory fragmentation. This change reduces the allocation to once per block.


dpgeorge:
IMO this was a reasonable improvement, always good to reduce memory fragmentation.  But I was waiting for the mimxrt WiFi integration to be finished before fully reviewing it, so I never got to it.

PR #655, opened 2023-05-07
Micropython/upysh: imporve ux

Igorello74:
Micropython/upysh: Improve UX by allowing calling commands interactively;
make internal classes and functions not importable (put _ before their names),
so that `from upysh import *` won't import unnecessary entities.

Interactively means you don't have to type tons of brackets and quotes
(something alike what you expect from a normal shell).

### DIFF
--- a/micropython/upysh/upysh.py
+++ b/micropython/upysh/upysh.py
@@ -148,9 +148,7 @@ pwd = _Pwd()
 ls = _Ls()
 clear = _Clear()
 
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks for the contribution, but this is not the right direction for this module.  They are fine as they are (and this PR is a big breaking change for most commands).

I suggest instead to maintain your own version of upysh.

PR #657, opened 2023-05-09
micropython/ucontextlib-async: Add asynccontextmanager in extension package.

andrewleech:
I had a need for `async with` support so ported `@asynccontextmanager ` from cpython, with just a basic amount of simplifying of the needed functions from https://github.com/python/cpython/blob/3.11/Lib/contextlib.py

Arguably this `micropython/ucontextlib` really could be "squashed" into stdlib/contextlib now as it really is a subset of cpython functionality, even if the existing stdlib/contextlib was moved into a separate extension package.  I didn't do this so far though as it would presuably break a bunch of existing stuff expecting to find ucontextlib/

I know there's a plan to add automatic handing / searching of u vs u-less modules in micropython, was this expected to be able to extend to micropython-lib ones? Or do we leave a placeholder package called ucontextlib that requires / imports contextlib ?


colin-nolan:
I also encountered the need for `async with` and wanted to use `@asynccontextmanager` so I gave your change a try. 

The issue I encountered was that it uses `anext`, which is a built-in in CPython 3.10+  but it does not seem to be implemented in MicroPython yet: https://docs.micropython.org/en/latest/differences/python_310.html. 

Am I missing something?

andrewleech:
Oh good point... I wonder... my need for this evaporated so I didn't end up using it in project, however I thought I had the unit tests passing before I pushed this so now I'm not sure

PR #659, opened 2023-05-14
micropython/utarfile: Support creating tar files.

dpwe:
Currently utarfile supports a small subset of the cpython tarfile functionality to read items from an existing tar archive.
This adds a `mode='w'` option to the constructor, and a `tarfile.add('filename')` method to support creating tar files. 
Only regular files and directories are supported.


jimmo:
Thanks @dpwe this looks useful.

Something we need to consider is code size -- someone who just wants to read tar files shouldn't need to pay the extra flash cost. So to that end we've started splitting packages into base and extension packages. This might be a good candidate for that.

The approach is:
- rename `utarfile.py` to `utarfile/__init__.py` (i.e. utarfile is now a package rather than a module).
- update `manifest.py` to use `package("utarfile")` rather than `module("utarfile.py")`
- add a new library package "utarfile-create" with `utarfile/create.py` and a manifest that has `require("utarfile")` and `package("utarfile")`. (Other suggestions welcome for the name.. maybe `utarfile-add`?).
- update utarfile/__init__.py to try to do `from . import create` and handle the ImportError gracefully, and add the extra methods to the `TarFile` class (either with thunks or by conditionally making `TarFile` inherit from a class provided by `create.py` (lots of options here).

Conceptually the idea is that installing the `utarfile-create` library package adds an extra optional file to the `utarfile` package that extends the behavior. You can see this in use in `collections`, `unittest`, `mip`, `aioble` (they all actually do this in slightly different ways which suit each of the packages).

(Note: I'm using both meanings of "package" in the above... see https://github.com/micropython/micropython-lib/#notes-on-terminology)

Unrelated to this PR... just a note while we're looking at `tarfile`. I think we should rename `utarfile` to `tarfile` and move it to `python-stdlib`. Need to verify that it matches the CPython API though. See https://github.com/micropython/micropython-lib/pull/540 for thoughts on renaming of urequests (and backwards compatibility).

dpwe:
Thanks.  I converted to a package and split the tarfile writing into a separate utarfile-write extension that layers on top of utarfile, with minimal additions to the original module.

Because the new methods are top-level members of the TarFile class, I had to monkeypatch them in in utarfile/__init__.py, let me know if this looks OK.

I also added "append" mode support since it was a small change.

dpwe:
Yes, looks great.  That's a nice size reduction!

I was a bit uncomfortable with the ambiguity in how we represent if the TarInfo is a directory (between .type and .mode).  Your unification is great.

Do you want me to do anything more?  Should I update my PR to use your version?

dpwe:
I copied your changes into this branch and verified that everything still works.

jimmo:
Thanks @dpwe, looks good.

One thing that is worth pointing out is that as a result of it going from being a module to a package is that if someone already has it installed and then updates to the latest version, mip will not remove the old utarfile.py from the filesystem. However, packages of the same name always take precedence over .py files (it importer does dir, .py, .mpy in that order), so this should not be an issue.

It will also need a version bump to manifest.py, but I will do that when I merge this (tomorrow).

micropython/utarfile-write/utarfile/write.py
```diff
@@ -0,0 +1,135 @@
...
+        self.f.write(_NUL * (_BLOCKSIZE * 2))
+        self.offset += _BLOCKSIZE * 2
+        remainder = self.offset % _RECORDSIZE
+        if remainder:
+            self.f.write(_NUL * (_RECORDSIZE - remainder))
+    self.f.close()
+
+
+# Inject extra functionality into TarFile.
+from . import TarFile, TarInfo
```
dpgeorge, line 130: This is a circular import, but I think it works because the package (`__init__`) is already loaded by this point.

Would it be smaller code to do the following within the `TarFile` class:
```
try:
    from .write import _open_write, addfile, add, close
except:
    pass
```
??

micropython/utarfile-write/utarfile/write.py
```diff
@@ -0,0 +1,135 @@
...
+    "mode": (uctypes.ARRAY | 100, uctypes.UINT8 | 7),
+    "uid": (uctypes.ARRAY | 108, uctypes.UINT8 | 7),
+    "gid": (uctypes.ARRAY | 116, uctypes.UINT8 | 7),
+    "size": (uctypes.ARRAY | 124, uctypes.UINT8 | 12),
+    "mtime": (uctypes.ARRAY | 136, uctypes.UINT8 | 12),
+    "chksum": (uctypes.ARRAY | 148, uctypes.UINT8 | 8),
+    "typeflag": (uctypes.ARRAY | 156, uctypes.UINT8 | 1),
+}
+
+# Following https://github.com/python/cpython/blob/3.11/Lib/tarfile.py
```
dpgeorge, line 23: Does the code here copy anything verbatim from that Python file?  Just need to be careful of licensing.

dpgeorge:
Thanks @dpwe , this is a really nice addition!

jimmo:
Squashed and merged in 7128d423c2e7c0309ac17a1e6ba873b909b24fcc (including suggestions from @dpgeorge). Thanks again @dpwe !

PR #660, opened 2023-05-18
stdlib/logging: Add full support for logging exception tracebacks.

andrewleech:
I noticed recently that the current `log.exception()` function doesn't always print the tracebacks of the exception.

The current handling requires your build to be compiled with `MICROPY_PY_SYS_EXC_INFO` which is not always on by default.

This PR allows you to pass an exception object in as the `exc_info` kwarg (cpython allows this) so the following works regarless of the above build setting.

``` python
try:
    run_func()
except Exception as ex:
  log.exception("run_func failed", exc_info=ex)
```

Separately to that, currently even when `sys.exc_info()` is enabled, it's only printing the traceback to `_stream` = `sys.stderr` - not to the configured logging handlers. This means for instance if you've got a file log handler, it misses out on the tracebacks.
This PR also addresses this.

### DIFF
--- a/python-stdlib/logging/logging.py
+++ b/python-stdlib/logging/logging.py
@@ -1,5 +1,5 @@
 from micropython import const
-
+import io
 import sys
 import time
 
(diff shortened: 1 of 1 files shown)


dpgeorge:
This looks good, thanks.  And it's CPython compatible (CPython also allows `exc_info` to be a 3-tuple, but we don't need to support that).

PR #661, opened 2023-05-19
storage/sdcard: Clean sdcard test txt files to free card spaces.

luaNewer:


dpgeorge:
Thanks, but I think it's fair enough to keep the files there after the test.  In case you want to inspect them.

PR #663, opened 2023-05-19
python-stdlib/pathlib: Add __rtruediv__ magic method to pathlib.Path

BrianPugh:
Now works due to https://github.com/micropython/micropython/pull/10844#issuecomment-1554135154.

Code is still valid with older versions of micropython, just the actual rtruediv operations won't work (they previously still didn't work, so no change in functionality).


BrianPugh:
bump

dpgeorge:
Thanks for the fix, and the bump!  Looks good.

Issue #665, opened 2023-05-20
iperf3 module on pico-w: KeyError: end_time


graeme-winter:
```
>>> iperf3.client("192.168.1.249")
CLIENT MODE: TCP sending
Connecting to ('192.168.1.249', 5201)
Interval           Transfer     Bitrate
  0.00-1.00   sec   655 KBytes  5.36 Mbits/sec
  1.00-2.00   sec   639 KBytes  5.23 Mbits/sec
  2.00-3.00   sec   656 KBytes  5.38 Mbits/sec
  3.00-4.00   sec   669 KBytes  5.48 Mbits/sec
  4.00-5.00   sec   669 KBytes  5.48 Mbits/sec
  5.00-6.00   sec   666 KBytes  5.45 Mbits/sec
  6.00-7.00   sec   686 KBytes  5.62 Mbits/sec
  7.00-8.00   sec   642 KBytes  5.25 Mbits/sec
  8.00-9.00   sec   680 KBytes  5.56 Mbits/sec
  9.00-10.00  sec   555 KBytes  4.56 Mbits/sec
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
  0.00-10.00  sec  6.36 MBytes  5.34 Mbits/sec  sender
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "iperf3.py", line 491, in client
  File "iperf3.py", line 150, in report_receiver
KeyError: end_time
```

graeme-winter:
Suspect that this is just a race condition, where the packets are being handled less quickly than a usual `iperf3` implementation & so one arrives out of order

graeme-winter:
(all this is assuming the `iperf3.mpy` from `mip install` is derived from the code in `micropython-lib`)

jimmo:
> Suspect that this is just a race condition, where the packets are being handled less quickly than a usual iperf3 implementation & so one arrives out of order

Unless I'm missing something silly, that line is computing `dt` which is never actually used. It would fail on the next line anyway though.

I'm curious though -- I've used iperf3 a fair bit (both with micropython and other iperf3 implementations at the other end) and never seen this error. And weirdly `stats["streams"][0]` is clearly still a dictionary, so clearly it received something from the other end, it just happened to be missing these keys. @graeme-winter Would it be possible to add a try/except KeyError and print `stats["streams"][0]` to see what the other end sent?

Also, what is the other iperf3 implementation that it's talking to?

[...]

PR #666, opened 2023-05-21
iperf3: macOS server compatibility

graeme-winter:
Result message does not appear to encode start, end time, so work around. Probably better fix incoming once I better understand.
    
Fixes micropython/micropython-lib#665

### DIFF
--- a/python-ecosys/iperf3/iperf3.py
+++ b/python-ecosys/iperf3/iperf3.py
@@ -147,10 +147,10 @@ class Stats:
 
     def report_receiver(self, stats):
         st = stats["streams"][0]
-        dt = st["end_time"] - st["start_time"]
+        dt = ticks_diff(self.t3, self.t0)
         self.print_line(
-            st["start_time"],
-            st["end_time"],
+            st.get("start_time", 0.0),
+            st.get("end_time", dt * 1e-6),
             st["bytes"],
             st["packets"],
             st["errors"],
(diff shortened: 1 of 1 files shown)


Josverl:
Confirm that this also solves similar errors on Windows 11.
There are a few other errors in iperf3.py that prevent it from running against a recent windows iperf3 server that ill raise a seperate issue and PR for

Josverl (approved):
Confirmed working OK on rp2 iperf3.py client to Windows iperf3server

Josverl (approved):
Confirmed working OK on rp2 iperf3.py client to Windows iperf3server

python-ecosys/iperf3/iperf3.py
```diff
@@ -147,10 +147,10 @@ def stop(self):
 
     def report_receiver(self, stats):
         st = stats["streams"][0]
-        dt = st["end_time"] - st["start_time"]
+        dt = ticks_diff(self.t3, self.t0)
```
dpgeorge, line 150: I think this line needs a comment, to mention that servers pre-3.2 don't send these values and so in this case the dt is computed as our (local) dt.

graeme-winter: Done in 2056dac

dpgeorge:
Thanks for updating the comment.  Rebased and squashed and merged in 01ab7ba6e27e0b7350e9f6da5b7096fa0a7b0d70

PR #672, opened 2023-05-30
unix-ffi: Isolate packages from python-stdlib version of `os`

jefferyto:
Currently, installing `email.message` (or any package that requires it, i.e. `email.feedparser`, `email.parser`, `http.client`) or `glob` causes the python-stdlib version of `os` to be installed instead of the unix-ffi version.

This adds unix-ffi versions of `os-path` and `uu` so that the unix-ffi version of `os` is installed in these cases.

### DIFF
--- a/unix-ffi/email.message/manifest.py
+++ b/unix-ffi/email.message/manifest.py
@@ -1,7 +1,7 @@
 metadata(version="0.5.3")
 
 require("re", unix_ffi=True)
-require("uu")
+require("uu", unix_ffi=True)
 require("base64")
 require("binascii")
 require("email.utils", unix_ffi=True)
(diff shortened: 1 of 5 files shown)


jimmo:
Thanks for the PR @jefferyto. Sorry the unix-ffi thing is a bit complicated, but I think this is a very reasonable way to solve this.

jefferyto:
A more general solution could be to persist the `unix_ffi=True` flag into nested `require()` calls, that would allow a top-level unix-ffi package to include non-unix-ffi packages and have the stream of requires flow back into unix-ffi when necessary, but I'm not sure how difficult this would be or if there is any interest in this kind of functionality.

dpgeorge:
See https://github.com/micropython/micropython/pull/13620 for an alternative solution.

dpgeorge:
> A more general solution could be to persist the `unix_ffi=True` flag into nested `require()` calls

That's effectively what the above-linked PR does.

PR #674, opened 2023-06-01
hashlib: Provide all built-in hash functions.

jimmo:
If a port had md5 enabled, then it would no longer be available.

Retain the existing logic of using the built-in preferentially.


KipCrossing (approved):
LGTM - although there is no easy way for me to test this (that I know of yet)

PR #678, opened 2023-06-12
errno: Add missing error definitions.

ned-pcs:
Several error codes issued by lwip and other libraries were missing from errno/errno.py. This PR adds these codes.


dpgeorge:
How did you come across these "missing" error codes, what was the actual error that you received?  What port/system are you using?

MicroPython never uses `ENOMSG` so I'm curious why you see that.

You shouldn't need to use this `errno.py` file in most cases.  The built-in `errno` module in the MicroPython firmware should be enough.

ned-pcs:
OK, closing. I've been using the esp32 port and have seen some of these error codes.

PR #685, opened 2023-06-17
utarfile: sanitize file size bytearray

### LINKED ISSUE #684: utarfile: invalid syntax for integer with base 8
I am using [OTA library](https://github.com/mkomon/uota) that I have tweaked to get my firmware from Github Release Assets, which are packed into Tar file (`tar -czf source.tar.gz *`) using [Github Workflow](https://github.com/AuHau/projector-cabinet/blob/master/.github/workflows/release_please.yaml#L43). 

The new firmware is downloaded to the flash correctly but when it is about to be extracted, it fails with error: `invalid syntax for integer with base 8` coming from [extracting the file's size in `utarfile`](https://github.com/micropython/micropython-lib/blob/7128d423c2e7c0309ac17a1e6ba873b909b24fcc/micropython/utarfile/utarfile/__init__.py#L119). 

I have printed the byte arrays of the `h.size` and got: `bytearray(b'00000002264 \x00')`.
Then I tried to repack the `tar` on my Mac and reuploaded that to the device and suddenly all works, but noticed that the printed byte array of size is: `bytearray(b'00000002264 ')`. Eq. `\x00` is missing.

In the Github Actions it is packed on Ubuntu machine. Maybe there should be some sanitation of the bytearray?

AuHau:
Fixes #684 

Not sure if it is correct way how to handle it, but it works for me 😅

### DIFF
--- a/micropython/utarfile/utarfile/__init__.py
+++ b/micropython/utarfile/utarfile/__init__.py
@@ -116,7 +116,7 @@ class TarFile:
         # Update the offset once we're sure it's not the run-out.
         self.offset += len(buf)
         d = TarInfo(str(h.name, "utf-8").rstrip("\0"))
-        d.size = int(bytes(h.size), 8)
+        d.size = int(bytes(h.size.replace(b"\x00", b"")), 8)
         self.subf = d.subf = FileSection(self.f, d.size, _roundup(d.size, _BLOCKSIZE))
         self.offset += _roundup(d.size, _BLOCKSIZE)
         return d


massimosala (changes requested):
To be consistent with h.name sanitizing, I suggest you to remove the .replace() and use

> h.size.rstrip("\0")

dpgeorge:
Fixed instead by 0a5b63559421a791083cdfadfdba2633f34f4d65

PR #687, opened 2023-06-19
aiorepl: Document passing an object to be in scope

sandyscott:
Add documentation to help if you want to access the properties & methods of a specific object at the `aiorepl` prompt. It has to be passed with some kind of wrapper, instead of directly using `__dict__` as one might expect.


jimmo:
Thanks @sandyscott -- I updated the README in c48b17dd17d3104d397fa779e41a94098f32734f

PR #693, opened 2023-06-24
aiorepl: Fix usage of const function

mishal:
aiorepl: Fix usage of const function

### DIFF
--- a/micropython/aiorepl/aiorepl.py
+++ b/micropython/aiorepl/aiorepl.py
@@ -15,7 +15,7 @@ _RE_GLOBAL = re.compile("^([a-zA-Z0-9_]+) ?=[^=]")
 _RE_ASSIGN = re.compile("[^=]=[^=]")
 
 # Command hist (One reserved slot for the current command).
-_HISTORY_LIMIT = const(5 + 1)
+_HISTORY_LIMIT = micropython.const(5 + 1)
 
 
 async def execute(code, g, s):


jimmo:
@mishal `const` is handled specially in MicroPython's parser. In fact, by adding `micropython.const` it prevents this from applying the `const` optimisation, it _must_ be written as just plain `const`.

```
_A = const(1)
print(_A)
```
compiles to
```
00 LOAD_NAME print
02 LOAD_CONST_SMALL_INT 1
03 CALL_FUNCTION n=1 nkw=0
05 POP_TOP
```
whereas
```
import micropython
_A = micropython.const(1)
print(_A)
```
compiles to
```
06 LOAD_NAME micropython
08 LOAD_METHOD const
10 LOAD_CONST_SMALL_INT 1
11 CALL_METHOD n=1 nkw=0
13 STORE_NAME _A
15 LOAD_NAME print
17 LOAD_NAME _A
19 CALL_FUNCTION n=1 nkw=0
21 POP_TOP
```

You can see that `_A` is now a global, and also has to be loaded as the argument to `print`.

(I'm not actually sure why we provide `micropython.const` at all... seems like it's an easy way to catch people out. I will investigate).

jimmo:
See https://github.com/micropython/micropython/issues/11929

mishal:
OK, thanks for info. It looks like that almost every lib uses `micropython.const` to optimize the code, but no optimisation is really done.

jimmo:
> It looks like that almost every lib uses `micropython.const` to optimize the code, but no optimisation is really done.

@mishal Could you clarify what you mean by "almost every lib"? Thanks

mishal:
sorry...

I wanted to say, that (If I understand it correctly) lib authors trying to optimize their code using `const` (from `micropython` module) are not optimizing anything...

I use i2c eeprom in my project: 

 * https://github.com/peterhinch/micropython_eeprom/blob/master/bdevice.py
 * https://github.com/peterhinch/micropython_eeprom/blob/master/eeprom/i2c/eeprom_i2c.py

Also other authors use `const` from `micropython` module

see: https://github.com/search?q=%22from+micropython+import+const%22&type=code

The docs say that micropython.const is what should be used:

https://docs.micropython.org/en/latest/library/micropython.html?highlight=const#micropython.const

mishal:
@jimmo thanks, now I get it!

This PR was about getting my IDE happy... there's no global const function in CPython. In your airepl.py you don't import the const from micropython module...

PR #694, opened 2023-06-27
python-stdlib/zlib: Add zlib module.

jimmo:
This is a replacement for the `zlib` module that used to be built-in and is now replaced by the gzip module (see https://github.com/micropython/micropython/pull/11879).

The `gzip.GzipFile` class (in conjunction with `io.BytesIO`) can be used to provide the old functionality of the `zlib` module (`decompress`, `DecompIO`) as well as adding `compress`.


dpgeorge:
This needs updating so that it doesn't use keyword args in `DeflateIO` construtor.

jimmo:
> This needs updating so that it doesn't use keyword args in the `DeflateIO` constructor.

Done.

python-stdlib/gzip/gzip.py
```diff
@@ -1,29 +1,28 @@
...
-        i += 1
-    if flg & FCOMMENT:
-        while data[i]:
-            i += 1
-        i += 1
-    if flg & FHCRC:
-        i += 2
-    return zlib.decompress(memoryview(data)[i:], -15)
+    f = io.BytesIO(data)
+    with gzip.GzipFile(f, deflate.GZIP, _WBITS) as g:
```
dpgeorge, line 19: Should this be `deflate.DeflateIO`?

jimmo: It should be GZipFile, but without the extra args. Updated.

dpgeorge:
Thanks!

PR #698, opened 2023-07-06
stdlib/fnmatch: Fix compatibility with ure -> re.

andrewleech:
With the recent micropython change to remove the `u` prefix by default on builtins (https://github.com/micropython/micropython/pull/11740) the format checker in fnmatch which was detecting `ure` no longer works. 

The MR updates the module to filter the regex automatically as needed.


jimmo:
Thanks, that's a much better solution than what was there before.

PR #702, opened 2023-07-21
all: Rename utarfile->tarfile and urequests->requests.

jimmo:
Extends on #540, with four main differences:
- Also renames utarfile
- Makes urequests a separate package. Rather than having requests provide urequest, you have to explicitly require it.
- Makes requests a package with `__init__.py` to allow extension in the future.
- Update bundle-networking and mip to use `requests`.

Note that the unix port currently gets `urequests` via mip, so now it will only get `requests`. For backwards compatibility, when the submodule is updated we will might want to consider adding `require("urequests")` to `unix/variants/manifest.py`. All other ports currently get `urequests` via bundle-networking and so will now get both.


jimmo:
I haven't provided a wrapper for utarfile... not sure this is necessary?

dpgeorge:
> I haven't provided a wrapper for utarfile... not sure this is necessary?

I'd say it's not necessary.  `tarfile-write` support was only added recently (there hasn't been a MicroPython release with this in it yet) so I think it's fair enough to just rename it.

Users can easily provide their own `utarfile` -> `tarfile` wrapper if needed.

PR #706, opened 2023-07-25
Add pre-commit support for micropython-lib.

jimmo:
1. Simplify codeformat.py back to be the same as the main repo version, but without .c/.h support. (It's pretty much just a wrapper around `black` now)
2. Add pre-commit support (same as main repo).
3. Split the ruff action into its own CI action (same as main repo).

### DIFF
--- a/.github/workflows/code_formatting.yml
+++ b/.github/workflows/code_formatting.yml
@@ -7,8 +7,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - run: pip install --user ruff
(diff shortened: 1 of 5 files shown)


jimmo:
In separate PRs, there are two follow-ups:
 - Add checking of the git commit messages. In the main repo we use our own tool (tools/verifygitlog.py) but I don't really want to duplicate that here, and I have a proof-of-concept use the standard `gitlint` tool that works nicely with pre-commit.
 - Same as the main, repo, add codespell checks too.

dpgeorge:
This looks good, thanks.

I also agree about the follow-up actions.

PR #712, opened 2023-08-08
lora: Remove the pin parameter from IRQ callback.

projectgus:
It's not necessary to know which pin triggered the IRQ, and it saves some code size.

Noticed while making changes for #572, but is actually an unrelated change.

---


dpgeorge:
A nice little simplification, thanks!

PR #715, opened 2023-08-10
Update ntptime.py

yanzhonghui:
Add time zone offset support

### DIFF
--- a/micropython/net/ntptime/ntptime.py
+++ b/micropython/net/ntptime/ntptime.py
@@ -13,6 +13,8 @@ except:
 host = "pool.ntp.org"
 # The NTP socket timeout can be configured at runtime by doing: ntptime.timeout = 2
 timeout = 1
+# The timezone can be configured at runtime by doing: ntptime.timezone = 8
+timezone = 0
 
 
 def time():
@@ -38,7 +40,7 @@ def time():
     else:
         raise Exception("Unsupported epoch: {}".format(EPOCH_YEAR))
 
-    return val - NTP_DELTA
+    return val - NTP_DELTA + timezone * 3600
 
 
 # There's currently no timezone support in MicroPython, and the RTC is set in UTC time.


projectgus:
Thanks @MrEbbinghaus for gathering all these PRs together. I generally agree with your conclusion and suggested approach.

Many (but I think not all) ports are built with `MICROPY_PY_TIME_GMTIME_LOCALTIME_MKTIME` support which implements 90% of what's needed for CPython compatible timezone support inside `extmod/modtime.c`. At first glance, the only thing that's missing to implement for basic embedded-friendly time zones is [time.timezone](https://docs.python.org/3.5/library/time.html#time.timezone), and the fact that `gmtime()` and `localtime()` are implemented as the same function.

I feel like there's probably a good reason why this isn't implemented already, though. @dpgeorge and @jimmo might know?

projectgus:
Oh, I see it. [`struct time` in Micropython](https://docs.micropython.org/en/latest/library/time.html#time.localtime) only has the first 8 fields of the [equivalent in CPython](https://docs.python.org/3.5/library/time.html#time.struct_time). So there's no way to add timezones here and keep backwards compatibility.

MrEbbinghaus:
A complete time zone database support would be unnecessary.

One could implement something like `time.tzset()`
For my current timezone this would be: 
`CET-1` (without DST)
`CET-1CEST,M3.5.6,M10.5.6/3` (with DST)

From that we could set 
`time.``daylight`

projectgus:
@MrEbbinghaus Good ideas, both on moving this discussion to a more suitable place, and the possibility of simplified timezone string support.

I've summarised all of the existing discussion and options I could find, here: https://github.com/orgs/micropython/discussions/12378 - please reply with more thoughts or suggestions.

@yanzhonghui As discussed above, everyone agrees MicroPython should solve this problem but the best place is probably not in the `ntptime` module. As such, I'm going to close this PR. Please comment on the linked discussion if you have any suggestions for what should be done!

PR #719, opened 2023-08-21
stdlib/os: Provide namedtuple response for os.stat().

andrewleech:
Having the option to wrap / return os.stat() with the cpython standard format can greatly aid in porting third party libraries.

### DIFF
--- a/python-stdlib/os/os/__init__.py
+++ b/python-stdlib/os/os/__init__.py
@@ -6,3 +6,28 @@ try:
     from . import path
 except ImportError:
     pass
+
+from collections import namedtuple
+
+# https://docs.python.org/3/library/os.html#os.stat_result
+stat_result = namedtuple(
+    "stat_result",
+    (
+        "st_mode",
+        "st_ino",
+        "st_dev",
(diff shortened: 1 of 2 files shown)


python-stdlib/os/os/__init__.py
```diff
@@ -1,8 +1,40 @@
...
-from uos import *
+# Include built-in os module.
+import sys
+
+__path = sys.path
+try:
+    sys.path.clear()
+    from os import *
+finally:
+    sys.path.extend(__path)
```
andrewleech, line 9: Hey @jimmo is this what you had in mind as the new / current best practice for importing and extending builtins here?

dpgeorge: My suggestion would be to just leave this as it was.

andrewleech: I've reverted the import change and rebased thanks.

dpgeorge (approved):

jonnor:
This change breaks mpremote. Fix in https://github.com/micropython/micropython/pull/19483

PR #723, opened 2023-09-03
bundle-networking: Include urllib.urequest

aveao:
`urllib.urequest.urlopen` provides an easy way to stream HTTP requests (which urequests can't do), and is seemingly generally recommended over urequests for anything that is more involved than doing a small HTTP request.

it's already [included by default with esp8266](https://github.com/micropython/micropython/blob/65f0cb11afb92cc9c333158ea467667f2127c9a2/ports/esp8266/boards/ESP8266_GENERIC/manifest_2MiB.py#L15) but somehow isn't with the rest of network-capable boards.

this PR changes it so that all boards that install bundle-networking get urllib.urequest too.

### DIFF
--- a/micropython/bundles/bundle-networking/manifest.py
+++ b/micropython/bundles/bundle-networking/manifest.py
@@ -7,6 +7,7 @@ require("mip")
 require("ntptime")
 require("requests")
 require("webrepl")
+require("urllib.urequest")
 
 # Provide urequests (which just forwards to requests) for backwards
 # compatibility.


jimmo:
Thanks @aveao -- Unfortunately, although I agree it's definitely convenient to have as much stuff installed by default as possible, I don't think we're likely to make this change.

> it's already included by default with esp8266 but somehow isn't with the rest of network-capable boards.

This is a bit of a historical accident with the esp8266, it includes too much stuff by default, I think at the time there was excitement about having the first network-capable board and providing features for the kickstarter. It's always easy to add stuff but then we can almost never remove it.

We've also made it easier to install packages since then (e.g. `mip`).

> provides an easy way to stream HTTP requests *(which urequests can't do)*

Can you explain more what you mean by this? If you have a Response object from `requests`, you can use `resp.raw` to access the underlying socket. (This is true of upstream `requests` too). Not sure if that's what you meant by streaming though?

> this PR changes it so that all boards that install bundle-networking get urllib.urequest too.

This is the problem -- it's a huge decision to add the additional firmware size to every single board, especially when there's already `requests`, when people who need it can just install it from mip.

andrewleech:
For what it's worth my vote is not not include it by default; I regularly use network capable boards and my primary two use cases are:
* Based on mqtt to communicate with home automation gear
* With microdot to host a small web server interface to the hardware

I for one almost never call out to external http servers so never use requests. 
On the rare occasion I do want to make a http request I'll use an asyncio library like [`uaiohttpclient`](https://github.com/micropython/micropython-lib/tree/master/micropython/uaiohttpclient)

aveao:
ah, didn't realize `urequests` had that functionality, it's not documented on micropython side and I tend to not rely too much on upstream python docs as micropython tends to implement a limited selection of it.

also: it's weird that `urllib.urequest.urlopen` exists at all, it's pretty much a [two-line change](https://git.flow3r.garden/flow3r/flow3r-firmware/-/merge_requests/352/diffs#diff-content-c9dae8680ee66d91580384c2940bda22ca4409b1) to go from `urllib.urequest.urlopen` to using the `.raw` and it's not much smaller than urequests.

anyhow, closing this.

PR #724, opened 2023-09-05
micropython/uaiohttpclient: Add ssl support.

Carglglz:
Add ssl support for `uaiohttpclient` compatible with `SSLContext` from:

- [ ] https://github.com/micropython/micropython/pull/11897

### DIFF
--- a/micropython/uaiohttpclient/clientsession_example.py
+++ b/micropython/uaiohttpclient/clientsession_example.py
@@ -1,0 +1,17 @@
+import uaiohttpclient as aiohttp
+import asyncio
+
+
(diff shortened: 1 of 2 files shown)


ned-pcs:
@Carglglz How does one do a `POST` or `PUT` with a body?

Carglglz:
> @Carglglz How does one do a `POST` or `PUT` with a body?

I haven't implemented those yet, the main goal of this PR is to add SSL support to the existing `aiohttpclient`. But I guess if there is interest I could implement those methods in a separate PR following this one. 👍🏼

PR #727, opened 2023-09-14
pathlib: add Path.expanduser()

scivision:


python-stdlib/pathlib/pathlib.py
```diff
@@ -184,6 +184,13 @@ def with_suffix(self, suffix):
         index = -len(self.suffix) or None
         return Path(self._path[:index] + suffix)
 
+    def expanduser(self):
+        if self._path == "~" or self._path.startswith("~/"):
```
stinos, line 188: Should use `_SEP` not `/`

python-stdlib/pathlib/pathlib.py
```diff
@@ -184,6 +184,13 @@ def with_suffix(self, suffix):
...
+            return Path(os.getenv("HOME") + self._path[1:])
+        if self._path[0] == "~":
```
stinos, line 190: CPython doesn't raise an error in this case, maybe just leave this out? Or what was the intent (all code paths should probably be unittested)?

python-stdlib/pathlib/tests/test_pathlib.py
```diff
@@ -322,3 +322,8 @@ def test_with_suffix(self):
         self.assertTrue(Path("foo/test").with_suffix(".tar") == Path("foo/test.tar"))
         self.assertTrue(Path("foo/bar.bin").with_suffix(".txt") == Path("foo/bar.txt"))
         self.assertTrue(Path("bar.txt").with_suffix("") == Path("bar"))
+
+    def test_expanduser(self):
+        self.assertFalse(Path("~") == "~")
```
stinos, line 327: Can a Path instance ever be equal to a string?

BrianPugh (approved):

dpgeorge (approved):
Thanks for the contribution!

PR #728, opened 2023-09-14
aioble/server.py: Add data arg for indicate.

jimmo:
In micropython/micropython#11239 we added support for passing data to gatts_indicate (to make it match gatts_notify).

This adds the same to aioble.

Also update the documentation to mention this (and fix some mistakes and add a few more examples).

_This work was funded through GitHub Sponsors._

PR #729, opened 2023-09-14
aioble/server.py: Allow BufferedCharacteristic to support all ops.

jimmo:
Updated version of #629 (adds a test and slightly simplifies the argument handling). I'm not too fussed about the backwards compatibility here because the existing behavior was broken, it never should have been read=True by default, rather it should have been `write=True`. (The whole point of the buffered characteristic is to allow the client to _write_ larger values, and optionally have them accumulated via the append flag).

---

Previously a BufferedCharacteristic could only be read by the client, where it should have been writeable. This makes it support all ops (read / write / write-with-response, etc).

Adds a test to check the max_len and append functionality of BufferedCharacteristic.

### DIFF
--- a/micropython/bluetooth/aioble-server/manifest.py
+++ b/micropython/bluetooth/aioble-server/manifest.py
@@ -1,4 +1,4 @@
-metadata(version="0.4.0")
+metadata(version="0.4.1")
 
 require("aioble-core")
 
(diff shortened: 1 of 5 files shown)


jimmo:
I will update this will the version number bump and aioble/README.md update after #[728](https://github.com/micropython/micropython-lib/pull/728) is merged.

PR #730, opened 2023-09-14
os.path: refactor: use sep instead of literal /

scivision:


dpgeorge:
Thanks for the contribution.  Can you please explain what prompted this change and why it's better this way?

python-stdlib/os-path/os/path.py
```diff
@@ -13,28 +13,28 @@ def normpath(s):
...
+    if s[0] != sep:
+        return os.getcwd() + sep + s
     return s
 
 
 def join(*args):
     # TODO: this is non-compliant
     if type(args[0]) is bytes:
-        return b"/".join(args)
+        return bytes(sep).join(args)
```
projectgus, line 24: This line will raise an exception:

```
MicroPython v1.20.0-330-g076bfc7ed8-dirty on 2023-09-26; linux [GCC 13.2.1] version
>>> bytes("/")
TypeError: string argument without an encoding
>>> "/".encode()
b'/'
```

projectgus (changes requested):
Change LGTM apart from one small fix needed. However like @dpgeorge I'm curious what the motivation for making it was. (Windows, maybe?)

dpgeorge:
Closing due to inactivity.

PR #737, opened 2023-09-28
unix-ffi: re: convert to PCRE2

Ansuel:
PCRE is marked as EOL and won't receive any new security update.

Convert the re module to PCRE2 API to enforce security. Additional dependency is now needed with uctypes due to changes in how PCRE2 return the match_data in a pointer and require special handling.

The converted module is tested with the test_re.py with no regression.

### DIFF
--- a/unix-ffi/re/re.py
+++ b/unix-ffi/re/re.py
@@ -1,36 +1,55 @@
 import sys
 import ffilib
 import array
+import uctypes
+
+pcre2 = ffilib.open("libpcre2-8")
+
+#       pcre2_code *pcre2_compile(PCRE2_SPTR pattern, PCRE2_SIZE length,
+#           uint32_t options, int *errorcode, PCRE2_SIZE *erroroffset,
+#           pcre2_compile_context *ccontext);
+pcre2_compile = pcre2.func("p", "pcre2_compile_8", "siippp")
+
+#       int pcre2_match(const pcre2_code *code, PCRE2_SPTR subject,
+#           PCRE2_SIZE length, PCRE2_SIZE startoffset, uint32_t options,
+#           pcre2_match_data *match_data, pcre2_match_context *mcontext);
+pcre2_match = pcre2.func("i", "pcre2_match_8", "Psiiipp")
+
+#       int pcre2_pattern_info(const pcre2_code *code, uint32_t what,
+#           void *where);
+pcre2_pattern_info = pcre2.func("i", "pcre2_pattern_info_8", "Pip")
+
+#       PCRE2_SIZE *pcre2_get_ovector_pointer(pcre2_match_data *match_data);
+pcre2_get_ovector_pointer = pcre2.func("p", "pcre2_get_ovector_pointer_8", "p")
+
+#       pcre2_match_data *pcre2_match_data_create_from_pattern(const pcre2_code *code,
+#           pcre2_general_context *gcontext);
+pcre2_match_data_create_from_pattern = pcre2.func(
+    "p", "pcre2_match_data_create_from_pattern_8", "Pp"
+)
+
+# PCRE2_SIZE that is of type size_t.
+# Use ULONG as type to support both 32bit and 64bit.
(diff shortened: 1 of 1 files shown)


Ansuel:
(this is needed as we are deprecating pcre in openwrt and we are converting any user to pcre2)

dpgeorge:
Thanks for the contribution.

What's the difference between `libpcre2-posix`, `libpcre2-8`, `libpcre2-16` and `libpcre2-32`?

It would be possible to support both the old PCRE and PCRE2, but I don't think it's worth the effort.

Ansuel:
The POSIX variant just provide wrapper to have POSIX interface for pcre2 API.
Pcre had the same variant and wasn't used here.

The -16 and -32 variant are used to support 16bit and 32bit string (UTF-16 and UTF-32)
UTF-8 should be enough and what is commonly used when migrating from pcre to pcre2

Supporting both versions would bloat the code with lots of ifdef and IMHO it's not worth it and pcre won't receive any update and would pose security risks.

dpgeorge:
> The -16 and -32 variant are used to support 16bit and 32bit string (UTF-16 and UTF-32)
> UTF-8 should be enough and what is commonly used when migrating from pcre to pcre2

Aah, I see, thanks for the info.  Indeed we should use the -8 variant.

I see that the `test_re.py` provided in this `re` directory doesn't pass anymore.  It needs to pass.

Ansuel:
@dpgeorge i remember it did pass correctly. Maybe i'm running it wrong? my test was compiling the lib and run the test using the compiled library. Something wrong in this?

dpgeorge:
I'm simply running the following in the `unix-ffi/re` directory with this PR checked out:
```
$ micropython
MicroPython v1.20.0-557-g5aec051f9 on 2023-10-05; linux [GCC 13.2.1] version
>>> import re
>>> re.search(r"a+", "caaab").group(0)
''
```
That's not correct output.

Ansuel:
@dpgeorge mhh i'm confused

```
root@OpenWrt:~# micropython
MicroPython v1.20.0 on 2023-10-04; linux [GCC 12.3.0] version
>>> import re
>>> re.search(r"a+", "caaab").group(0)
'aaa'
```

Ansuel:
Also by putting the uncompiled re.py in /usr/lib/micropython/unix produce the same output

Ansuel:
@dpgeorge I finally manage to repro the problem... It seems to be a problem with PCRE2_SIZE that is size_t and this change between 32bit and 64bit. Using ULONG handled both systems. (my test were done on a 32bit system)

unix-ffi/re/re.py
```diff
@@ -67,19 +80,26 @@ def __init__(self, compiled_ptn):
...
+        match_data = pcre2_match_data_create_from_pattern(self.obj, None)
+        num = pcre2_match(self.obj, s, len(s), pos, _flags, match_data, None)
         if num == -1:
             # No match
             return None
+        ov_ptr = pcre2_get_ovector_pointer(match_data)
+        # pcre2_get_ovector_pointer return PCRE2_SIZE that is of type
+        # size_t. Use ULONG as type to support both 32bit and 64bit.
+        ov_buf = uctypes.bytearray_at(
+            ov_ptr, uctypes.sizeof({"field": 0 | uctypes.ULONG}) * (cap_count + 1) * 2
```
dpgeorge, line 94: Can this call to sizeof to get the size be done at global module level, so it only has to be done once?

dpgeorge:
Thanks, I can confirm that it now works on my (64-bit) machine.

unix-ffi/re/re.py
Ansuel: Better now?

BKPepe:
So, @dpgeorge it can be merged, right? :)

projectgus (approved):
Changes LGTM, thanks @Ansuel!

dpgeorge:
Thanks for updating, this looks good now.

Merged.

PR #749, opened 2023-10-17
top: Pin black version.

jimmo:
This replaces the codeformat.py script with just running black directly, both from GitHub CI and from pre-commit, same as how we do `ruff`.

This allows us to avoid needing to use the system black version for pre-commit, and lets us pin the same version for both.

pyproject.toml is now used for configuration.

The old codeformat.py is still kept (but now just runs black directly).


.github/workflows/code_formatting.yml
```diff
@@ -1,16 +1,10 @@
...
-    - uses: actions/checkout@v3
-    - uses: actions/setup-python@v4
-    - name: Install packages
-      run: source tools/ci.sh && ci_code_formatting_setup
-    - name: Run code formatting
-      run: source tools/ci.sh && ci_code_formatting_run
-    - name: Check code formatting
-      run: git diff --exit-code
+    - uses: actions/checkout@v4
+    - run: pip install --user black==23.7.0
```
projectgus, line 9: Maybe I have too much DRY brainworms, but could we add a `[project.optional-dependencies]` section to pyproject.toml for these versions? (I/we could also do this in a follow-up PR for this & ruff together, if that's easier.)

The main advantage is that we could then tell developers to install the linter tools from there (and Python developers will instinctively go there anyway). This means they will have matching versions in any IDE integrations, pylsp, etc and [hopefully] won't run into the situation of the pre-commit hook re-formatting (or re-linting) code that their IDE had already formatted or linted slightly differently.

(AFAIK the version still needs to be duplicated in two places even then, because pre-commit maintains its own sandboxes and reads those versions `.pre-commit-config.yaml`. No way around that, from what I can see.)

projectgus (approved):
Yay pinned tool versions!

PR #710, opened 2023-08-05
logging: add call to superclass initialiser; set default formatter

colin-nolan:
Addresses minor issues in `logging` module, reported in https://github.com/micropython/micropython-lib/issues/691. 

## Change 1: Makes subclasses of `logging.Handler` call the superclasses' initialiser
CPython:
```
>>> import logging
>>> logger = logging.StreamHandler()
>>> logger.level 
0
```

MicroPython (`master`):
```
>>> logger = logging.StreamHandler()
>>> logger.level
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'StreamHandler' object has no attribute 'level'
```

MicroPython (this branch):
```
>>> logger = logging.StreamHandler()
>>> logger.level
0
```

## Change 2: Sets a default formatter if not set
CPython:
```
>>> record = logging.LogRecord("Test", 20, "/example", 1, "Hello", (), None)
>>> logger.emit(record)
Hello
```

MicroPython (`master` + fix 1):
```
>>> record = logging.LogRecord()
>>> record.set("Test", 20, "Hello")
>>> logger.emit(record)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "logging.py", line 72, in emit
  File "logging.py", line 56, in format
AttributeError: 'NoneType' object has no attribute 'format'
```

MicroPython (this branch):
```
>>> record = logging.LogRecord()
>>> record.set("Test", 20, "Hello")
>>> logger.emit(record)
INFO:Test:Hello
```

The CPython implementation of `logging` can be found here:
https://github.com/python/cpython/blob/main/Lib/logging/__init__.py

PR #750, opened 2023-10-18
logging: CPython-compatible logging improvements

ned-pcs:
This commit allows for logging handlers to be added to the root logger and then used by non-root loggers that don't have their own handlers.

It also adds the ([CPython-compatible](https://docs.python.org/3.13/library/logging.html#logging.basicConfig)) `handlers` argument to `logging.basicConfig()` to ease this initialization:
```py
    import logging
    sh = logging.StreamHandler()
    fh = logging.FileHandler("my.log", mode="a")
    logging.basicConfig(handlers=[sh, fh])

    root_logger = logging.getLogger() # uses sh and fh
    another_logger = logging.getLogger("another") # inherits handlers
```

It also adds the `Logger.removeHandler()` method and avoids repeated handler addition.

It also correctly calls the superclass constructor from the `StreamHandler` constructor and uses a default formatter if a Handler has none set (as in PR #710).

### DIFF
--- a/python-stdlib/logging/logging.py
+++ b/python-stdlib/logging/logging.py
@@ -54,6 +54,9 @@ class Handler:
     def close(self):
         pass
 
+    def flush(self):
+        pass
+
     def setLevel(self, level):
         self.level = level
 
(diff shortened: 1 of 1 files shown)


ned-pcs:
build failure appears to be due to manifest issues unrelated to this PR.

ned-pcs:
Closed; replaced with #806

PR #751, opened 2023-10-19
requests: Added support for some non-standard URLs

PermissionDenied7335:
In some cases, servers may return non-standard URLs like `https://ecard.v.zzu.edu.cn?tid=xxxx` in "Location" line, which could not be properly treated since there are no slashes between hostname and remote path. So I added my code to separate remote path from hostname.

### DIFF
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -68,6 +68,14 @@ def request(
         port = 443
     else:
         raise ValueError("Unsupported protocol: " + proto)
+    
+    if "?" in host:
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
Thanks for the patch.  This looks correct to me, it's certainly allowed to have URLs like `http://example.com?query`.  So we should support them, as done with this PR.

dpgeorge:
Rebased and merged in fea2adb47b927f62e840563a09ef1856527e9e81, with a test added in f309584ad9608097f185612f8546c4ae414e3926

PR #758, opened 2023-10-26
tools/verifygitlog.py: Add git commit message checking.

jimmo:
This adds verifygitlog.py from the main repo, adds it to GitHub workflows, and also pre-commit.

I spent a considerable amount of effort trying to make `gitlint` (https://jorisroovers.com/gitlint) work (for both the main repo and here) but it was not flexible enough to make it work with our rules. See https://github.com/jimmo/micropython/tree/gitlint-config

The three main limitations were:
 - I ended up having to write a non-trivial amount of Python code as custom rules.
 - It doesn't have a way to link to documentation or add suggestions like verifygitlog.py does.
 - I couldn't find a way to make `WIP` and `!fix` commits work in pre-commit but not in CI.

So in the end it's just easier to keep using verifygitlog.py (even if it means duplicating it here).

### DIFF
--- a/.github/workflows/commit_formatting.yml
+++ b/.github/workflows/commit_formatting.yml
@@ -1,0 +1,18 @@
+name: Check commit message formatting
+
+on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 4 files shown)


dpgeorge:
Thanks!

PR #764, opened 2023-11-07
uaiohttpclient: parse port instead of hard code 80

bulletmark:
I was getting an error using `uaiohttpclient` and it was due to it hard coding port 80 for the target host. This PR makes a trivial change to split `host:port` if present in the URL. I thought it also sensible to update all the `yield from` to `await` everywhere and of course preface those functions `async`. I made a couple of other minor changes which were suggested by ruff and flake8, i.e. a variable `l` I changed to `ln` and there was a comment > 80 chars which I wrapped.

### DIFF
--- a/micropython/uaiohttpclient/example.py
+++ b/micropython/uaiohttpclient/example.py
@@ -1,31 +1,16 @@
 #
 # uaiohttpclient - fetch URL passed as command line argument.
 #
+import sys
 import uasyncio as asyncio
 import uaiohttpclient as aiohttp
 
 
-def print_stream(resp):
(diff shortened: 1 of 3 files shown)


bulletmark:
Although trivial, I have changed this PR from 1 commit into 4 separate commits because they are logically distinct changes and this makes it easier to review. I can also easily remove or squash any commits you request.

dpgeorge:
Thank you, this is a good fix for the port number, and also a nice clean up with the other commits.  And thanks also for splitting the commits into logical changes.

PR #765, opened 2023-11-09
SDcard sdcard.py rewrite for efficiency and function

mendenm:
This module has been heavily reworked.  I went through the logic, and compared it to the SDCard standard, and found numerous ways it could be made faster, and work more reliably.  Lots of code has disappeared.  Also, it computes  CRC7 on al commands (some cards didn't work without this), and has the option to compute CRC16 on data transfers, if the user provides a suitable function.


jimmo:
Thanks @mendenm -- at first glance this all looks very good. I will try to find some time soon to review in detail.

One quick note... we don't currently have a way to publish .mpy files with native code to `mip`, so we will need to solve that first. Also we will need a fallback implementation for devices that do not support loading native code.

mendenm:
@jimmo Is there any chance you can take a quick look at the PR to see if you can see why ruff keeps crashing (exiting with error 1)? Running is locally seems to return no problems, but I can't get past the github process.

Thanks.

mendenm:
> @jimmo Is there any chance you can take a quick look at the PR to see if you can see why ruff keeps crashing (exiting with error 1)? Running is locally seems to return no problems, but I can't get past the github process.
> 

Never mind, I got past it.  

Thanks.

dpgeorge:
As @jimmo mentioned we don't yet have support for `@micropython.viper` (or native or asm_thumb) in this repository, because the `.py` file is compiled to `.mpy` and that compilation must use a different architecture flag for different target machines, and hence we would need different `.mpy` files, one for each supported architecture.

So, to make progress with this PR I suggest:
- move the fast crc routines to example/util files which are not included in the `manifest.py`, and hence not included in the distributed package
- implement only the `crc7` function and do it in pure Python
- still support passing in a function for CRC16 data checks, the user can then copy the fast `crc16` routines if they want to use this feature

micropython/drivers/storage/sdcard/crc7.py
```diff
@@ -0,0 +1,85 @@
...
+#         0x1c, 0x0e, 0x38, 0x2a, 0x54, 0x46, 0x70, 0x62,
+#         0x8c, 0x9e, 0xa8, 0xba, 0xc4, 0xd6, 0xe0, 0xf2
+#     ]):
+#         crc7_be_syndrome_table[i] = x #fill pre-allocated table
+#
+#
+#     import gc
+#     gc.collect() #make sure residual list gets cleaned up to save memory
+
+crc7_be_syndrome_table = bytearray(256)  # pre-allocate so it is low in memory
```
dpgeorge, line 50: I suggest just making this table a pre-computed `bytes` object.  That will have the smallest memory footprint, and take no time to create when importing this code.

micropython/drivers/storage/sdcard/sdcard.py
```diff
@@ -63,89 +77,89 @@ def init_spi(self, baudrate):
             # on pyboard
             self.spi.init(master, baudrate=baudrate, phase=0, polarity=0)
 
+    @staticmethod
+    def gb(bigval, b0, bn):
```
dpgeorge, line 81: I suggest making this a global function in the module, ie not part of the `SDCard` class.  Also name it `_gb` to indicate it's a private function.

micropython/drivers/storage/sdcard/sdcard.py
```diff
@@ -63,89 +77,89 @@ def init_spi(self, baudrate):
...
+        # get numbered bits from a buf_to_int from, for example, the CSD
+        return (bigval >> b0) & ((1 << (1 + bn - b0)) - 1)
+
+    def spiff(self):
```
dpgeorge, line 85: Suggest calling it `_spiff` to indicate a private method.

micropython/drivers/storage/sdcard/sdcard.py
```diff
@@ -63,89 +77,89 @@ def init_spi(self, baudrate):
...
-                    # SDSC card, uses byte addressing in read/write/erase commands
-                    self.cdv = 512
-                else:
-                    # SDHC/SDXC card, uses block addressing in read/write/erase commands
-                    self.cdv = 1
-                # print("[SDCard] v2 card")
-                return
-        raise OSError("timeout waiting for v2 card")
-
     def cmd(self, cmd, arg, crc, final=0, release=True, skip1=False):
```
dpgeorge, line 155: I suggest removing the `crc` argument and just always computing it in this function.  The only times it's passed in as non-None is for a few initialisation/detection commands at the start, and that's only done once so doesn't need to be efficient (in time).

Removing the `crc` argument will make code size smaller and faster for the case when `None` is passed (which is the majority of uses).

mendenm: I agree with this.  The passing in of fixed CRCs is historically done for the few initialization packets, but why?  Do you want me to make the change in my fork, or do you want to do this after you pull it?

mendenm: is this OK style for the crc table, including expectation of memory footprint after parsing?:

```python
crc7_be_syndrome_table = (
b'\x00\x12$6HZl~\x90\x82\xb4\xa6\xd8\xca\xfc\xee2 \x16\x04zh^L\xa2\xb0\x86\x94\xea\xf8'
b'\xce\xdcdv@R,>\x08\x1a\xf4\xe6\xd0\xc2\xbc\xae\x98\x8aVDr`\x1e\x0c:(\xc6\xd4\xe2\xf0'
b'\x8e\x9c\xaa\xb8\xc8\xda\xec\xfe\x80\x92\xa4\xb6XJ|n\x10\x024&\xfa\xe8\xde\xcc\xb2\xa0'
b'\x96\x84jxN\\"0\x06\x14\xac\xbe\x88\x9a\xe4\xf6\xc0\xd2<.\x18\ntfPB\x9e\x8c\xba\xa8'
b'\xd6\xc4\xf2\xe0\x0e\x1c*8FTbp\x82\x90\xa6\xb4\xca\xd8\xee\xfc\x12\x006$ZH~l\xb0\xa2'
b'\x94\x86\xf8\xea\xdc\xce 2\x04\x16hzL^\xe6\xf4\xc2\xd0\xae\xbc\x8a\x98vdR@>,\x1a\x08'
b'\xd4\xc6\xf0\xe2\x9c\x8e\xb8\xaaDV`r\x0c\x1e(:JXn|\x02\x10&4\xda\xc8\xfe\xec\x92\x80'
b'\xb6\xa4xj\\N0"\x14\x06\xe8\xfa\xcc\xde\xa0\xb2\x84\x96.<\n\x18ftBP\xbe\xac\x9a\x88\xf6'
b'\xe4\xd2\xc0\x1c\x0e8*TFpb\x8c\x9e\xa8\xba\xc4\xd6\xe0\xf2'
)
```

dpgeorge: > The passing in of fixed CRCs is historically done for the few initialization packets, but why? Do you want me to make the change in my fork, or do you want to do this after you pull it?

I'm not exactly sure why the CRC was there only for the few initialisation packets.  But my guess would be 1) it's a static value so doesn't need a function to compute it, 2) for initialisation it may be that some cards require the CRC on these commands.

Anyway, I'm happy for you to make this change as part of this PR.



> is this OK style for the crc table, including expectation of memory footprint after parsing?

Yes that looks OK.  Maybe indent each line within the parenthesis, if `ruff format` allows?  (we switched from black to ruff).  Eg:
```py
table = (
    b'...'
    b'...'
)
```

Also I suggest putting this crc7 function within `sdcard.py`.  It's simple enough and saves having separate files for the driver, ie everything is self contained in one file.

mendenm:
I did a massive squash of the git repo, to get rid of all the drafts.  I think this should be very close to consistent with comments and request.  One thing I did not do was make gb() in _gb().  It's not really that private.  Future users may want to parse other fields out the the big status block to get more info about their sdcards.

Gadgetoid:
Is there a reason this is still stuck in PR purgatory?

Some cursory checks suggest that:

1. It still works
2. It's slightly faster

I'd be happy to vendor these libs into one of our builds to sniff out potential issues if it helps.

mendenm:
@Gadgetoid I'd sure be glad to have more people stress-testing it.  I think the code is robust, but it's always the case that more test cases are better.  I personally have only tested it on my Pi Pico, but with a fairly wide variety of SD cards, from ancient (< 2 GB) to 32 GB flavors.

dpgeorge (commented):
Sorry this got lost for so long.  I've now done a final review and tested it on a PYBv1.0.  It works well, and this will be a really great improvement to have!

dpgeorge:
> I have a question about the `gb()` naming and the `spiff` and the `blocks()as naming.  I was under the impression that micropython actually pays a small time penalty for longer names, so I tried to keep most names fairly short.  If this isn’t really an issue, I’ll be glad to make these changes

It's more about the memory use than speed.  Shorter names take less flash/memory, are smaller in compiled .mpy files and faster to transmit/copy.  I'm not sure it's faster to use shorter names, if it is it would only be a very tiny improvement due to a shorter `memcmp()`.

For `gb()` I just feel that it's too cryptic for a public function.  Maybe `bits()` would be better and still short?

For `spiff()`: this is purely a helper function and can be called anything, even `_s()` if you like.  But maybe it's not even needed, because it's such a short function it may be better to just write it out inline wherever it's used.  After all, in most cases the code has a `w = self.spi.write` local variable so `spiff` would be `w(b'\xff')`.

If you want to measure the impact of a change to the Python code, you can run it through `mpy-cross -O3 sdtest.py` and see how big the output `sdcard.mpy` is.  The smaller the better!  (Because smaller means less bytecode (unless you also shorten names), means faster execution.)

mendenm:
Thanks for the clarification.  I may move _spiff back inline.  I created before I started pre-binding the writes to the short name ‘w’, so its purpose may be superseded now. I may have time this weekend to do all the editing.MarcusOn Apr 10, 2025, at 7:55 PM, Damien George ***@***.***> wrote:﻿
  
I have a question about the gb() naming and the spiff and the `blocks()as naming.  I was under the impression that micropython actually pays a small time penalty for longer names, so I tried to keep most names fairly short.  If this isn’t really an issue, I’ll be glad to make these changes

It's more about the memory use than speed.  Shorter names take less flash/memory, are smaller in compiled .mpy files and faster to transmit/copy.  I'm not sure it's faster to use shorter names, if it is it would only be a very tiny improvement due to a shorter memcmp().
For gb() I just feel that it's too cryptic for a public function.  Maybe bits() would be better and still short?
For spiff(): this is purely a helper function and can be called anything, even _s() if you like.  But maybe it's not even needed, because it's such a short function it may be better to just write it out inline wherever it's used.  After all, in most cases the code has a w = self.spi.write local variable so spiff would be w(b'\xff').
If you want to measure the impact of a change to the Python code, you can run it through mpy-cross -O3 sdtest.py and see how big the output sdcard.mpy is.  The smaller the better!  (Because smaller means less bytecode (unless you also shorten names), means faster execution.)—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>



dpgeorge left a comment (micropython/micropython-lib#765)

I have a question about the gb() naming and the spiff and the `blocks()as naming.  I was under the impression that micropython actually pays a small time penalty for longer names, so I tried to keep most names fairly short.  If this isn’t really an issue, I’ll be glad to make these changes

It's more about the memory use than speed.  Shorter names take less flash/memory, are smaller in compiled .mpy files and faster to transmit/copy.  I'm not sure it's faster to use shorter names, if it is it would only be a very tiny improvement due to a shorter memcmp().
For gb() I just feel that it's too cryptic for a public function.  Maybe bits() would be better and still short?
For spiff(): this is purely a helper function and can be called anything, even _s() if you like.  But maybe it's not even needed, because it's such a short function it may be better to just write it out inline wherever it's used.  After all, in most cases the code has a w = self.spi.write local variable so spiff would be w(b'\xff').
If you want to measure the impact of a change to the Python code, you can run it through mpy-cross -O3 sdtest.py and see how big the output sdcard.mpy is.  The smaller the better!  (Because smaller means less bytecode (unless you also shorten names), means faster execution.)

mischif:
I can't say for sure yet where the issue is, but this driver always errors out when I try to bring up an sd card:

```python
  File "/lib/freedeej/hw.py", line 188, in sd_init
  File "sdcard/sdcard.py", line 85, in __init__
  File "sdcard/sdcard.py", line 121, in init_card
  File "sdcard/sdcard.py", line 209, in cmd
OSError: [Errno 5] EIO: CRC err on cmd: 00
```

What's weird is I swear this code _used_ to work intermittently with the reader/card I'm using, but maybe it's possible this aliexpress reader just gave up the ghost.

JoGeDuBo:
@mischif 
I had the same problem. 
My investigations revealed that the error occurs reproducibly exactly when 
the SD card is accessed for the first time 
(after switching on the power supply or after inserting the SD card).
The solution in init_card() in sdcard.py:

```
        # CMD0: init card; should return _R1_IDLE_STATE (allow 5 attempts)
        for _ in range(5):
            try: 
                if self.cmd(0, 0, 0x95) == _R1_IDLE_STATE:
                    break
            except OSError as err:
                if err.errno != EIO:
                    raise
        else:
            raise OSError(ENODEV, "no SD card")

```

PR #778, opened 2023-12-10
micropython/aiohttp: Add aiohttp package.

Carglglz:
Implement an updated version of `uaiohttpclient` (renamed `aiohttp`) with `ClientSession`, `WebSocketClient` and `SSLContext` support, which is mostly compatible with CPython [aiohttp-v3.9.1.](https://docs.aiohttp.org/) client i.e. https://docs.aiohttp.org/en/stable/client_quickstart.html

This replaces #752 and #724 and requires:
- [x] https://github.com/micropython/micropython/pull/11897

### DIFF
--- a/python-ecosys/aiohttp/aiohttp/__init__.py
+++ b/python-ecosys/aiohttp/aiohttp/__init__.py
@@ -1,0 +1,264 @@
+# MicroPython aiohttp library
+# MIT license; Copyright (c) 2023 Carlos Gil
+
+import asyncio
+import json as _json
+from .aiohttp_ws import (
+    _WSRequestContextManager,
+    ClientWebSocketResponse,
+    WebSocketClient,
+    WSMsgType,
+)
+
+HttpVersion10 = "HTTP/1.0"
+HttpVersion11 = "HTTP/1.1"
+
+
+class ClientResponse:
+    def __init__(self, reader):
+        self.content = reader
+
+    def _decode(self, data):
+        c_encoding = self.headers.get("Content-Encoding")
+        if c_encoding in ("gzip", "deflate", "gzip,deflate"):
+            try:
+                import deflate, io
+
+                if c_encoding == "deflate":
+                    with deflate.DeflateIO(io.BytesIO(data), deflate.ZLIB) as d:
+                        return d.read()
+                elif c_encoding == "gzip":
+                    with deflate.DeflateIO(io.BytesIO(data), deflate.GZIP, 15) as d:
+                        return d.read()
+            except ImportError:
+                print("WARNING: deflate module required")
+        return data
+
+    async def read(self, sz=-1):
+        return self._decode(await self.content.read(sz))
+
+    async def text(self, encoding="utf-8"):
+        return (await self.read(sz=-1)).decode(encoding)
+
+    async def json(self):
+        return _json.loads(await self.read())
+
+    def __repr__(self):
+        return "<ClientResponse %d %s>" % (self.status, self.headers)
+
+
+class ChunkedClientResponse(ClientResponse):
+    def __init__(self, reader):
+        self.content = reader
+        self.chunk_size = 0
+
+    async def read(self, sz=4 * 1024 * 1024):
+        if self.chunk_size == 0:
+            l = await self.content.readline()
+            l = l.split(b";", 1)[0]
+            self.chunk_size = int(l, 16)
+            if self.chunk_size == 0:
+                # End of message
+                sep = await self.content.read(2)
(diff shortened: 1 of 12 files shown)


dpgeorge:
This is a really great contribution, thank you!

A few top-level comments/questions:
- Did you write this from scratch, or copy pieces from other locations (maybe `uaiohttpclient`)?  The main question here is about licensing/copyright.
- For the things that are implemented do they match CPython's aiohttp exactly?  If not, can you document any differences?
- This library should be placed in the `python-ecosys` directory, because it's aiming to match something that already exists (in the Python ecosystem).  And the manifest.py should have the option `pypi="aiohttp"` added to the `metadata()` entry there (see other packages in the `python-ecosys` dir for hints).
- I suggest to keep the old `uaiohttpclient` library as-is, and it can be deprecated and removed in the future.  Essentially `aiohttp` will be a completely new package.

Carglglz:
Thanks for the feedback!

> Did you write this from scratch, or copy pieces from other locations (maybe uaiohttpclient)? The main question here is about licensing/copyright. 

The original idea was to add this to `uaiohttpclient` , but I felt that it was a bit outdated so I though it would be better to add this as a replacement. ( although I agree with keeping `uaiohttpclient` as-is). So the source code I used comes from:
-  `uaiohttpclient` (MIT) -> `ClientResponse` , `ChunkedClientResponse`, `_request`, `request_raw`
-  `aiohttp` (Apache-2) -> https://docs.aiohttp.org/en/stable/client_reference.html
And for WebSockets I used this too:
-  (MIT) https://github.com/danni/uwebsockets
- (MIT) https://github.com/miguelgrinberg/microdot/blob/main/src/microdot_asyncio_websocket.py

> For the things that are implemented do they match CPython's aiohttp exactly? If not, can you document any differences?

I would say so, e.g. all examples run as expected in MicroPython and CPython.

The **TLDR** of the things that are implemented is: 
most features of https://docs.aiohttp.org/en/stable/client_quickstart.html (except Timeouts, and file uploads)
and some of https://docs.aiohttp.org/en/stable/client_advanced.html (i.e. headers and `SSLContext`)

I think the main difference is that MicroPython implementation is missing some keyword arguments in `ClientSession` and in requests method (I may update this with the "exact" diff if necessary)

Also if you see anything that can be improved let me know or feel free to add any changes and I can review them later (whatever is faster 👍🏼 )

python-ecosys/aiohttp/aiohttp/__init__.py
```diff
@@ -0,0 +1,330 @@
...
+                data=data,
+                json=json,
+                ssl=ssl,
+                params=params,
+                headers=dict(**self._base_headers, **headers),
+            ),
+        )
+
+    def get(self, url, ssl=None, params=None, headers={}):
+        return _RequestContextManager(
```
dpgeorge, line 234: To simplify and reduce code, all these convenience functions could be written like this:
```py
def get(self, url, **kwargs):
    return self.request("GET", url, **kwargs)
```
Even if that's slightly different in terms of allowed args (eg `get` now allows the `data` argument) I think it's worth doing because it greatly simplifies these functions.

python-ecosys/aiohttp/aiohttp/__init__.py
```diff
@@ -0,0 +1,330 @@
+import asyncio
```
dpgeorge, line 1: Please put license and copyright info here.  The following would be enough:
```py
# MicroPython aiohttp library
# MIT license; Copyright (c) 2023 <your name>
```

python-ecosys/aiohttp/aiohttp/aiohttp_ws.py
```diff
@@ -0,0 +1,267 @@
+# adapted from https://github.com/danni/uwebsockets
```
dpgeorge, line 1: Please add license and copyright header, the same as in `__init__.py`.

dpgeorge:
> So the source code I used comes from:
> ...

Great, thanks for those details.  That all looks fine, the code is all MIT.

> The **TLDR** of the things that are implemented is: most features of https://docs.aiohttp.org/en/stable/client_quickstart.html (except Timeouts, and file uploads) and some of https://docs.aiohttp.org/en/stable/client_advanced.html (i.e. headers and `SSLContext`)

That's really good, getting the quick-start code mostly working is great.

> I think the main difference is that MicroPython implementation is missing some keyword arguments in `ClientSession` and in requests method (I may update this with the "exact" diff if necessary)

That's fine.  These things can be added in the future if needed.

dpgeorge:
Thanks for updating.  Now merged.

PR #779, opened 2023-12-21
micropython/senml: Rename senml module.

iabdalkader:
MicroPython's senml module conflicts with https://pypi.org/project/senml/ when it's installed on PC (with pip) it overwrites/gets merged with pypi's senml module, so it needs to be renamed. We originally assumed that there's no other `senml` module on pypi, but we were wrong, there's one and if it's installed before/after this module things break.. 
I've only renamed the inner package/module that gets installed, this way mainfest tools can still find it, and we don't have to change boards' manifests that freeze it, but I can rename everything. Also I've used `mp` prefix but we can use anything else (`lsenml` ? as in lightweight ?)


dpgeorge:
I'm not sure exactly what the problem is here, our packages are in separate namespaces to PyPI and so we can have the same name.  (Ideally if a package in this repo has the same name as one on PyPI then it goes in the `python-ecosys` directory and is a subset of the API, but that doesn't always need to be the case.)

In particular, what happens if someone one day creates a new PyPI package that has the same name as an existing package here, eg `aiorepl`.  Does that mean we need to go and rename it here?

The senml at https://pypi.org/project/senml/ looks quite outdated and unmaintained.  Will anyone install it?  Regardless if they do install that one or not, I'm not sure how it clashes with our `senml` module.  Can you elaborate?

iabdalkader:
> looks quite outdated and unmaintained. Will anyone install it? 

I agree, but apparently at least 2 people already did so by mistake, causing both modules to be unusable, and that's why I'm here.

> Regardless if they do install that one or not, I'm not sure how it clashes with our `senml` module. Can you elaborate?

Both https://pypi.org/project/senml/ and our newly published senml https://pypi.org/project/micropython-senml/ install the same module/directory called `senml` in site-packages, when that happens the last module installed gets merged with the existing one and overwrites some files.

> In particular, what happens if someone one day creates a new PyPI package that has the same name as an existing package here, eg `aiorepl`. Does that mean we need to go and rename it here?

If you plan on publishing more packages to PyPi, the module name needs to match the package name, which is guaranteed to be unique on PyPi. That's what we should have done with this package, `micropython-` is added to the package name when it's published ? It should also be used for the module name itself (or something else that's valid).

dpgeorge:
> apparently at least 2 people already did so by mistake, causing both modules to be unusable

IMO that needs to be fixed by documentation/tutorials/etc.  In particular it's advised to always use a Python virtual environment when installing from PyPI, and following that there shouln't be any clashes.

Also it might be a good idea to provide a `requirements.txt` with the correct PyPI packages listed so that things go smoothly for users.


> Both https://pypi.org/project/senml/ and our newly published senml https://pypi.org/project/micropython-senml/ install the same module/directory called `senml` in site-packages, when that happens the last module installed gets merged with the existing one and overwrites some files.

This is probably true for quite a few packages on PyPI.  For example you can install `serial` and `pyserial` and they will clash badly because they are both available by `import serial`.

In general that's a problem with Python packages and PyPI and we cannot solve this problem.

> If you plan on publishing more packages to PyPi, the module name needs to match the package name, which is guaranteed to be unique on PyPi.

Even if the module name matches the package name that doesn't mean the module name will be unique (see `serial` example above).  So it's impossible for us to prevent this happening with other packages in the future, someone can easily publish a new package to PyPI that uses some name that we are already using.

PR #782, opened 2024-01-03
python-ecosys/aiohttp: Fix binary data treatment.

### LINKED ISSUE #781: aiohttp converts data to string
Nice to have aiohttp, since it supports the https protocol with 1.22!! I did not see aiohttp in the 1.22 announcement, it really is useful to have async https now!

When posting data with the data= option, the data gets converted to string before sending. This encloses the data in b'' before sending. For example data=b'hello' will send str(b'hello'), i.e. "b'hello'"

What is expected: I expected data= to accept bytes or bytearray, just like urequests or CPython aiohttp, and to transmit the raw bytes without converting nor encoding/decoding. 

I am using: MicroPython v1.22.0 on 2023-12-27; Generic ESP32S3 module with Octal-SPIRAM with ESP32S3

Sample code (based on aiohttp examples):
```py
import aiohttp
import asyncio
async def main():
    do_connect()
    async with aiohttp.ClientSession("http://httpbin.org") as session:
        async with session.post("/post",data=b'hello world') as resp:
            assert resp.status == 200
            rpost = await resp.text()
            print(f"POST: {rpost}")

asyncio.run(main())
```
The output is
```
POST: {
  "args": {}, 
  "data": "b'hello world'", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Content-Length": "14", 
    "Host": "httpbin.org", 
    "User-Agent": "compat", 
    "X-Amzn-Trace-Id": "Root=1-6594aa03-4f076b3a3480ef2659dfc035"
  }, 
  "json": null, 
  "origin": "181.43.38.11", 
  "url": "http://httpbin.org/post"
}
```

Carglglz:
Fixes #781, #784, #788

### DIFF
--- a/python-ecosys/aiohttp/aiohttp/__init__.py
+++ b/python-ecosys/aiohttp/aiohttp/__init__.py
@@ -38,10 +38,10 @@ class ClientResponse:
         return self._decode(await self.content.read(sz))
 
     async def text(self, encoding="utf-8"):
-        return (await self.read(sz=-1)).decode(encoding)
+        return (await self.read(int(self.headers.get("Content-Length", -1)))).decode(encoding)
 
     async def json(self):
-        return _json.loads(await self.read())
+        return _json.loads(await self.read(int(self.headers.get("Content-Length", -1))))
 
     def __repr__(self):
         return "<ClientResponse %d %s>" % (self.status, self.headers)
@@ -121,7 +121,7 @@ class ClientSession:
                     if b"chunked" in line:
                         chunked = True
                 elif line.startswith(b"Location:"):
-                    url = line.rstrip().split(None, 1)[1].decode("latin-1")
+                    url = line.rstrip().split(None, 1)[1].decode()
 
             if 301 <= status <= 303:
                 redir_cnt += 1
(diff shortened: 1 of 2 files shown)


bixb922 (changes requested):
I'd suggest not to specify "latin-1" as encoding. This parameter is AFAIK ignored for .encode(), and the encoding for JSON has to be UTF-8, which is the default for .encode()

python-ecosys/aiohttp/aiohttp/__init__.py
```diff
@@ -38,10 +38,10 @@ async def read(self, sz=-1):
         return self._decode(await self.content.read(sz))
 
     async def text(self, encoding="utf-8"):
-        return (await self.read(sz=-1)).decode(encoding)
+        return (await self.read(sz=int(self.headers.get("Content-Length", -1)))).decode(encoding)
```
dpgeorge, line 41: I suggest just using positional arguments (not doing `sz=`) because it will make slightly smaller and slightly faster code.

python-ecosys/aiohttp/aiohttp/__init__.py
```diff
@@ -38,10 +38,10 @@ async def read(self, sz=-1):
...
 
     async def json(self):
-        return _json.loads(await self.read())
+        return _json.loads(await self.read(sz=int(self.headers.get("Content-Length", -1))))
```
dpgeorge, line 44: Suggest using positional arguments for `read()`.

dpgeorge:
Thanks for updating, now merged.

PR #783, opened 2024-01-03
typofix in l2cap_file_server.py

n0npax:
fix a typo. os has `listdir`, not `ilistdir`. `i` prefix seems to be wrong.


sosi-deadeye:
It's not a typo, and why does the unittest not fail?

There is a huge difference between [os.listdir](https://docs.micropython.org/en/latest/library/os.html#os.ilistdir) and [os.listdir](https://docs.micropython.org/en/latest/library/os.html#os.listdir).

dpgeorge:
As mentioned, this is not a typo.  MicroPython has a custom `os.ilistdir()` function.

PR #789, opened 2024-01-20
python-stdlib/hmac/hmac: Fix for hashlib.

Pharkie:
hashlib has no .new method?


dpgeorge:
Thanks for the fix.

Actually the version of `hashlib` in this repository does provide a `new()` function, but the built-in library does not.  So this is a good fix.

PR #793, opened 2024-01-30
ssl: restructure micropython interface in a tls module

felixdoerre:
This changes correspond to [micropython/12259](https://github.com/micropython/micropython/pull/12259).

### DIFF
--- a/micropython/umqtt.simple/umqtt/simple.py
+++ b/micropython/umqtt.simple/umqtt/simple.py
@@ -16,8 +16,7 @@ class MQTTClient:
         user=None,
         password=None,
         keepalive=0,
-        ssl=False,
-        ssl_params={},
+        ssl=None,
     ):
         if port == 0:
             port = 8883 if ssl else 1883
@@ -26,7 +25,6 @@ class MQTTClient:
         self.server = server
         self.port = port
         self.ssl = ssl
-        self.ssl_params = ssl_params
         self.pid = 0
         self.cb = None
         self.user = user
@@ -67,9 +65,7 @@ class MQTTClient:
         addr = socket.getaddrinfo(self.server, self.port)[0][-1]
         self.sock.connect(addr)
         if self.ssl:
-            import ussl
-
-            self.sock = ussl.wrap_socket(self.sock, **self.ssl_params)
+            self.sock = self.ssl.wrap_socket(self.sock, server_hostname=self.server)
         premsg = bytearray(b"\x10\0\0\0\0\0")
         msg = bytearray(b"\x04MQTT\x04\x02\0\0")
 
--- a/micropython/umqtt.simple/umqtt/simple.py
+++ b/micropython/umqtt.simple/umqtt/simple.py
@@ -75,7 +75,7 @@ class MQTTClient:
 
         sz = 10 + 2 + len(self.client_id)
         msg[6] = clean_session << 1
-        if self.user is not None:
+        if self.user:
             sz += 2 + len(self.user) + 2 + len(self.pswd)
             msg[6] |= 0xC0
         if self.keepalive:
@@ -101,7 +101,7 @@ class MQTTClient:
         if self.lw_topic:
             self._send_str(self.lw_topic)
             self._send_str(self.lw_msg)
-        if self.user is not None:
+        if self.user:
             self._send_str(self.user)
             self._send_str(self.pswd)
         resp = self.sock.read(4)
--- a/micropython/urllib.urequest/urllib/urequest.py
+++ b/micropython/urllib.urequest/urllib/urequest.py
@@ -12,7 +12,7 @@ def urlopen(url, data=None, method="GET"):
     if proto == "http:":
         port = 80
     elif proto == "https:":
-        import ussl
+        import tls
 
         port = 443
     else:
@@ -29,7 +29,9 @@ def urlopen(url, data=None, method="GET"):
     try:
         s.connect(ai[-1])
         if proto == "https:":
-            s = ussl.wrap_socket(s, server_hostname=host)
+            context = tls.SSLContext(tls.PROTOCOL_TLS_CLIENT)
+            context.verify_mode = tls.CERT_NONE
+            s = context.wrap_socket(s, server_hostname=host)
 
         s.write(method)
         s.write(b" /")
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -63,7 +63,7 @@ def request(
     if proto == "http:":
         port = 80
     elif proto == "https:":
-        import ussl
+        import tls
 
         port = 443
     else:
@@ -90,7 +90,9 @@ def request(
     try:
         s.connect(ai[-1])
         if proto == "https:":
-            s = ussl.wrap_socket(s, server_hostname=host)
+            context = tls.SSLContext(tls.PROTOCOL_TLS_CLIENT)
+            context.verify_mode = tls.CERT_NONE
+            s = context.wrap_socket(s, server_hostname=host)
         s.write(b"%s /%s HTTP/1.0\r\n" % (method, path))
         if "Host" not in headers:
             s.write(b"Host: %s\r\n" % host)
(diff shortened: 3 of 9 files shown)


python-stdlib/ssl/ssl.py
```diff
@@ -1,36 +1,72 @@
...
-        kw["cert_reqs"] = cert_reqs
-    if ca_certs is not None:
-        kw["ca_certs"] = ca_certs
-    if server_hostname is not None:
-        kw["server_hostname"] = server_hostname
-    return _ussl.wrap_socket(sock, **kw)
+    con = SSLContext(PROTOCOL_TLS_SERVER if server_side else PROTOCOL_TLS_CLIENT)
+    if cert is not None or key is not None:
+        con.load_cert_chain(cert, key)
+    if cadata is not None:
```
dpgeorge, line 64: This would be simpler and smaller code as `if cert or key` and `if cadata`.

python-stdlib/ssl/ssl.py
```diff
@@ -1,36 +1,72 @@
...
+        if isinstance(certfile, str):
+            with open(certfile, "rb") as f:
+                certfile = f.read()
+        if isinstance(keyfile, str):
+            with open(keyfile, "rb") as f:
+                keyfile = f.read()
+        self._context.load_cert_chain(certfile, keyfile)
+
+    def load_verify_locations(self, cafile=None, cadata=None):
+        if cafile is not None:
```
dpgeorge, line 35: Simpler as `if cafile`.

micropython/umqtt.simple/umqtt/simple.py
```diff
@@ -16,17 +16,15 @@ def __init__(
         user=None,
         password=None,
         keepalive=0,
-        ssl=False,
-        ssl_params={},
+        ssl=None,
     ):
         if port == 0:
-            port = 8883 if ssl else 1883
+            port = 8883 if ssl is not None else 1883
```
dpgeorge, line 22: Can this just continue to use `if ssl`?

micropython/urllib.urequest/urllib/urequest.py
```diff
@@ -12,7 +12,7 @@ def urlopen(url, data=None, method="GET"):
     if proto == "http:":
         port = 80
     elif proto == "https:":
-        import ussl
+        import ssl
```
dpgeorge, line 15: I suggest using `import tls` directly.  Because these libraries don't need to be backwards compatible.

micropython/umqtt.simple/umqtt/simple.py
```diff
@@ -67,15 +65,13 @@ def connect(self, clean_session=True):
...
-
-            self.sock = ussl.wrap_socket(self.sock, **self.ssl_params)
+            self.sock = self.ssl.wrap_socket(self.sock, server_hostname=self.server)
         premsg = bytearray(b"\x10\0\0\0\0\0")
         msg = bytearray(b"\x04MQTT\x04\x02\0\0")
 
         sz = 10 + 2 + len(self.client_id)
         msg[6] = clean_session << 1
-        if self.user is not None:
+        if self.user:
```
dpgeorge, line 74: This should not change.  The existing code supports passing in an empty string for the `user`.

micropython/umqtt.simple/umqtt/simple.py
```diff
@@ -101,7 +97,7 @@ def connect(self, clean_session=True):
         if self.lw_topic:
             self._send_str(self.lw_topic)
             self._send_str(self.lw_msg)
-        if self.user is not None:
```
dpgeorge, line 104: This should not change.

micropython/umqtt.simple/umqtt/simple.py
felixdoerre: Initially, I had `is not None` for all checks, as that was the behavior of the C-code. Doing `if cadata` changes that behavior in the way, that passing `cadata = ""` will now not produce an error anymore, but will configure the field empty.

I've looked around a bit for MQTT implementations that expect an empty username (instead of the field missing), and I couldn't find any. Do you know of any implementation that expects passing an empty string for `user`?

So I just cleaned up more code, but I can undo that again.

dpgeorge: > I've looked around a bit for MQTT implementations that expect an empty username (instead of the field missing), and I couldn't find any.

OK, then let's keep your change.

For reference the commit that added this line is fcec7f7c1d3485474596a0db38a83e7b9ce5870b

python-ecosys/requests/requests/__init__.py
```diff
@@ -90,7 +90,7 @@ def request(
     try:
         s.connect(ai[-1])
         if proto == "https:":
-            s = ussl.wrap_socket(s, server_hostname=host)
+            s = tls.SSLContext(tls.PROTOCOL_TLS_CLIENT).wrap_socket(s, server_hostname=host)
```
dpgeorge, line 93: Testing this change, it no longer works because `verify_mode` defaults to `CERT_REQUIRED`... I guess that's secure, but we really need to make this work like it did before.  So this needs to do `ctx.verify_mode = CERT_NONE` before wrapping the socket.

Similarly with the `umqtt` and `urllib` changes in this PR.

dpgeorge:
Please can you change the `module()` call in `ssl/manifest.py` to:
```py
module("ssl.py", opt=3)
```

That well help keep the code size down.

python-ecosys/requests/requests/__init__.py
felixdoerre: I'll change the behavior for `requests` and `urllib`, as we don't ship a CA list by default we can't enable ssl verification without the user choosing it. (Previously, using the `ssl` library, with the legacy `ssl.wrap_socket` did disable verification).  For `umqtt`, I've adjusted the constructor to not take a boolean parameter for `ssl` anymore, but an sslcontext instead. So if someone wants to enable insecure ssl, they need to disable verification in the ssl context themselves (or load ca certificates or install a cert callback).

Ideally we should have a way to do that for `urllib` and `requests` as well. Looking at the documentation of their python counterpart, `urllib` has a `capath` parameter, and `requests` has a `verify` parameter, to point to a ca file. I'm not sure how strictly to `cpython` subset we adhere here, but I would like to implement parameters to allow passing a ca certificate for both modules.

dpgeorge: Oh, yes, `umqtt` now passes in the `SSLContext`, so no need to change the `verify_mode` for that.



> I'm not sure how strictly to `cpython` subset we adhere here, but I would like to implement parameters to allow passing a ca certificate for both modules.

Yes, that would be a good addition, allowing to pass in certs.  Ideally it would follow the CPython way.  Of course that's for a separate PR!

dpgeorge:
Thanks for this, now merged.

micropython/umqtt.simple/umqtt/simple.py
```diff
@@ -16,8 +16,7 @@ def __init__(
         user=None,
         password=None,
         keepalive=0,
-        ssl=False,
-        ssl_params={},
```
Carglglz, line 20: This seems now like a "breaking" change, one needs to create and pass a `SSLContext`, see e.g. https://github.com/orgs/micropython/discussions/13624, I think this should be something like 
```py

if self.ssl:
             if hasattr(self.ssl, "wrap_socket"):
                 self.sock = self.ssl.wrap_socket(self.sock, **self.ssl_params)
             else:
                 import ssl as _ssl

                 self.sock = _ssl.wrap_socket(self.sock, **self.ssl_params)
```

felixdoerre: Yes, I was not aware that breaking changes here are not possible.  I would argue that this breaking change is good here.

dpgeorge: I think this breaking change is a fair change to make.  We need to improve and make progress on things and sometimes it's not practical to retain backwards compatibility on everything.

In this case using the new library with old user code will raise an exception if the `ssl` argument is used.  So it's easy for the user to know that things need to be changed.

And the new scheme matches better how SSL works in `asyncio`, passing in an `SSLContext`.

In the end a user can just keep using an old version of `umqtt` if needed.

Carglglz: Then document this somewhere in the `umqtt`'s README or next release info should be fine 👍🏼

Issue #797, opened 2024-02-03
tarfile-write issue with file/folder permissions


ubidefeo:
### Context
I've been playing with `tarfile` for a few days and yesterday started trying to create my own archives using `tarfile-write` installed via `mip`.

Adding a folder to it works recursively, but transferring the file to the PC via `mpremote` and expanding it yields all files and folders with `--- --- ---` permissions (0x000).

### Discovery
I investigated and found the source of the issue is related to how the `TarInfo` object [is composed in these lines](https://github.com/micropython/micropython-lib/blob/7cdf70881519c73667efbc4a61a04d9c1a49babb/python-stdlib/tarfile-write/tarfile/write.py#L102C9-L106C32)

in particular on line 102 we find `mode` to be set to the `os.stat(FILE)[0]`

```
tarinfo.mode = stat[0]
```

`stat[0]` is either `0x8000` (file) or `0x4000` (folder).
And the subsequently applied masking yields `0x000` as permissions set while adding it to the archive.
I think that mode should be safely set to `0x1a4 > 644 > rw- r-- r--`

I have tested by manually composing a TarInfo object to be handed to `addfile()` rather than patching `add()`

### How to reproduce:
```python
import tarfile
import os

os.mkdir('test_folder')
f = open('test_folder/test_file_01.txt', 'w')
f.write('file 01')
f.close()

os.mkdir('test_folder/sub_folder')

f = open('test_folder/sub_folder/test_file_02.txt', 'w')
f.write('file 02')
f.close()

archive = tarfile.TarFile('test_archive.tar', 'w')
archive.add('test_folder')
archive.close()

```

Copy the archive from the board to the PC via `mpremote cp :test_archive.tar test_archive.tar`

Expand the archive and verify the permissions for **test_folder**.
From Mac os I was not able to `chmod 644 test_folder`, but using the UI I could unrestrict access.
![CleanShot 2024-02-03 at 21 50 23@2x](https://github.com/micropython/micropython-lib/assets/1894487/f8b0541d-f7cb-4a81-9bc3-cd1f91fe3177)

Looking at the folder content you can verify the same permission issue happens with `sub_folder` and contained file(s)
![CleanShot 2024-02-03 at 21 51 27@2x](https://github.com/micropython/micropython-lib/assets/1894487/9a079fd0-51e3-406a-9465-169b14e97442)

Environment:
I run a custom build for ESP32-S3, but this issue is not related to changes I have applied
[...]

PR #799, opened 2024-02-05
python-stdlib/tarfile: Fixed permissions when adding to archive.

ubidefeo:
This change fixes issue https://github.com/micropython/micropython-lib/issues/797 which I opened a couple of days ago.
More context to be found there.

This is my first PR in which I tried to be neater, hope it'll get reviewed by someone :)

thanks
ubi

### DIFF
--- a/python-stdlib/tarfile-write/tarfile/write.py
+++ b/python-stdlib/tarfile-write/tarfile/write.py
@@ -67,7 +67,7 @@ def addfile(self, tarinfo, fileobj=None):
             name += "/"
     hdr = uctypes.struct(uctypes.addressof(buf), _TAR_HEADER, uctypes.LITTLE_ENDIAN)
     hdr.name[:] = name.encode("utf-8")[:100]
-    hdr.mode[:] = b"%07o\0" % (tarinfo.mode & 0o7777)
+    hdr.mode[:] = b"%07o\0" % ((0o755 if tarinfo.isdir() else 0o644) & 0o7777)
     hdr.uid[:] = b"%07o\0" % tarinfo.uid
     hdr.gid[:] = b"%07o\0" % tarinfo.gid
     hdr.size[:] = b"%011o\0" % size
@@ -96,9 +96,10 @@ def addfile(self, tarinfo, fileobj=None):
 def add(self, name, recursive=True):
     from . import TarInfo
 
-    tarinfo = TarInfo(name)
     try:
         stat = os.stat(name)
+        res_name = (name + '/') if (stat[0] & 0xf000) == 0x4000 else name
+        tarinfo = TarInfo(res_name)
         tarinfo.mode = stat[0]
         tarinfo.uid = stat[4]
         tarinfo.gid = stat[5]
--- a/python-stdlib/tarfile-write/manifest.py
+++ b/python-stdlib/tarfile-write/manifest.py
@@ -1,4 +1,4 @@
-metadata(description="Adds write (create/append) support to tarfile.", version="0.1.1")
+metadata(description="Adds write (create/append) support to tarfile.", version="0.1.2")
 
 require("tarfile")
 package("tarfile")


python-stdlib/tarfile-write/tarfile/write.py
```diff
@@ -67,7 +67,7 @@ def addfile(self, tarinfo, fileobj=None):
             name += "/"
     hdr = uctypes.struct(uctypes.addressof(buf), _TAR_HEADER, uctypes.LITTLE_ENDIAN)
     hdr.name[:] = name.encode("utf-8")[:100]
-    hdr.mode[:] = b"%07o\0" % (tarinfo.mode & 0o7777)
+    hdr.mode[:] = b"%07o\0" % (0o0755 & 0o7777)
```
projectgus, line 70: I think this change can be dropped, yeah? Otherwise all tar contents will have RWX permissions.

projectgus (commented):
One small request, but this change LGTM. Thanks @ubidefeo!

python-stdlib/tarfile-write/tarfile/write.py
ubidefeo: unfortunately `tarinfo.mode` takes the `0x4000` or `0x8000` depending on wether the resource is a file or a folder, turning the file permission matrix into `0o4000` or `0o8000`, hence no permissions from anyone.
The MicroPython file system does not return permissions in `stat[0]`, as it happens for python on a PC.
This would lead to archives that require multiple permissions changes from the user, especially when compressing folder structures

projectgus: Oh, right, gotcha. I should have looked up the numeric values of `_S_IFDIR` and `_S_IF_REG` .

What about `(0o755 if tarinfo.mode == S_IFDIR else 0o644)` for this argument, will that give reasonable results without setting execute on normal files?

ubidefeo: @projectgus I initially tried setting `0o644` for everything and it didn't work, but will try this one and update the PR :)

Thank you for taking the time to follow up

ubidefeo: @projectgus , now I have a problem: `S_IFDIR` is not recognised within `write.py`.
I guess it's because it's a private member of `tarfile` defined in `__init__.py`

I'm not so deeply skilled in Python to know how to get around that, learning as I go :D 

```
>>> tw.add('test_folder')
  File "/lib/tarfile/write.py", line 115, in add
  File "/lib/tarfile/write.py", line 70, in addfile
NameError: name 'S_IFDIR' isn't defined
```

I found out that this works, though
`(0o755 if tarinfo.isdir() else 0o644)`

would it be an acceptable solution?

projectgus: @ubidefeo That version looks even better to me actually, good idea!

projectgus (approved):

python-stdlib/tarfile/tarfile/__init__.py
```diff
@@ -63,7 +64,7 @@ def skip(self):
 class TarInfo:
     def __init__(self, name=""):
         self.name = name
-        self.mode = _S_IFDIR if self.name[-1] == "/" else _S_IFREG
+        self.mode = _S_IFDIR if os.stat(name)[0] == 0x4000 else _S_IFREG
```
dpgeorge, line 67: What happens if `name` doesn't exist on disk?  In the case that you are extracting from a tar file, the files most likely do not exist on disk, so this will now raise an exception because the file isn't found.

ubidefeo: thank you @dpgeorge 
I didn't think of that at all 🤔 

would you be ok if I added a `try/except` condition?
I like the flexibility offered by not having to rely on what the user hands in.

In my code I always test if the fs item exists, but I neglected to do it in this case

ubidefeo: I dug deeper and probably have to handle archive and expand in two different ways.
During expansion it does need to rely on the name coming from the item in the archive 🤔

ubidefeo: @dpgeorge @projectgus 
please take a look at my last commit, I think I found a compromise 😄

python-stdlib/tarfile/tarfile/__init__.py
```diff
@@ -1,6 +1,7 @@
 """Subset of cpython tarfile class methods needed to decode tar files."""
 
 import uctypes
+import os
```
dpgeorge, line 4: This import shouldn't be needed anymore.

ubidefeo: true!
fixing it now

ubidefeo: done!
I really appreciate your time, @dpgeorge 🙏🏼

dpgeorge:
Thanks for updating.  Rebased and merged.

PR #806, opened 2024-02-15
logging.py: CPython-compatible logging improvements.

ned-pcs:
This PR allows for logging handlers to be added to the root logger and then used by non-root loggers that don't have their own handlers.

It replaces #750 (in which I forgot a period at the end of a commit message title).

It also adds the (CPython-compatible) handlers argument to `logging.basicConfig()` to ease this initialization:
```python
    import logging
    sh = logging.StreamHandler()
    fh = logging.FileHandler("my.log", mode="a")
    logging.basicConfig(handlers=[sh, fh])

    root_logger = logging.getLogger() # uses sh and fh
    another_logger = logging.getLogger("another") # inherits handlers
```

It also adds the `Logger.removeHandler()` method and avoids repeated handler addition.

It also adds the `flush()` method to `StreamHandler` and its subclasses.

It also correctly calls the superclass constructor from the `StreamHandler` constructor and uses a default formatter if a `Handler` has none set (as in PR #710).


ned-pcs:
build failure is due to an unrelated `manifest.py` issue.

projectgus:
@ned-pcs If you rebase on master then the build should be fixed now.

python-stdlib/logging/logging.py
```diff
@@ -43,25 +44,40 @@ def __init__(self, level=NOTSET):
...
 
 
 class StreamHandler(Handler):
     def __init__(self, stream=None):
+        super().__init__()
         self.stream = _stream if stream is None else stream
         self.terminator = "\n"
 
-    def close(self):
+    def flush(self):
```
dpgeorge, line 80: Should there still be a `close()` method which just calls `self.flush()`?

ned-pcs: Good point. Cpython does not have a `StreamHandler.close()` method, and so the inherited `close()` method doesn't do any output.

PR #809, opened 2024-02-22
json: Move to unix-ffi.

jimmo:
It requires the unix pcre-based re module.

See https://github.com/orgs/micropython/discussions/13570


unix-ffi/json/manifest.py
```diff
@@ -0,0 +1,4 @@
+metadata(version="0.2.0")
+
+require("re", unix_ffi=True)
```
dpgeorge, line 3: The `unix_ffi` argument is no longer supported.

(I wonder how the CI even passes on this PR??)

dpgeorge: CI passes because `unix_ffi` is swallowed by the `**kwargs` and ignored.

Issue #754, opened 2023-10-21
umqtt.simple got OSError: -1 after publishing 65535 packets message on esp32


mzhboy:
platform: `MicroPython v1.21.0 on 2023-10-06; Generic ESP32 module`
broker: mosquitto
description:
got `OSError: -1`  after sending 65535 packets 

Result:

> pkcnt 65000
> Traceback (most recent call last):
>   File "<stdin>", line 47, in <module>
>   File "<stdin>", line 40, in publish_test
>   File "umqtt/simple.py", line 144, in publish
>   File "umqtt/simple.py", line 184, in wait_msg
> OSError: -1

code:

```python
from umqtt.simple import MQTTClient
import time
import ujson as json
from machine import unique_id
import sys, os

def str_current_time():
    tm_ns = time.time_ns()
    tm_rem_ms = (tm_ns % 1_000_000_000) // 1_000_000
    tm = time.localtime(tm_ns//1_000_000_000)
    str_tm = str(tm[3])+':'+str(tm[4])+':'+str(tm[5])+'.'+ str(tm_rem_ms)
    return str_tm
   

id = unique_id() #machine.unique_id()
id_str = '{:02x}{:02x}{:02x}{:02x}'.format(id[0], id[1], id[2], id[3])

mq_id = id_str
mq_machine = os.uname().machine.split()[0]
mq_server = '192.168.12.25'
mq_user='esp32-srv'
mq_pass='123456'
mq_topic= b'esp/test'
mq_message = {'id':mq_id, 'machine': mq_machine, 'time': str_current_time(), 'hello': 'hello'}
mqClient = MQTTClient(mq_id, mq_server, port=1883, user=mq_user, password=mq_pass, keepalive=60*60*12)
mqClient.connect(clean_session=True) # only support clean_sesson=True
mqClient.publish(b'esp/hello', json.dumps(mq_message), qos=1)

pkcnt = 0x00_00_00_01
pkcnt_max = 999_999

def publish_test():
    global pkcnt, pkcnt_max
    pkcnt += 1
    if pkcnt >= pkcnt_max :
        pkcnt = 1
    tm = time.time()
    message = {'id':mq_id, 'machine': mq_machine, 'time': tm, 'current': 1, 'ctrl': 1, 'pkcnt':pkcnt}
    mq_message = json.dumps(message)
    mqClient.publish(mq_topic, mq_message, qos=1)
    if pkcnt % 1000 == 0:
[...]

PR #812, opened 2024-02-28
umqtt.simple: Fix MQTTClient publish for connection disruption.

khrystynaO:
Method now handles OSError: -1 after many packets, attempting reconnection.

Issue #754

Modified the MQTTClient publish method to gracefully handle connection disruptions. Previously, the method encountered an OSError: -1 after sending a large number of packets, leading to a disconnection from the server. The updated method now includes error handling to detect this situation and attempts to reconnect before retrying the operation. This enhancement ensures that the client can recover from connection disruptions seamlessly, maintaining stable communication with the server.


dpgeorge:
This error handling is already handled by `umqtt.robust`, it overrides `.publish()` with a try-except.

PR #814, opened 2024-02-29
webrepl: Changes for more webrepl features while making it smaller.

felixdoerre:
This change:
- Moves the password checking to python
- Removes the special file transfer protocol (This changes allows to extend `mpremote` to support webrepl just like a serial port)
- Moves the REPL data to websocket binary packages

The change should be compatible with the current micropython version, however a new webrepl client needs to be deployed under https://micropython.org/webrepl/

Currently, to help you testing, I've adjusted the default URL to https://felix.dogcraft.de/webrepl/, where I host the corresponding draft "new" webrepl client. I've pushed the modified js code here, if you want to take a look: https://github.com/felixdoerre/webreplv2. I'm not completely sure, if I should open a PR in https://github.com/micropython/webrepl, as this seems to be the authoritative source of the client, but the repo seems to be abandoned for some time now.

The new webrepl client currently features (See also [#13540](https://github.com/micropython/micropython/issues/13540)):
- A file browser, allowing directory listings, downloads and uploads. This uses the same injected code as `mpremote`.
- An implementation of the `mount`-feature from `mpremote` that works with the same injected code. It allows mounting files from the browser's local storage (which can host some python scripts that one wants to bring for debugging), or from a local folder, that is drag-and-dropped into the webrepl (read-only), this is useful for rapid development/testing.
- An implementation of mip with a package browser to list and download packages directly from https://micropython.org/pi/

Currently the "mount" feature has to be activated "manually" (by calling `hookme()` from the developer tools, and then executing `__mount()` in the repl manually). Also the UI for the file browser might not be super intuitive yet (the + for upload and mounting a local folder might be too small and not clear enough). So I'm hoping for UI improvement suggestions.

If you want, I can split the `mount` and the `mip` implementation into separate PRs, but the file-browser and this PR are dependent on each other.

With this change the `_webrepl` module from micropython is not needed anymore and can be removed.

### DIFF
--- a/micropython/net/webrepl/webrepl.py
+++ b/micropython/net/webrepl/webrepl.py
@@ -7,56 +7,73 @@ import socket
 import sys
 import websocket
 import io
+from micropython import const
+from legacy_file_transfer import LegacyFileTransfer
 
 listen_s = None
 client_s = None
 
 DEBUG = 0
 
-_DEFAULT_STATIC_HOST = const("https://felix.dogcraft.de/webrepl/")
+_DEFAULT_STATIC_HOST = const("https://micropython.org/webrepl/")
 _WELCOME_PROMPT = const("\r\nWebREPL connected\r\n>>> ")
 static_host = _DEFAULT_STATIC_HOST
 webrepl_pass = None
 
+legacy = LegacyFileTransfer()
+
+
 class WebreplWrapper(io.IOBase):
     def __init__(self, sock):
         self.sock = sock
-        self.sock.ioctl(9, 2)
+        self.sock.ioctl(9, 1 if legacy else 2)
         if webrepl_pass is not None:
             self.pw = bytearray(16)
             self.pwPos = 0
             self.sock.write("Password: ")
         else:
             self.pw = None
-            self.sock.write(_WELCOME_PROMPT);
+            self.sock.write(_WELCOME_PROMPT)
 
     def readinto(self, buf):
         if self.pw is not None:
-            buf = bytearray(1)
+            buf1 = bytearray(1)
             while True:
-                l = self.sock.readinto(buf)
+                l = self.sock.readinto(buf1)
                 if l is None:
                     continue
                 if l <= 0:
                     return l
-                if buf[0] == 10 or buf[0] == 13:
+                if buf1[0] == 10 or buf1[0] == 13:
                     print("Authenticating with:")
-                    print(self.pw[0:self.pwPos])
-                    if bytes(self.pw[0:self.pwPos]) == webrepl_pass:
+                    print(self.pw[0 : self.pwPos])
+                    if bytes(self.pw[0 : self.pwPos]) == webrepl_pass:
                         self.pw = None
                         del self.pwPos
                         self.sock.write(_WELCOME_PROMPT)
                         break
                     else:
-                        print(bytes(self.pw[0:self.pwPos]))
+                        print(bytes(self.pw[0 : self.pwPos]))
                         print(webrepl_pass)
                         self.sock.write("\r\nAccess denied\r\n")
                         return 0
                 else:
                     if self.pwPos < len(self.pw):
-                        self.pw[self.pwPos] = buf[0]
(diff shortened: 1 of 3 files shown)


Carglglz:
@felixdoerre this could be a good opportunity
 to add SSL support to WebREPL using `SSLContext` see https://github.com/micropython/micropython/pull/5611#issuecomment-1314686679 👍🏼

felixdoerre:
I've pushed a few adjustments and an (optional) bonus commit to allow specifying an ssl context.

micropython/net/webrepl/webrepl.py
```diff
@@ -7,19 +7,71 @@
...
-import _webrepl
+import io
 
 listen_s = None
 client_s = None
 
 DEBUG = 0
 
-_DEFAULT_STATIC_HOST = const("https://micropython.org/webrepl/")
+_DEFAULT_STATIC_HOST = const("https://felix.dogcraft.de/webrepl/")
```
dpgeorge, line 17: To maintain the old webrepl webpage for older devices, does it make sense to change this URL to something like `https://micropython.org/webrepl2/`?  Or is it enough to just specify different JS code as you've done below with the line `webreplv2_content.js`?  Ie, can both the old and new web clients exist at this same URL?

felixdoerre: For people who open the webrepl web-client like I do, by navigating to the micropython-device and then being served [this webpage](https://github.com/micropython/micropython-lib/blob/master/micropython/net/webrepl/webrepl.py#L87), the micropython device can select the correct file and from there the correct dependencies. These clients can choose the correct files and exist at the same URL.

It does make a difference for people who navigate to `https://micropython.org/webrepl/` manually and then enter their device's IP address in the text field. `https://micropython.org/webrepl/` can only serve one version. How many people do that? I am not directly aware of documentation that sends people directly to that URL. Do you have some access statistics of how often `https://micropython.org/webrepl/`  is accessed vs `https://micropython.org/webrepl/webrepl_content.js`? Also those people who manually navigate to the website would need to be aware that there are two different versions and then choose the correct one.

I would argue that we recommend people to navigate to their device directly so it can choose the correct version and don't recommend to navigate to the page on `micropython.org` manually.

The main "pro"-argument for keeping the same base url would be, that we can "eventually" switch the version served there to show webreplv2, so we start catching those people who navigate to the URL (from wherever they might have found the link).

So that's the rationale for why I chose this variant, but I have no strong preference at all for either.

micropython/net/webrepl/webrepl.py
```diff
@@ -110,11 +159,16 @@ def setup_conn(port, accept_handler):
 
 
 def accept_conn(listen_sock):
-    global client_s
+    global client_s, webrepl_ssl_context
     cl, remote_addr = listen_sock.accept()
+    req = cl.makefile("rwb", 0)
```
dpgeorge, line 164: Is this `makefile` necessary?  On bare-metal ports it just returns `cl`.  It's really only there for CPython compatibility, and I don't think this code here needs to be compatible with CPython.

micropython/net/webrepl/webrepl.py
dpgeorge: OK, let's keep it as the same URL.

micropython/net/webrepl/webrepl.py
felixdoerre: I've moved the code here from the old line 22, and yes it seems this is a no-op. I'll test this code with `makefile` removed.

micropython/net/webrepl/webrepl.py
```diff
@@ -146,11 +200,12 @@ def stop():
         listen_s.close()
 
 
-def start(port=8266, password=None, accept_handler=accept_conn):
-    global static_host
+def start(port=8266, password=None, ssl_context=None, accept_handler=accept_conn):
```
dpgeorge, line 203: `start_foreground()` needs to be changed to work with this new signature.

PR #818, opened 2024-03-04
Add functools.total_ordering

winni2k:
As part of my journey to port dataclasses to micropython-lib/python-stdlib (see https://github.com/orgs/micropython/discussions/13741#discussioncomment-8579351), I needed to port `functools.total_ordering` first.

I have tried to follow all the contribution guidelines, but I have not tried to optimize code size as I am completely new to micropython and would not know where to start.

This code (and test code) is copied from CPython v3.7.17 in its entirety. The PSF license appears to allow for this, but please do correct me if I am wrong.

### DIFF
--- a/python-stdlib/functools/Dockerfile
+++ b/python-stdlib/functools/Dockerfile
@@ -1,0 +1,6 @@
+FROM micropython/unix:v1.22
+
+RUN micropython -m mip install unittest
+RUN micropython -m mip install unittest-discover
+
+
--- a/python-stdlib/functools/functools/total_ordering.py
+++ b/python-stdlib/functools/functools/total_ordering.py
@@ -1,0 +1,133 @@
+"""total_ordering
+
+This code was extracted from the CPython v3.7.17 implementation of Lib/functools.py
+
+Written by Nick Coghlan <ncoghlan at gmail.com>,
+Raymond Hettinger <python at rcn.com>,
+and Łukasz Langa <lukasz at langa.pl>.
+  Copyright (C) 2006-2013 Python Software Foundation.
+See C source code for _functools credits/copyright
+
+This code is distributed under the Python Software License.
+"""
+
+
+################################################################################
+### total_ordering class decorator
+################################################################################
+
+# The total ordering functions all invoke the root magic method directly
+# rather than using the corresponding operator.  This avoids possible
+# infinite recursion that could occur when the operator dispatch logic
+# detects a NotImplemented result and then calls a reflected method.
+
+def _gt_from_lt(self, other, NotImplemented=NotImplemented):
+    'Return a > b.  Computed by @total_ordering from (not a < b) and (a != b).'
+    op_result = self.__lt__(other)
+    if op_result is NotImplemented:
+        return op_result
+    return not op_result and self != other
+
+def _le_from_lt(self, other, NotImplemented=NotImplemented):
+    'Return a <= b.  Computed by @total_ordering from (a < b) or (a == b).'
+    op_result = self.__lt__(other)
+    return op_result or self == other
+
+def _ge_from_lt(self, other, NotImplemented=NotImplemented):
+    'Return a >= b.  Computed by @total_ordering from (not a < b).'
+    op_result = self.__lt__(other)
+    if op_result is NotImplemented:
+        return op_result
+    return not op_result
+
+def _ge_from_le(self, other, NotImplemented=NotImplemented):
+    'Return a >= b.  Computed by @total_ordering from (not a <= b) or (a == b).'
+    op_result = self.__le__(other)
+    if op_result is NotImplemented:
+        return op_result
+    return not op_result or self == other
+
+def _lt_from_le(self, other, NotImplemented=NotImplemented):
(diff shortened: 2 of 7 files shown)


winni2k:
Any chance you could review this PR @mattytrentini ?

mattytrentini:
> Any chance you could review this PR @mattytrentini ?

Sorry, been trying to find time to do so!

Perhaps consider adding a `README.md` explaining how to run the tests? Ideally in a container but even if it's just a description it would be helpful (I mean, it looks like straightforward `unittest` code but it should be clarified). Eventually we'd like to create Github Actions to run them continuously so if we have clear instructions it'll make that task easier. 

It's also important to spell out that this code is licensed under the Python license. See the root [LICENSE](https://github.com/micropython/micropython-lib/blob/master/LICENSE) file that mentions MIT "unless explicitly stated otherwise". I _think_ an entry in the README will suffice. 

Otherwise, the code looks good! I'm curious about the failing unit test but I'll have to look into that another time...

(PS thanks for the contribution!)

winni2k:
Have a look now. 

- moved the PSF licensed code into a separate file, which required creating a package for functools. 
- added instructions for running tests using Docker.

projectgus (commented):
The new feature in functools look good to me, thanks @winni2k! Thanks also for your patience while someone got back to you.

Regarding the Dockerfile and README: I appreciate you providing some steps for a docker container to run tests, but this probably isn't something we want to implement differently for each module. There are already several other micropython-lib modules that include tests based on `unittest`. Most of these can be run on a host as `micropython FILENAME`, which looks like it's also the case for these tests after the dependencies are installed.

If you did want to submit another PR that adds standard instructions for "Running micropython-lib tests", and includes docker as one option for this, then that could potentially be very useful. However, suggest doing that separately to this total_ordering PR.

python-stdlib/functools/functools/total_ordering.py
```diff
@@ -0,0 +1,133 @@
...
+################################################################################
+### total_ordering class decorator
+################################################################################
+
+# The total ordering functions all invoke the root magic method directly
+# rather than using the corresponding operator.  This avoids possible
+# infinite recursion that could occur when the operator dispatch logic
+# detects a NotImplemented result and then calls a reflected method.
+
+def _gt_from_lt(self, other, NotImplemented=NotImplemented):
```
dpgeorge, line 24: I'm not sure why the `NotImplemented=NotImplemented` argument is there, but I doubt we need it in MicroPython.  So I suggest removing it (and all below) to keep the code small.

python-stdlib/functools/functools/total_ordering.py
```diff
@@ -0,0 +1,133 @@
...
+    'Return a > b.  Computed by @total_ordering from (not a < b) and (a != b).'
+    op_result = self.__lt__(other)
+    if op_result is NotImplemented:
+        return op_result
+    return not op_result and self != other
+
+def _le_from_lt(self, other, NotImplemented=NotImplemented):
+    'Return a <= b.  Computed by @total_ordering from (a < b) or (a == b).'
+    op_result = self.__lt__(other)
+    return op_result or self == other
```
dpgeorge, line 34: This is slightly more efficient written as `return self.__lt__(other) or self == other`.

python-stdlib/functools/functools/total_ordering.py
```diff
@@ -0,0 +1,133 @@
...
+    'Return a < b.  Computed by @total_ordering from (not a > b) and (a != b).'
+    op_result = self.__gt__(other)
+    if op_result is NotImplemented:
+        return op_result
+    return not op_result and self != other
+
+def _ge_from_gt(self, other, NotImplemented=NotImplemented):
+    'Return a >= b.  Computed by @total_ordering from (a > b) or (a == b).'
+    op_result = self.__gt__(other)
+    return op_result or self == other
```
dpgeorge, line 74: This is slightly more efficient written as `return self.__gt__(other) or self == other`.

python-stdlib/functools/functools/total_ordering.py
```diff
@@ -0,0 +1,133 @@
...
+    """Class decorator that fills in missing ordering methods"""
+    # Find user-defined comparisons (not those inherited from object).
+    roots = {op for op in _convert if getattr(cls, op, None) is not getattr(object, op, None)}
+    if not roots:
+        raise ValueError('must define at least one ordering operation: < > <= >=')
+    root = max(roots)       # prefer __lt__ to __le__ to __gt__ to __ge__
+    for opname, opfunc in _convert[root]:
+        if opname not in roots:
+# function objects have no attributes in micropython
+#            opfunc.__name__ = opname
```
dpgeorge, line 129: Please remove this commented-out code.

python-stdlib/functools/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM micropython/unix:v1.22
+
+RUN micropython -m mip install unittest
+RUN micropython -m mip install unittest-discover
```
dpgeorge, line 4: As @projectgus mentioned, this is not the right place for docker files, so please remove all docker-related things from this PR.

PR #821, opened 2024-03-06
Senml/cbor2 update.

iabdalkader:
This update deprecates the decoder and encoder module, to maintain compatibility with the CPython cbor2 on pypi. The cbor2 module should now be imported and used directly instead of the decoder/encoder modules, for example:

```python
import cbor2
cbor2.dumps(...)
```

The `senml` library, which uses `cbor2`, is updated to use the new API.


python-ecosys/cbor2/manifest.py
```diff
@@ -1,3 +1,3 @@
-metadata(version="0.1.0", pypi="cbor2")
+metadata(version="0.1.1", pypi="cbor2")
```
dpgeorge, line 1: This will need more than a patch release bump.

I suggest to bump the major release to 1.0.0 and not even include the backwards compatible `decoder.py` and `encoder.py` files.

Otherwise, if you think backwards compatibility (with a message printed...) is needed, then I suggest version 0.2.0.

iabdalkader: No it's not needed for our use case.

dpgeorge:
Thanks, this is a good improvement, to make `cbor2` match CPython.

PR #822, opened 2024-03-07
Implement the utop module

DvdGiessen:
Implements a new module, `utop`, as suggested in https://github.com/micropython/micropython/pull/12732#issuecomment-1790040708.

Currently only works on the `esp32` port, and requires https://github.com/micropython/micropython/pull/12732.

An animated GIF to demonstrate what the output looks like:

utop

Opted to give it a generic name instead of ESP32-specific, as there was mention of other platforms also having options to retrieve runtime performance data, which might be nice to support at a later time.

Compared to the version posted in that `esp32` MR ([here](https://github.com/micropython/micropython/blob/e2316bd17741dd74e49faa1efdea89712695657b/docs/library/esp32.rst#functions)) this version is a few lines longer to make it a bit easier to read and expand later. For example it includes an optional argument for renaming threads, useful to distinguish multiple `mp_thread` instances.

First commit implements the functionality as demo'd in the other MR. The second commit adds two lines of information about the current state of the ESP-IDF heap, as an example of how we could expand this later to be a more useful development utility.

### DIFF
--- a/micropython/utop/utop.py
+++ b/micropython/utop/utop.py
@@ -1,3 +1,4 @@
+import micropython
 import time
 
 try:
@@ -71,6 +72,12 @@ def top(update_interval_ms=1000, timeout_ms=None, thread_names={}):
         else:
             print("INFO: Platform does not support listing active tasks.\x1b[K")
             line_count += 1
+
+        print("\x1b[K")
+        line_count += 1
+        print("MicroPython ", end="")
+        micropython.mem_info()
+        line_count += 3
 
         if previous_line_count > line_count:
             for _ in range(previous_line_count - line_count):
--- a/micropython/utop/utop.py
+++ b/micropython/utop/utop.py
@@ -79,6 +79,23 @@ def top(update_interval_ms=1000, timeout_ms=None, thread_names={}):
         micropython.mem_info()
         line_count += 3
 
+        if esp32 is not None:
+            print("\x1b[K")
+            line_count += 1
+            for name, cap in (("data", esp32.HEAP_DATA), ("exec", esp32.HEAP_EXEC)):
+                heaps = esp32.idf_heap_info(cap)
+                print(
+                    "IDF heap ({}): {} regions, {} total, {} free, {} largest contiguous, {} min free watermark\x1b[K".format(
+                        name,
+                        len(heaps),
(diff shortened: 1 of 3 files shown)


micropython/utop/utop.py
```diff
@@ -0,0 +1,108 @@
...
+                    )
+                )
+                line_count += 1
+
+        if previous_line_count > line_count:
+            for _ in range(previous_line_count - line_count):
+                print("\x1b[K")
+            print("\x1b[{}A".format(previous_line_count - line_count), end="")
+
+        previous_total_runtime = total_runtime
```
dpgeorge, line 102: On non-esp32 boards, or esp32 without `idf_task_stats()`, `total_runtime` is undefined.

micropython/utop/utop.py
```diff
@@ -0,0 +1,108 @@
...
+                    task_runtime,
+                    task_stackhighwatermark,
+                    task_coreid,
+                ) in tasks:
+                    task_runtime_percentage = "-"
+                    if previous_total_runtime != 0 and task_id in previous_task_runtimes:
+                        task_runtime_percentage = "{:.2f}%".format(
+                            100
+                            * time.ticks_diff(task_runtime, previous_task_runtimes[task_id])
+                            / time.ticks_diff(total_runtime, previous_total_runtime)
```
dpgeorge, line 50: Is `ticks_diff()` the correct thing to use here?  AFAIK, these times are from FreeRTOS and are not in the correct wrap-around-range to use with `ticks_diff`.

dpgeorge (commented):
Sorry this took ages to get to.  It looks good!  I like how you've made it generic, for any target.  That's a good idea, so it can be expanded in the future.

micropython/utop/utop.py
DvdGiessen: You're right, indeed that should just be a plain 32bit unsigned wraparound.

micropython/utop/utop.py
```diff
@@ -0,0 +1,108 @@
+import micropython
+import time
+import _thread
```
dpgeorge, line 3: This import needs to be optional, only needed on esp32.

dpgeorge (approved):

PR #826, opened 2024-03-18
pyjwt: ES256 algorithm support for PyJWT.

jonahbron:
Add optional support for ES256 JWT signing/verifying to PyJWT using @dmazzella's cryptography port.

Test output:

```
Testing HS256
     Encode/decode test: OK
     Invalid signature test: OK
     Expired token test: OK
Testing ES256
     Encode/decode ES256 test: OK
     Invalid signature test: OK
```

### DIFF
--- a/python-ecosys/pyjwt/jwt.py
+++ b/python-ecosys/pyjwt/jwt.py
@@ -4,6 +4,17 @@ import hmac
 import json
 from time import time
 
+# Optionally depend on https://github.com/dmazzella/ucryptography
+try:
+    # Try importing from ucryptography port.
+    import cryptography
+    from cryptography import hashes, ec, serialization, utils
+
+    _ec_supported = True
+except ImportError:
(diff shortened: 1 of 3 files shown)


dpgeorge:
Thanks for the contribution!

This isn't a full review, but some initial questions/feedback:
- Is this ES256 support compatible with the CPython PyJWT library?
- MicroPython tries to be minimal where possible, so I suggest inlining `_sig_jws_to_der` and `_sig_der_to_jws` at their place of use (they are only used once).  That will save a bit of code size.
- Also maybe the error message about required dependencies can be a little shorter/terser, to save size.
- Can you please put the new test into a separate test for ES256, so that the tests can be run independently, and the original one still pass without `cryptography` installed.
- Would be great to also add the tests to CI, see `tools/ci.sh`.

dpgeorge:
@jonahbron any chance you'll be able to respond to the above feedback/questions?

PR #835, opened 2024-03-25
bluetooth/aioble/examples/temp_sensor.py: Properly notifies on update.

mores:


konistehrad:
I think this should be merged. The `timeout_ms=None` call in particular keeps something deep in the internals from blocking service queries from the central, and without this the central doesn't get timely updates.

aetelani:
Thanks for example update. I was doing this myself because the current example in master was not working.

dpgeorge:
Thanks for the contribution.  I can confirm that both fixes are needed:
- for the char write, so that subscribers are notified
- for the disconnected, so that the temp sensor remains connected until the central initiates a disconnect

PR #846, opened 2024-04-14
aiohttp: Fix type of `Sec-WebSocket-Key` in `aiohttp_ws.py`

AuroraTea:
The function `binascii.b2a_base64()` returns a `bytes`.
Here needs a `string`
Otherwise, the value of `Sec-WebSocket-Key` in the headers will be `b'<BASE64-ENCODED_RANDOM_VALUE>'`.
It means there will be `b'` at the beginning and `'` at the end.

### DIFF
--- a/python-ecosys/aiohttp/aiohttp/aiohttp_ws.py
+++ b/python-ecosys/aiohttp/aiohttp/aiohttp_ws.py
@@ -86,7 +86,7 @@ class WebSocketClient:
 
     def _process_websocket_frame(self, opcode, payload):
         if opcode == self.TEXT:
-            payload = payload.decode()
+            payload = str(payload, "utf-8")
         elif opcode == self.BINARY:
             pass
         elif opcode == self.CLOSE:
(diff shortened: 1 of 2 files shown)


python-ecosys/aiohttp/aiohttp/aiohttp_ws.py
```diff
@@ -143,7 +143,7 @@ async def handshake(self, uri, ssl, req):
         headers["Host"] = f"{uri.hostname}:{uri.port}"
         headers["Connection"] = "Upgrade"
         headers["Upgrade"] = "websocket"
-        headers["Sec-WebSocket-Key"] = key
+        headers["Sec-WebSocket-Key"] = key.decode("utf-8")
```
dpgeorge, line 146: Please use `str(key, "utf-8")`, because the `decode()` method is not always available.

dpgeorge:
Thanks for updating.  This is now merged.

PR #849, opened 2024-04-17
aiohttp_ws: convert Sec-WebSocket-Key from bytes to str

iot49:
Without the fix, the client submits a header formatted as bytes:

```
Sec-WebSocket-Key     b'aaaaaaaaaaaaa'
```

resulting in the server rejecting the connection (I use the python websocket module in the server for checking). This PR decodes the key, and thus correctly formats the header:

```
Sec-WebSocket-Key     aaaaaaaaaaaaa
```


Carglglz:
I think this the same as #846

iot49:
yes. Let's hope it will be merged soon ...

dpgeorge:
Thanks for the patch, but I'll close this in favour of #846.

Issue #332, opened 2019-02-27
PINGRESP ignored in umqtt.simple.py


nfj25:
Hello,

In wait_msg, the PINGRESP message is ignored:

```python
if res == b"\xd0":  # PINGRESP
            sz = self.sock.read(1)[0]
            assert sz == 0
            return None
```
Shouldn't this return some information so that the calling code can determine if the server is indeed responding to pings?

michaeka79:
I'm going to have to upvote this... I feel like this should be the case as well. umqtt.robust doesn't fix this either. We need to know if the connection is still active or not. 

I found some code on https://github.com/AntonisKekempanos/SonoffMicropythonMQTT.git

Where they return b'PINGRESP' this seems like an appropriate return code for the ping.

andrewleech:
I don't know much about mqtt however from a quick search it appears that clients (like this module) are supposed to send a PINGREQ to the broker regularly, to which the broker responds with PINGRESP.

Changing wait_msg to return something different when this comes in would affect existing users of the library who haven't had to expect this message and know how to deal with it.

Perhaps instead of returning something, The library could have a new attribute in the mqtt client class where a timestamp is updated every time one of these comes in, that way an application that cares about it can query that attribute to see how long it's been since the last server communication?

michaeka79:
+1 I love that idea! That would help with about half the code I'd be trying to write to check the time intervals between pings. So yes, love it!

michaeka79:
@andrewleech were you thinking of https://github.com/fizista/micropython-umqtt.simple2.git when you suggested having it be an attribute of the client? Because it seems like that's exactly what's been implemented.

andrewleech:
I've never seen that library myself, to be honest I haven't used mqtt at all.

vitaliishchudlo:
@michaeka79 Maybe you have any ideas to fix it?
@nfj25 did you solve your problem?

nfj25:
Not using it anymore...

trescenzi:
Spent some time trying to figure out why pings weren't working before finding this issue. Is there a rationale behind this? The docs even seem to [imply that this should be handling the ping](https://mpython.readthedocs.io/en/v2.2.1/library/mPython/umqtt.simple.html#umqtt.simple.MQTTClient.ping)

> Ping server (response is automatically handled by wait_msg（））

That implies to me not that the ping is ignored, but that `wait_msg` will tell you a ping has been responded to.

[...]

PR #850, opened 2024-04-23
umqtt.simple: add ping_response_received for ping handling

trescenzi:
Currently when using `ping` the response is swallowed which makes it not particularly useful. This PR adds a flag `ping_response_received` to the client which is set to `True` if a response has been received in `wait_msg`. I've been using this approach in a personal project without any issues and it's not particularly invasive so I wanted to propose it as a solution to #332.

Another possible approach is to actually respond in `wait_msg` with an actual ping response however I'm not certain how that would interact with subscriptions and those currently using `wait_msg` so this approach was taken in order to not be breaking while still being useful.

### DIFF
--- a/micropython/umqtt.simple/README.rst
+++ b/micropython/umqtt.simple/README.rst
@@ -51,6 +51,7 @@ follows MQTT control operations, and maps them to class methods:
   clean_session=True argument is used (default)).
 * ``disconnect()`` - Disconnect from a server, release resources.
 * ``ping()`` - Ping server (response is processed automatically by wait_msg()).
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks for the contribution.

I think it would be more useful to make a callback to the user upon ping reception.  There's already the `.cb` attribute (set via `set_callback`) which could be reused for this purpose.  Maybe passing an empty bytes for topic and msg (passing `None` could break existing code that expects bytes for these arguments).  Eg:
```py
        if res == b"\xd0":  # PINGRESP
            sz = self.sock.read(1)[0]
            assert sz == 0
            if self.cb:
                self.cb(b"", b"")
            return None
```

PR #856, opened 2024-05-14
all: Enable ruff F841 'Local variable is assigned to but never used'.

projectgus:
Most of these look like they were used for print debugging and then kept in when the print statements were removed or commented.

Some look like missing or incomplete functionality, these have been marked with comments where possible.

This PR shouldn't change runtime behaviour at all, however I can't rule out that one of these warnings wasn't masking an existing bug.


dpgeorge:
Thank you!  Looks good to me.

PR #857, opened 2024-05-14
all: Enable Ruff lint F541 'f-string without any placeholders'.

projectgus:
This is a pretty straightforward lint and fix, should save a small amount of bytecode.


dpgeorge:
Thanks!

PR #858, opened 2024-05-14
all: Enable ruff E401,E402 import lints.

projectgus:
Mostly small cleanups, and explicit disable for examples/tests which insert the current directory into the path before importing.

### DIFF
--- a/micropython/bluetooth/aioble/examples/l2cap_file_client.py
+++ b/micropython/bluetooth/aioble/examples/l2cap_file_client.py
@@ -5,6 +5,7 @@
 
 import sys
 
+# ruff: noqa: E402
 sys.path.append("")
 
 from micropython import const
--- a/micropython/bluetooth/aioble/examples/l2cap_file_server.py
+++ b/micropython/bluetooth/aioble/examples/l2cap_file_server.py
@@ -16,6 +16,7 @@
 
 import sys
 
+# ruff: noqa: E402
 sys.path.append("")
 
 from micropython import const
--- a/micropython/bluetooth/aioble/examples/temp_client.py
+++ b/micropython/bluetooth/aioble/examples/temp_client.py
@@ -1,5 +1,6 @@
 import sys
 
+# ruff: noqa: E402
 sys.path.append("")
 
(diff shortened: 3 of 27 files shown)


dpgeorge:
This is a bit of a change, making each import go on a separate line.  But I guess we have to do that because we have committed to using `ruff check` for our style.

Luckily there is no difference in bytecode between `import a, b` and `import a; import b`.

projectgus:
> This is a bit of a change, making each import go on a separate line. But I guess we have to do that because we have committed to using ruff check for our style.

I think we could reasonably commit to ignoring these, especially E402 (imports not at top ) because we need to add `noqa` for legitimate use case in quite a few places.

However, the current usage is pretty arbitrary - sometimes imports on the same line, sometimes imports on different lines, sometimes the same source file has both approaches for no apparent reason. So it is also nice to be consistent, IMO.

stinos:
All imports on one line (or even going further: `isort` style sorted and grouped) just seems cleaner overall: more readable, nicer diffs, consistency.

projectgus:
> All imports on one line (or even going further: `isort` style sorted and grouped) just seems cleaner overall: more readable, nicer diffs, consistency.

I don't exactly disagree, but I do want to point out that theseindividual small differences in style preference is why many projects just pick a formatter (like ruff) and have it auto-format the code. Instant consistency, minimal time overhead. No one gets their preferred individual style, but the code base all has the same style.

dpgeorge:
> or even going further: `isort` style sorted and grouped

AFAIK, `isort` keeps separate modules on separate import lines??

> nicer diffs

Aren't diffs smaller if imports are on separate lines?

stinos:
> AFAIK, isort keeps separate modules on separate import lines??

Depends, it has a bit too much configuration optins actually. Though I think ruff's isort implementation is betrer in that regard.

> Aren't diffs smaller if imports are on separate lines?

Oops. I actually did mean to say 'all imports on their own line' ! Will edit original comment.

dpgeorge:
OK, we're all on the same page now :)

Merged.

Issue #844, opened 2024-04-11
urequests still using HTTP/1.0


linusoft:
The urequests module is still using HTTP/1.0 for HTTP requests:

line 39 in micropython-lib/micropython/urllib.urequest/urllib/urequest.py:

 ` s.write(b" HTTP/1.0\r\nHost: ")`

Since a host header is included, can this be updated to use HTTP/1.1 or even HTTP/2.0?

Some sites (e.g. mongodb) are now responding with '426 Update Required' for HTTP/1.0 requests.

Thanks

tharuka-pavith:
Hey! I also think it is better to update the header in  [line 39](https://github.com/micropython/micropython-lib/blob/583bc0da70049f3b200d03e919321ac8dbeb2eb8/micropython/urllib.urequest/urllib/urequest.py#L39C9-L39C40) of `micropython-lib/micropython/urllib.urequest/urllib/urequest.py` as,
```diff
- s.write(b" HTTP/1.0\r\nHost: ")
+ s.write(b" HTTP/2.0\r\nHost: ") 
```
As I know, most major web servers and CDNs provide backward compatibility. Therefore there will be no harm by updating it.

sosi-deadeye:
[HTTP/2](https://http2.github.io/) is a binary protocol and uses [HPACK](https://httpwg.org/specs/rfc7541.html) for header compression. Changing this, require additional code to handle the protocol and header compression.

I also think the change to HTTP/1.1 may require some additional code, to handle the different connection types : `close`, `keep-alive`.

In addition, there are different [Transfer-Endocings](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding).

mattytrentini:
It would be good to merge _back_ some of the changes from the CircuitPython requests library (they've updated it since originally starting with the MicroPython implementation). In it, they respond with v1.1, though I'm not sure it 100% conforms to spec...

https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/main/adafruit_requests.py

PR #861, opened 2024-05-24
Change HTTP/1.0 to HTTP/1.1 in `requests` Python module

sgbaird:
On the Pico W, I'm suddenly having trouble accessing an API that used to be compatible. The issue seems to be the HTTP version, which I documented at https://github.com/orgs/micropython/discussions/15112. As somewhat of a last try, I copied over https://github.com/micropython/micropython-lib/blob/e025c843b60e93689f0f991d753010bb5bd6a722/python-ecosys/requests/requests/__init__.py into a `urequests_2.py`, used `import urequests_2 as urequests`, and simply replaced `1.0` with `1.1`:
```python
s.write(b"%s /%s HTTP/1.0\r\n" % (method, path))
```
with:
```python
s.write(b"%s /%s HTTP/1.1\r\n" % (method, path))
```
I was able to successfully send data through the API after that. I also wanted to note that I couldn't use the version on `master` due to `ImportError: no module named 'tls'`, hence the pinned version from above. I have tutorials, videos, and a module within a course dependent on this, so it would be great to have this incorporated into a stable version of Pico W MicroPython if it's feasible.

### DIFF
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -93,7 +93,7 @@ def request(
             context = tls.SSLContext(tls.PROTOCOL_TLS_CLIENT)
             context.verify_mode = tls.CERT_NONE
             s = context.wrap_socket(s, server_hostname=host)
-        s.write(b"%s /%s HTTP/1.0\r\n" % (method, path))
+        s.write(b"%s /%s HTTP/1.1\r\n" % (method, path))
         if "Host" not in headers:
             s.write(b"Host: %s\r\n" % host)
         # Iterate over keys to avoid tuple alloc


mattytrentini:
I think it's a good idea to merge this as-is. 

However, it would also be useful to take a look at CircuitPython's request library. It was based on this one but has since had some updates - it'd be good to evaluate if other changes should be integrated too.

dpgeorge:
See related #844.

It's a shame that HTTP/1.0 is becoming obsolete because it's a nice and simple (and still functional) protocol.

But, yeah, we need to update to use HTTP/1.1.  The problem is it's not as simple as changing the `1.0` to a `1.1`.  The code also needs to be improved to support chunked responses, at the very least.  Also probably need to add `Accept-Encoding: identity` to the headers to prevent the server from responding with compressed data, which we can't deal with.

Note that our `aiohttp` library does support HTTP/1.1, but it's not really a drop-in replacement for `requests`.

In summary: let's update to HTTP/1.1 but it needs to be done to the spec, and tested well.

RGSMA:
I was banging my head against the wall regarding a similar problem I was having with an API. I appreciate you figuring this out; your solution works perfectly for my needs. Thanks again!

dpgeorge:
Closing in favour of #1124.

PR #881, opened 2024-06-12
requests: Make possible to override headers and allow raw data upload.

kapetan:
This removes all the hard-coded request headers from the requests module so they can be overridden by user provided headers dict.
Furthermore allow streaming request data without chunk encoding in those cases where content length is known but it's not desirable to load the whole content into memory. Also some servers (e.g. nginx) reject HTTP/1.0 requests with the Transfer-Encoding header set.
The change should be backwards compatible as long as the user hasn't provided any of the previously hard-coded headers.


dpgeorge:
Thank you!

PR #886, opened 2024-06-18
usb-device-cdc: Fix lost data in read() path if short reads happened.

### LINKED ISSUE #885: Serial connections dropping characters, not flushing buffers
I tested this on a Pico running MP 1.23 and the latest version of usb-device-cdc on mip
main.py:
```python
from io import BytesIO
from select import poll, POLLIN, POLLHUP, POLLERR
from time import sleep_ms

import usb.device

SERIAL_TWO = None


def ingest():
	f = BytesIO(512)
	ba = bytearray(256)
	mv = memoryview(ba)

	print("beginning ingest")
	# SERIAL_TWO.read(4)
	while True:
		bytes_read = SERIAL_TWO.readinto(mv)
		if bytes_read:
			f.write(mv[:bytes_read])
			print("{} written, total {}".format(bytes_read, f.tell()))
		else:
			break

	f.seek(0)
	print(f.read(-1))


def main():
	global SERIAL_TWO

	buf = bytearray(4)
	SERIAL_TWO = CDCInterface(timeout=0)
	usb.device.get().init(SERIAL_TWO, builtin_driver=True)
	while not SERIAL_TWO.is_open():
		sleep_ms(100)

	api_poll = poll()
	api_poll.register(SERIAL_TWO, POLLIN)

	while True:
		cmd_waiting = api_poll.poll(0)
		if cmd_waiting and cmd_waiting[0][1] is POLLIN:
			# SERIAL_TWO.read(4)
			ingest()

		sleep_ms(1000)

if __name__ == '__main__': main()
```

client.py:
```python
from serial import Serial
from string import ascii_letters

def main():
	test_str = ascii_letters * 7
	conn = Serial("/dev/ttyACM1", 115200)
	conn.read_all()
	data = "\x03" + "\n" + "\x21" + "\n" + test_str
	conn.write(data.encode())

if __name__ == '__main__': main()
```

I need to read four bytes off the input to confirm I should call `ingest()`, but in doing so I consistently lose eight bytes after the first call to `readinto()` (the output skips V -> e after 256 bytes) and some of the content never appears to leave the buffer. Moving the `read()` call into `ingest()` doesn't help, nor does using `readinto()` instead.

projectgus:
If the CDC receive buffer was full and some code read less than 64 bytes (wMaxTransferSize), the CDC code would submit an OUT transfer with N<64 bytes length to fill the buffer back up.

However if the host had more than N bytes to send then it would still send the full 64 bytes (correctly) in the transfer. The remaining (64-N) bytes would be lost.

Fix is to not send a new transfer until 64 bytes is free. Adds the restriction that CDCInterface rxbuf has to be at least 64 bytes.

Closes #885.

### DIFF
--- a/micropython/usb/usb-device-cdc/manifest.py
+++ b/micropython/usb/usb-device-cdc/manifest.py
@@ -1,3 +1,3 @@
-metadata(version="0.1.0")
+metadata(version="0.1.1")
 require("usb-device")
 package("usb")
--- a/micropython/usb/usb-device-cdc/usb/device/cdc.py
+++ b/micropython/usb/usb-device-cdc/usb/device/cdc.py
@@ -144,8 +144,8 @@ class CDCInterface(io.IOBase, Interface):
         if flow != 0:
             raise NotImplementedError  # UART flow control currently not supported
 
-        if not (txbuf and rxbuf):
-            raise ValueError  # Buffer sizes are required
+        if not (txbuf and rxbuf >= _BULK_EP_LEN):
+            raise ValueError  # Buffer sizes are required, rxbuf must be at least one EP
 
         self._timeout = timeout
         self._wb = Buffer(txbuf)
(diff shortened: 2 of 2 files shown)


dpgeorge:
Thanks!

PR #887, opened 2024-06-24
StreamHandler should call parent constructor

greezybacon:
Otherwise there's a crash on line 70 where `level` is not a property of the class unless explicitly set with ::setLevel


dpgeorge:
Thanks!

PR #889, opened 2024-07-03
usb-device-keyboard: Fix ; and ` keycode names.

dpgeorge:
They should be named as the un-shifted version.


dpgeorge:
@projectgus FYI

projectgus:
LGTM! I double checked the USB "HID Usages and Descriptions" document (section 10
 "Keyboard/Keypad Page (0x07)") and it describes them as "Keyboard ; and :" and "Keyboard Grave Accent and Tilde", so this is the right fix.

PR #890, opened 2024-07-03
settimeout for mqtt socket

prabhu-yu:
Problem statement:
If there are any network issues, mqtt will block on socket non-deterministically.
In such cases, only way to come out of the blocking is to reboot using watch dog timers.
This is costly solution.

Solution:
Alternatively, developer can set the max timeout for the socket.
Upon any issue, mqtt lib will throw exception. Developer can catch it, take
right actions like, restarting the task without rebooting the whole device.

This brings determinism and gives the control to developer to choose right time for her/his use case. This fix works for async applications too.

(I plan to make whole umqtt async compatible.)

### DIFF
--- a/micropython/umqtt.simple/umqtt/simple.py
+++ b/micropython/umqtt.simple/umqtt/simple.py
@@ -60,8 +60,9 @@ class MQTTClient:
         self.lw_qos = qos
         self.lw_retain = retain
 
-    def connect(self, clean_session=True):
+    def connect(self, clean_session=True, timeout=None):
         self.sock = socket.socket()
+        self.sock.settimeout(timeout)
         addr = socket.getaddrinfo(self.server, self.port)[0][-1]
         self.sock.connect(addr)
         if self.ssl:
--- a/micropython/umqtt.simple/manifest.py
+++ b/micropython/umqtt.simple/manifest.py
@@ -1,4 +1,4 @@
-metadata(description="Lightweight MQTT client for MicroPython.", version="1.4.0")
+metadata(description="Lightweight MQTT client for MicroPython.", version="1.5.0")
 
 # Originally written by Paul Sokolovsky.
 


micropython/umqtt.simple/umqtt/simple.py
```diff
@@ -60,8 +60,16 @@ def set_last_will(self, topic, msg, retain=False, qos=0):
...
+        '''Blocking socket can cause non determinstic hang incase of n/w failure.
+        Hence, by setting timeout, application programmer can get exception in case of failure.
+        This enable him/her to take appropriate action like restarting the task etc.
+        This works for async apps too.
+        '''
+        self._sock.settimeout(value)
+
     def connect(self, clean_session=True):
-        self.sock = socket.socket()
+        self._sock = self.sock = socket.socket()
```
dpgeorge, line 72: I think it would be better to instead add a `timeout=None` argument to this `connect()` method.  Then call `self.sock.settimeout(timeout)` after creating the socket.

That would mean that the connect call below, and wrap_socket, would also apply the timeout.

dpgeorge:
> I plan to make whole umqtt async compatible.

See https://github.com/tve/mqboard/tree/master/mqtt_async

micropython/umqtt.simple/umqtt/simple.py
prabhu-yu: Hi @dpgeorge ,
Agree with you. That will be better.

prabhu-yu:
> > I plan to make whole umqtt async compatible.
> 
> See https://github.com/tve/mqboard/tree/master/mqtt_async

Thank you for the link.
Is it going to be part of the micropython? [https://github.com/micropython/micropython-lib/tree/master/micropython]
If so, it will have higher visibility.

dpgeorge:
> Thank you for the link.
> Is it going to be part of the micropython?

Actually, this is a better version: https://github.com/peterhinch/micropython-mqtt

Eventually we would like to add links from this repository to external/third-party libraries like the above, to make them easier to install, and have better visibility.

micropython/umqtt.simple/umqtt/simple.py
dpgeorge: @prabhu-yu Will you update this PR with the above suggestion?  If not I can do it.

prabhu-yu: Hi @dpgeorge,
I will do. Please give me a day. 
I will test and get back to you with changes.

prabhu-yu:
> > I plan to make whole umqtt async compatible.
> 
> See https://github.com/tve/mqboard/tree/master/mqtt_async

This will be a lot helpful to all those use async. Async really shines as it gives better control to developer over threads/processes in terms of scheduling. This in turn solves many locking issues that are common in threads. in anycase, async mqtt blends nicely into all async applications.

micropython/umqtt.simple/umqtt/simple.py
```diff
@@ -60,8 +60,10 @@ def set_last_will(self, topic, msg, retain=False, qos=0):
         self.lw_qos = qos
         self.lw_retain = retain
 
-    def connect(self, clean_session=True):
+    def connect(self, clean_session=True, timeout=None):
         self.sock = socket.socket()
+        if timeout is not None:
+            self.sock.settimeout(timeout)
```
dpgeorge, line 66: If you pass `None` to `settimeout` then it sets it to blocking mode, which is also the default after creating a fresh socket.

So, the `if timeout is not None` line is not needed.  It can just set the timeout unconditionally and still have the same behaviour as before.  Doing it that way makes the code smaller and simpler.

prabhu-yu: Hi @dpgeorge ,
Agree with you. I submitted the change.
(I was trying to be little cautious by avoiding to execute settimeout when default value arrives! I agree, this is not a clean solution!)

pullagad: @dpgeorge 
May you please suggest the next step? 
Should I create a PR for this?

dpgeorge: This change looks good now, thank you.

But I see on your commits that the author is `Your Name <you@example.com>`.  Is that intentional?  Usually we require a proper name and email, for attribution of the work.

If you want to remain anonymous, I can change the author to my name.

prabhu-yu: Hi @dpgeorge ,
Thank you for notifying me 🙏.
I have updated my Name/E-mail id and pushed it again.
Please let me know if I need to do something further.

dpgeorge: This looks good now, thank you.

dpgeorge (approved):

Issue #873, opened 2024-06-07
HID keyboard won't release keys when called via schedule


mischif:
Calling `KeyboardInterface().send_keys()` from a function called via `micropython.schedule()` doesn't seem to allow keys to be released, causing a spamming issue. This code reproduces the issue:

```python
import usb.device
from usb.device.keyboard import KeyboardInterface
from time import sleep_ms
from micropython import schedule

KBD = None

def send_keys(foo):
    keys = [-0x02, 5]
    print("sending keypress")
    KBD.send_keys(keys)

    keys.clear()
    print("clearing keypress")
    KBD.send_keys(keys)


def main():
    global KBD
    KBD = KeyboardInterface()
    usb.device.get().init(KBD, builtin_driver=True)

    sleep_ms(5000)
    print("starting loop")

    while True:
        send_keys(None)
        # schedule(send_keys, None)
        sleep_ms(3000)

if __name__ == '__main__': main()
```

running `schedule(send_keys, None)` instead of `send_keys(None)` causes a flood of :b:, this is on a Pico running 1.23.0 and whatever the latest version of usb-device-keyboard is currently in mip.

projectgus:
Hi @mischif,

Thanks for the very clear report.

At the moment calling send_keys() will block for up to the timeout (default 100ms) waiting for the USB endpoint to be available to queue a new transfer, and then returns True for success or False for failure. In your scheduled send_keys() function I'm guessing the second call is returning False indicating that the transfer wasn't able to queue to be sent to the host. Probably this is not ideal behaviour for this function, it should throw an exception on timeout. I'll look at changing it.

However, the underlying problem will still be there. An operation like send_keys() returns when the transfer is queued for sending to the host, but this may still take some time before the USB host actually performs the transfer. In the meantime, the USB endpoint isn't available. The reason why the USB endpoint never becomes available, even after 100ms, is that the transfer completion callback is also processed via a `schedule()` internally. Blocking inside a schedule callback prevents the USB completion callback from running.

If you need to send and release a keypress with `micropython.schedule()` then you can probably do something like this to chain multiple schedule callbacks together:

```py
def send_and_release(keys):
    if not KBD.is_open():
        print("Can't send, USB disconnected")
    elif not KBD.send_keys(keys, 0):
        schedule(send_and_release, keys)  # Try again after other callbacks run
    elif keys:
        schedule(send_and_release, [])  # Schedule the release if needed
```

[...]

PR #892, opened 2024-07-04
usb/usb-device-keyboard: Add key release.

amogha1234:
Micropython 1.23 preview version is used with Raspberry pi pico w board.
Trying to send keys using usb-keyboard library. It was clicking the key, but not releasing it, similar to #873 . Resulting in continuous sending key.
Hence added release_all() function, which releases all keys. And it worked with Rpi pico w.

### DIFF
--- a/micropython/usb/usb-device-keyboard/manifest.py
+++ b/micropython/usb/usb-device-keyboard/manifest.py
@@ -1,3 +1,3 @@
-metadata(version="0.1.0")
+metadata(version="0.1.1")
 require("usb-device-hid")
 package("usb")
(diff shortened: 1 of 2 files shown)


dpgeorge:
You should be able to release all keys by calling `keyboard.send_keys(())`, ie pass an empty tuple (or empty list `[]` but an empty tuple is more efficient because it doesn't allocate memory).

amogha1234:
> You should be able to release all keys by calling `keyboard.send_keys(())`, ie pass an empty tuple (or empty list `[]` but an empty tuple is more efficient because it doesn't allocate memory).

Thank you,
I thought it is better to have a dedicated function to release keys as it is present in other keyboard libraries of circuit-python and Arduino, it will be useful for the users.

dpgeorge:
Instead of adding extra code, maybe it's acceptable to add some documentation on how to release keys.

What do you think @projectgus ?

dpgeorge:
I suggest closing this PR in favour of #1132.

PR #893, opened 2024-07-04
micropython/lora: Update the implicit header option in the _SX127x.

jandriea:
The `_reg_update` method must be called after updating the implicit header option's bit.


dpgeorge (approved):

dpgeorge:
Thanks for the fix, this looks correct.

PR #894, opened 2024-07-06
usb-device: Allow signaling capability of remote_wakeup.

felixdoerre:
To use this feature you need to create a usb device signaling remote wakeup and then enable remote wakeup on the host (on linux write `enabled` to `/sys/bus/usb/devices/<device>/power/wakeup`, or let `udev` do that). Then you can wake up the host when it is on standby using USBDevice.remote_wakeup.

See also [micropython#15262](https://github.com/micropython/micropython/pull/15262).

### DIFF
--- a/micropython/usb/usb-device/manifest.py
+++ b/micropython/usb/usb-device/manifest.py
@@ -1,2 +1,2 @@
-metadata(version="0.1.1")
+metadata(version="0.2.0")
 package("usb")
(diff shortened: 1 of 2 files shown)


dpgeorge (approved):

dpgeorge:
Thank you, this looks good.

dpgeorge:
I updated the version to 0.2.0, because (as per semver) this adds something to the public API.

PR #905, opened 2024-07-31
unix-ffi/sqlite3: Fix 64 bit support, statement finalization, and add `UFI`, `commit`, and `rollback` support.

rhermanklink:
The current `sqlite.py` file only takes 4 bytes for pointers, disallowing 64 bit platforms, and also doesn't finalize statements except in the `Cursor.close()`, causing leaks and issues with `Connections.close()`. This PR fixes these issues by determining the pointer size dynamically for the platform, and finalizes the previous statements when a new one is made in `execute()`.

Along with these fixes, this PR also adds the ability to use `URI`, and `commit` & `rollback` functionality to increase similarity with the CPython sqlite3 module. The addition of `commit` & `rollback` includes the addition of the autocommit and transaction posibilities as with the CPython sqlite3 module, with the same default mode enabled here.

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -66,6 +66,9 @@ function ci_package_tests_run {
         unix-ffi/gettext/test_gettext.py \
         unix-ffi/pwd/test_getpwnam.py \
         unix-ffi/re/test_re.py \
+        unix-ffi/sqlite3/test_sqlite3.py \
+        unix-ffi/sqlite3/test_sqlite3_2.py \
+        unix-ffi/sqlite3/test_sqlite3_3.py \
         unix-ffi/time/test_strftime.py \
         ; do
         echo "Running test $test"
--- a/unix-ffi/sqlite3/manifest.py
+++ b/unix-ffi/sqlite3/manifest.py
@@ -1,4 +1,4 @@
-metadata(version="0.2.4")
+metadata(version="0.3.0")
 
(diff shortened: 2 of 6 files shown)


mattytrentini:
This looks in a reasonable state to merge; thanks for the contribution!

I've ensured that the tests pass and, for convenience, have updated the `micropython/unix` container to include `libffi-dev` and `libsqlite3-dev` libraries to avoid the hassle of installing them if this module is to be used.

For reference, here's how one of the test files can be executed (running the command at the root of a micropython-lib clone):

```bash
> docker run -ti --rm -v $(pwd):/mplib -v $(pwd)/unix-ffi/sqlite3:/code -w /code micropython/unix
MicroPython v1.23.0 on 2024-08-20; linux [GCC 12.2.0] version
>>> import sys; sys.path.append("/mplib/unix-ffi/ffilib")
>>> import test_sqlite3
(1, 'foo', 3.0)
(3, 3, 3)
>>>
```

It could do with a README explaining how to use the library - but to be fair, the whole unix-ffi documentation should be improved!

Merging this as-is will mean that folks that want to use sqlite in the unix port have a good option.

dpgeorge:
Thanks for the contribution, this is a very nice improvement!

I have added commits to bump the manifest.py version, and include the sqlite3 tests in CI.

I also removed trailing whitespace in the last commit.

PR #907, opened 2024-08-07
Draft: micropython/streampair: Package to create bi-directional linked stream objects.

andrewleech:
Returns two bi-directional linked stream objects where writes to one can be read from the other and vice/versa.

This can be used somewhat similarly to a socket.socketpair in python, like a pipe of data that can be used to connect stream consumers (eg. asyncio.StreamWriter, mock Uart)


andrewleech:
The RingIO module has been merged in micropython so this should be updated to use it

PR #910, opened 2024-08-20
tools/build.py: Add "path" entry to index.json.

dpgeorge:
This points to the package's base directory of the within the micropython-lib directory structure.

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -64,7 +64,7 @@
 
 # index.json is:
 # {
-#   "v": 1,                      <-- file format version
+#   "v": 2,                      <-- file format version
 #   "updated": <utc-seconds-since-1970>,
 #   "packages": {
 #     {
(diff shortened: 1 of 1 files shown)


dpgeorge:
@mattytrentini hopefully this is what's needed for mim.

HowManyOliversAreThere:
@dpgeorge This looks like it should do the trick, I'll try this out to generate the json and insert it into mim to sanity check the path but should be great.

For reference, mim currently uses https://wb5lsudgvpt4f4j4pxomg5z3ui0dcgjt.lambda-url.ap-southeast-2.on.aws/ for the current proof of concept, where we've injected the `url` value into the existing json object by trawling the repo. Will be much nicer to use the official json directly!

HowManyOliversAreThere:
Added support for this parameter into mim and generated the json from this repo to test against and found it to be working as expected (from the spot checks I did), including things like finding the `aioble` README that was missed with my naive implementation. Thanks @dpgeorge !

dpgeorge:
@HowManyOliversAreThere thanks for testing, great that it works!  Now merged.

mattytrentini:
@dpgeorge when will this appear at https://micropython.org/pi/v2/index.json? I had thought it would update the build machine overnight but I'm seeing v1...

PR #927, opened 2024-10-04
Fix units in ADV_INTERVAL constant

DavesCodeMusings:
The aioble README already has this fix: https://github.com/micropython/micropython-lib/commit/57ce3ba95c65d9823e8cc5284003967ff621bd46

This brings the example makes a similar correction to the `temp_sensor.py` example.

### DIFF
--- a/micropython/bluetooth/aioble/examples/temp_sensor.py
+++ b/micropython/bluetooth/aioble/examples/temp_sensor.py
@@ -20,7 +20,7 @@ _ENV_SENSE_TEMP_UUID = bluetooth.UUID(0x2A6E)
 _ADV_APPEARANCE_GENERIC_THERMOMETER = const(768)
 
 # How frequently to send advertising beacons.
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks for the contribution.  Line 53 will also need fixing, as per the CI error.

Also, probably worth fixing the `examples/l2cap_file_server.py` example as well?

DavesCodeMusings:
@dpgeorge Unfortunately, I found out about the guidelines for commits _after_ I submitted a pull request. So yeah, there are some things I need to straighten out. Getting there... slowly. Still trying to figure out how I can do all this from the github.com web interface. Not so sure about the `git -s` signing bit. Can that be done in a browser or command-line only?

dpgeorge:
> Not so sure about the `git -s` signing bit. Can that be done in a browser or command-line only?

It should work through the browser now, I just enabled the relevant setting.  If you add a new commit it should be signed.

PR #937, opened 2024-11-18
unittest: Always use "raise" with an argument.

dpgeorge:
So this code can be compiled with the MicroPython native emitter, which does not support "raise" without any arguments.

### DIFF
--- a/python-stdlib/unittest/manifest.py
+++ b/python-stdlib/unittest/manifest.py
@@ -1,3 +1,3 @@
-metadata(version="0.10.3")
+metadata(version="0.10.4")
 
 package("unittest")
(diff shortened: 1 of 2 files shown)


andrewleech (approved):
Oh I wasn't aware of that with the native emitter, I'll keep that in mind for other things in future.

Regardless yes this is a straightforward change that does not change the intent of the functions at all.

It does highlight some slight style inconsistencies in the package already (`e`, `ex`) but that doesn't change the function and it's clearer to keep this change minimal.

dpgeorge:
Thanks for the review!

> Oh I wasn't aware of that with the native emitter, I'll keep that in mind for other things in future.

I try to stay away from `raise` without args, which will be hard to make it work (efficiently) with the native emitter (need to store the current exception somewhere).

I bumped into this issue for https://github.com/micropython/micropython/pull/16216

PR #941, opened 2024-12-02
aiohttp: fix header problem

jomasnash:
Latest version of aiohttp:
When making connection to a Websocket, the header argument is ignored.

The consequence is that you cannot make a connection to most online MQTT broker's over websocket
because they need the header entry: **"Sec-WebSocket-Protocol":"mqtt"** in the handshake of
the upgrade protocol.

See this small example code:
It connects to a MQTT broker and then sends the CONNECT mqtt packet.
Then it should get a reply of **opcode:2, data: b' \x02\x00\x00'** where 'data' is a CONNACK mqtt package
Because of the missing header entry **"Sec-WebSocket-Protocol":"mqtt"** most brokers will refuse the connection or
refuse to accept MQTT packets.

```python
import aiohttp
import asyncio

async def connect():
    url = "ws://test.mosquitto.org:8080"
    headers = {"Sec-WebSocket-Protocol":"mqtt"}
    connect_pkt = bytearray(b'\x10\x10\x00\x04MQTT\x04\x02\x00x\x00\x04fz54')
    
    async with aiohttp.ClientSession(headers=headers).ws_connect(url) as ws:
            print("Connected")
            await ws.send_bytes(connect_pkt)
            opcode, data = await ws.ws.receive()
            print(f"opcode:{opcode}, data{data}")

asyncio.run(connect())

```

### DIFF
--- a/python-ecosys/aiohttp/aiohttp/__init__.py
+++ b/python-ecosys/aiohttp/aiohttp/__init__.py
@@ -269,7 +269,7 @@ class ClientSession:
         return _WSRequestContextManager(self, self._ws_connect(url, ssl=ssl))
 
     async def _ws_connect(self, url, ssl=None):
-        ws_client = WebSocketClient(None)
+        ws_client = WebSocketClient(self._base_headers.copy())
         await ws_client.connect(url, ssl=ssl, handshake_request=self.request_raw)
         self._reader = ws_client.reader
         return ClientWebSocketResponse(ws_client)
--- a/python-ecosys/aiohttp/aiohttp/aiohttp_ws.py
+++ b/python-ecosys/aiohttp/aiohttp/aiohttp_ws.py
@@ -136,7 +136,7 @@ class WebSocketClient:
         return frame + payload
 
     async def handshake(self, uri, ssl, req):
-        headers = {}
+        headers = self.params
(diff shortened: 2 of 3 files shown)


Carglglz:
I've just tested this and it works as expected, hopefully it is merged for the next release, thanks! 👍🏼

python-ecosys/aiohttp/aiohttp/__init__.py
```diff
@@ -263,7 +263,7 @@ def ws_connect(self, url, ssl=None):
         return _WSRequestContextManager(self, self._ws_connect(url, ssl=ssl))
 
     async def _ws_connect(self, url, ssl=None):
-        ws_client = WebSocketClient(None)
+        ws_client = WebSocketClient(self._base_headers)
```
dpgeorge, line 266: I think this should be:
```py
         ws_client = WebSocketClient(self._base_headers.copy())
```
Otherwise the base headers will be mutated for each websocket connection.

dpgeorge: Do you want to make this change as part of your PR here?

dpgeorge:
Thanks for updating.  I tested the example code above and it works with this PR.

dpgeorge (approved):

PR #942, opened 2024-12-06
python-ecosys/requests/requests/__init__.py: Use the host in the redirect url, not the one in headers.

codemee:
removing the host header when redirect

The host in headers extracted from the original url may not be the same as the host in the redirect url. It shoud use the host in the redirect url, not the one in the headers.

### DIFF
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -181,8 +181,7 @@ def request(
     if redirect:
         s.close()
         # use the Host in the redirect URL
-        if "Host" in headers:
-            headers.pop("Host")
+        headers.pop("Host", None)
         if status in [301, 302, 303]:
             return request("GET", redirect, None, None, headers, stream)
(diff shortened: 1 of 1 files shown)


python-ecosys/requests/requests/__init__.py
```diff
@@ -180,6 +180,9 @@ def request(
 
     if redirect:
         s.close()
+        # use the Host in the redirect URL
+        if "Host" in headers:
+            headers.pop("Host")
```
dpgeorge, line 185: I think you could use `headers.pop("Host", None)` here to make it a little more efficient.

codemee: Thanks.

dpgeorge (approved):
Thanks for updating, looks good now.

dpgeorge:
Rebased and merged in 43ad7c58fd5fc247c255dacb37ab815ec212ee71

PR #945, opened 2024-12-10
micropython/umqtt.simple/umqtt: Add support for to umqtt.simple.

FallenPhoenix8:
**Summary**
1. Explicit UTF-8 Encoding:
- Both the topic and msg are now explicitly encoded using UTF-8 (topic.encode('utf-8') and msg.encode('utf-8')).
2. Accurate Size Calculation:
- The size of the message (sz) is calculated based on the byte length of the encoded topic and message.


FallenPhoenix8:
Ok, thanks for the information. I'll close the Pull Request.

PR #947, opened 2024-12-11
requests: Do not leak header modifications when calling request.

### LINKED ISSUE #839: SECURITY: Requests module leaks passwords & usernames for HTTP Basic Auth
While looking at the MicroPython `requests` module (on the git HEAD), I noticed this:

If you make a request with HTTP basic auth (a username/password) and did not specify a headers dict, then I believe the username and password would be added to the default headers to be used for **every subsequent HTTP request**.  Even if that request is to a completely different server, which you don't trust with your username and password.  That's probably not a good idea.

I haven't verified this, it's just from reading the code, but someone should probably look into it.

This is because there is `headers={}` in the function prototype, specifying a default for the `headers` parameter.  But (at least in cPython) that same dictionary will get reused for every call that doesn't explicitly specify a `headers` parameter.  So if the function changes the `headers` dictionary - such as by adding an `Authorization` header - that change will be there for every future call of the function.  This is a known dangerous part of the Python language, you're not the first people to write this kind of bug.

To fix this, you could keep the auth headers separate from the `headers` variable.  Something like this (totally untested!) commit: https://github.com/jonfoster/micropython-lib/commit/92e9b2208814faa22ba3ff2a19cbc8e0c5210a47 - feel free to use that as a starting point.

rweickelt:
The requests() function takes a headers dict argument (call-by-reference). This object is then modified in the function. For instance the host is added and authentication information. Such behavior is not expected. It is also problematic:

- Modifications of the header dictionary will be visible on the caller site.
- When reusing the same (supposedly read-only) headers object for differenct calls, the second call will apparently re-use wrong headers from the previous call and may fail.

This patch should also fix #839. Unfortunately the copy operation does not preserve the key order and we have to touch the existing test cases.

### DIFF
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -46,6 +46,8 @@ def request(
 ):
     if headers is None:
         headers = {}
+    else:
+        headers = headers.copy()
 
     redirect = None  # redirection url, None means no redirection
     chunked_data = data and getattr(data, "__next__", None) and not getattr(data, "__len__", None)
--- a/python-ecosys/requests/test_requests.py
+++ b/python-ecosys/requests/test_requests.py
@@ -102,11 +102,11 @@ def test_post_chunked_data():
 
 def test_overwrite_get_headers():
     response = requests.request(
-        "GET", "http://example.com", headers={"Connection": "keep-alive", "Host": "test.com"}
+        "GET", "http://example.com", headers={"Host": "test.com", "Connection": "keep-alive"}
     )
 
     assert response.raw._write_buffer.getvalue() == (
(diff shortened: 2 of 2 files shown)


projectgus (approved):
This looks good, thanks @rweickelt!

It looks like there's a slightly lower-resource-usage alternative where we don't add new entries to `headers` at all, but move all of those checks to when we're writing the response and write them all directly into the stream if they're not provided by the caller. But that approach would break up the flow of the function, and the improvement in memory usage would be very marginal.

rweickelt:
Will somebody complete this PR?

dpgeorge:
Thanks for the contribution, good to see a clean commit with good description and a test case.

PR #949, opened 2024-12-19
pymitter package added

olimaye:
The pymitter package provides capabilities for easy event subscription and publishing (emmitting).
The original version is available on [pypi](https://pypi.org/project/pymitter/) and supports the latest Python versions.

In consent with the author, the added source is a legacy branch of that project supporting Python >=2.7 up to 3.4. It would be extremely usefull to have it available for MicroPython.


olimaye:
Fixed ruff complains.

PR #954, opened 2024-12-30
drivers/radio/nrf24l01: Properly handle timeout.

marcsello:
The timeout condition was not handled before. Upon timeout, this caused the chip to stay active until another send command changed it's state.

Sometimes when it was unable to transmit the data, it got stuck in the tx fifo causing it to fill up over time, which set the TX_FULL flag in the STATUS register.

Since there was no exceptions raised, the user code could not differentiate a successful send or a timeout condition.

----

Note: timeout condition almost always indicate HW failure or improper control of the chip.


dpgeorge (approved):
Thanks for the patch!

PR #955, opened 2024-12-30
drivers/radio/nrf24l01: Optimized status reading.

marcsello:
The value of the STATUS register is always transmitted by the chip when reading any command. So a R_REGISTER command and the turnaround time can be spared by issuing a NOP command instead.

This implementation suggested by the datasheet.

This operation is compatible with both nRF24L01 and nRF24L01+.

----

In my testing, this reduced the STATUS read time from 41 μs to 27 μs. This allows for quicker polling, and thus more responsive operation. 

Old behavior:

Screenshot_2024-12-30_16-08-35

New behavior:

Screenshot_2024-12-30_16-06-31


dpgeorge (approved):
Looks good, thanks!

dpgeorge:
I also bumped the version here.

PR #972, opened 2025-02-07
python-ecosys/aiohttp: fix header - case-insensitivity

FuNK3Y:
According to RFC (https://datatracker.ietf.org/doc/html/rfc7230#section-3.2) header names are case-insensitive.

This PR makes sure that the module behaves consistently regardless of the casing of `Content-type` and `Content-Length` (other headers are not considered by the module)

Without this fix, the client seems to wait for the connection termination (~10 seconds) prior to returning any content if the casing of `Content-Length` is different.

### DIFF
--- a/python-ecosys/aiohttp/aiohttp/__init__.py
+++ b/python-ecosys/aiohttp/aiohttp/__init__.py
@@ -18,8 +18,14 @@ class ClientResponse:
     def __init__(self, reader):
         self.content = reader
 
+    def _get_header(self, keyname, default):
+        for k in self.headers:
+            if k.lower() == keyname:
+                return self.headers[k]
+        return default
+
     def _decode(self, data):
-        c_encoding = self.headers.get("Content-Encoding")
+        c_encoding = self._get_header("content-encoding", None)
         if c_encoding in ("gzip", "deflate", "gzip,deflate"):
             try:
                 import deflate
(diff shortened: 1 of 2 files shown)


FuNK3Y:
The pipeline is failing for an unrelated reason (version of GHA upload artifact)

python-ecosys/aiohttp/aiohttp/__init__.py
```diff
@@ -18,8 +18,14 @@ class ClientResponse:
     def __init__(self, reader):
         self.content = reader
 
+    def get_header(self, keyname, value=None):
+        try:
+            return next((v for k, v in self.headers.items() if k.lower() == keyname.lower()))
+        except StopIteration:
+            return value
```
dpgeorge, line 25: To make this more efficient for MicroPython I suggest:
```py
for k in self.headers:
    if k.lower() == keyname:
        return self.headers[k]
return default
```
Then make sure `keyname` is always lowercase when passed in.

FuNK3Y:
@dpgeorge thanks for the proposed changes. I implemented those and tested those successfully.

dpgeorge (approved):
Great, thanks for updating!

FuNK3Y:
@dpgeorge This should have generated a new version of the module - right? It shows here that the version increased: https://github.com/micropython/micropython-lib/blob/5b496e944ec045177afa1620920a168410b7f60b/python-ecosys/aiohttp/manifest.py#L3

Here is what I get when I try to install the package:
`mip.install("aiohttp",version="0.0.5")
Installing aiohttp (0.0.5) from https://micropython.org/pi/v2 to /lib
Package not found: https://micropython.org/pi/v2/package/6/aiohttp/0.0.5.json
Package may be partially installed`

There seems to be nothing above https://[micropython.org/pi/v2/package/6/aiohttp/0.0.3.json](https://micropython.org/pi/v2/package/6/aiohttp/0.0.3.json)

PR #976, opened 2025-02-24
github/workflows: Update actions/upload-artifact to v4.

dpgeorge:
Because v3 is now deprecated.


projectgus (approved):

projectgus:
LGTM, there is a breaking change in v4 where you [can't upload to the same name multiple times](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes) but I think this will only bite if someone pushes the same commit to two branches in their repo and enables package publishing for both branches.

PR #986, opened 2025-03-14
Fix #985 by accepting both str and bytes as arg for json.loads()

### LINKED ISSUE #985: unix-ffi/json json.loads() doesn't accept bytestring as arg but micropython internal json.loads() does
This breaks mip on unix if micropython-lib is installed.

asdil12:
Same as micropython's internal json lib does.


unix-ffi/json/json/__init__.py
```diff
@@ -413,4 +413,6 @@ def loads(
         kw["parse_int"] = parse_int
     if parse_constant is not None:
         kw["parse_constant"] = parse_constant
+    if isinstance(s, bytes):
```
dpgeorge, line 416: CPython also supports `bytearray` so this could be:
```
if isinstance(s, (bytes, bytearray)):
```

Also, please also update the documentation comment for the `loads` function to mention that `s` can be str, bytes or bytearray.

asdil12: Updated.

dpgeorge (approved):

PR #987, opened 2025-03-14
mip: Allow accessing index from local filesystem

asdil12:
Just set the index to `file://relative/path` or `file:///absolute/path`

### DIFF
--- a/micropython/mip/mip/__init__.py
+++ b/micropython/mip/mip/__init__.py
@@ -91,6 +91,14 @@ def _rewrite_url(url, branch=None):
 
 
 def _download_file(url, dest):
+    if url.startswith("file://"):
+        print("Copying:", dest)
+        src = url[7:]
+        _ensure_path_exists(dest)
+        with open(src, "b") as sf:
+            with open(dest, "wb") as f:
+                _chunk(sf, f.write)
+        return True
     response = requests.get(url)
     try:
         if response.status_code != 200:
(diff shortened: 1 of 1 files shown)


Josverl:
Can you explain what the problem is that you are trying to solve?

Your current description is so terse, I'm not sure that I understand what you are trying to solve.

asdil12:
There are two use cases I want to cover:

For local testing (especially but not necessarily limited to) using the unix port.
So you can locally generate an index from the repo and install packages from the index without starting a local http server.

And for packaging the complete index of this repo for the opensuse rpm (micropython-lib) without needing to launch an http server within the build environment. Previous we were using the makefile that was dropped from this repo so I decided to generate a local index and then install all packages contained in that index to the buildroot.

Josverl:
**Install of packages from a local filesystem**

I think I remember this being part of a `mpremote mip` PR I recently looked at, but I cant find it right now.
It might be worth double checking though.

**Packaging this repo for Suse**
I understand that you prefer you build to be simple and consistent.
However all the additions to this package will negatively affect all networked boards that have this package installed as it will take up space in the firmware image.
Its this really the only way this can be implemented, 
or is there a way to fix Suse's build process without these effect ?

asdil12:
If I get it right, `mpremote mip` is an independent implementation - only suited for installing stuff on an actual board, doesn't it.

In general I think this would be the cleanest way to implement this. I actually was surprized that it wasn't implemented, yet, which is why I created this PR.
The easiest way to "fix" this in the build env is having the patch of this PR in the sources of the RPM package.

I guess you are talking abount the 31 bytes of increased size of the (mpy-cross) compiled version of mpy?
I mean I could try to further optimize the code to make it even smaller if that is the issue.

asdil12:
Here is how to do it by modifying requests - it only costs 50 bytes:
```diff
diff --git a/python-ecosys/requests/requests/__init__.py b/python-ecosys/requests/requests/__init__.py
index 2951035..db22569 100644
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -71,6 +71,10 @@ def request(
         import tls

         port = 443
+    elif proto == "file:":
+        r = Response(open(url[7:], 'b'))
+        r.status_code = 200
+        return r
     else:
         raise ValueError("Unsupported protocol: " + proto)

asdil12:
Opened https://github.com/micropython/micropython-lib/pull/989

PR #989, opened 2025-03-24
requests: Support file:// scheme for local files

asdil12:
This allows installing packages using mip
from an index served from the local fs.

Usecases:

For local testing (especially but not necessarily limited to) using the unix port.
So you can locally generate an index from the repo and install packages from the index without starting a local http server.

And for packaging the complete index of this repo for the opensuse rpm (micropython-lib) without needing to launch an http server within the build environment. Previous we were using the makefile that was dropped from this repo so I decided to generate a local index and then install all packages contained in that index to the buildroot.


dpgeorge:
Thanks for the patch.

But, I'm not sure this is a good addition.  It doesn't seem to be supported by CPython's `requests` module?

Also, some work was done recently with mip to allow installing from the local filesystem, see https://github.com/micropython/micropython/pull/12477 .  Does that cover your use case?

PR #991, opened 2025-03-26
usb/cdc: writing small blocks should not require allocation

smurfix:
Only allocate a memoryview when the (first) write was partial.

### DIFF
--- a/micropython/usb/usb-device-cdc/usb/device/cdc.py
+++ b/micropython/usb/usb-device-cdc/usb/device/cdc.py
@@ -350,10 +350,9 @@ class CDCInterface(io.IOBase, Interface):
     ###
 
     def write(self, buf):
-        # use a memoryview to track how much of 'buf' we've written so far
-        # (unfortunately, this means a 1 block allocation for each write, but it's otherwise allocation free.)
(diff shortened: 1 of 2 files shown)


micropython/usb/usb-device-cdc/usb/device/cdc.py
```diff
@@ -350,10 +350,9 @@ def _rd_cb(self, ep, res, num_bytes):
     ###
 
     def write(self, buf):
-        # use a memoryview to track how much of 'buf' we've written so far
-        # (unfortunately, this means a 1 block allocation for each write, but it's otherwise allocation free.)
         start = time.ticks_ms()
-        mv = memoryview(buf)
+        # use a memoryview to track partial writes
```
dpgeorge, line 354: This is a neat optimisation, thanks!

But I suggest more detail in this comment, something like "if buf can't be fully written on the first attempt then it will be converted to a memoryview to track partial writes".

smurfix: Sure, will update.

dpgeorge (approved):

PR #992, opened 2025-03-29
micropython/drivers/sdcard.py: Add CRC7 values for SDCard.cmd() function calls

multicoder9:
Seems to be needed for some SDXC cards (SanDisk Ultra SDXC). FIXES [Issue #17039](https://github.com/orgs/micropython/discussions/17039)

Edit: I'm including the following from the issue to clarify this pull request.

_This is my first post to GitHub, but I've been working with micropython for about 10 years._

_Recently, I have been writing some micropython based sdcard dataloggers for a CANbus project, mostly with rp2040 and rp2350 devices. I have been using the [sdcard.py](https://github.com/micropython/micropython-lib/blob/e4cf09527bce7569f5db742cf6ae9db68d50c6a9/micropython/drivers/storage/sdcard/sdcard.py) source file as a driver in my projects._

_Today I noticed that switching from my 32GB SanDisk Ultra SDHC which worked just fine to a 32GB SanDisk Ultra SDXC caused it to fail on initialization. Others have noticed the same thing caused by missing CRC7 such as [issue #12687](https://github.com/orgs/micropython/discussions/12687). But none of them resulted in the fix getting merged back upstream. Here is a trivial diff for sdcard.py which adds the CRC7 values needed to get it working with my SDXC card, while still appearing to work on the older cards I've tested it on. Just for completeness, I've added a commented out minimalist Python function to compute these CRC7 values. (A quick internet search didn't pull up any Python functions for this.)_

_I figured I'd try to get the minimal changes merged upstream while more significant rewrites like the one in the linked thread above work their way through the process._

### DIFF
--- a/micropython/drivers/storage/sdcard/sdcard.py
+++ b/micropython/drivers/storage/sdcard/sdcard.py
@@ -295,14 +295,14 @@ class SDCard:
         assert nblocks and not err, "Buffer length is invalid"
         if nblocks == 1:
             # CMD24: set write address for single block
-            if self.cmd(24, block_num * self.cdv, 0) != 0:
+            if self.cmd(24, block_num * self.cdv, 0x7F) != 0:
                 raise OSError(5)  # EIO
 
             # send the data
             self.write(_TOKEN_DATA, buf)
         else:
             # CMD25: set write address for first block
-            if self.cmd(25, block_num * self.cdv, 0) != 0:
+            if self.cmd(25, block_num * self.cdv, 0x7F) != 0:
                 raise OSError(5)  # EIO
             # send the data
             offset = 0
--- a/micropython/drivers/storage/sdcard/sdcard.py
+++ b/micropython/drivers/storage/sdcard/sdcard.py
@@ -37,6 +37,20 @@ _TOKEN_CMD25 = const(0xFC)
 _TOKEN_STOP_TRAN = const(0xFD)
 _TOKEN_DATA = const(0xFE)
 
+## This function computes crc7 for SDcard commands
+##   It is not optimized, so do not call it where performance is critical
+##   You should usually be precomputing these values and hardcode them
+#def crc7( cmd:bytes ):
+#    crc = 0
+#    for value in cmd:
+#        for i in range(8):
+#            crc <<= 1
+#            if (value&0x80)^(crc&0x80):
+#                crc ^= 0x09
+#            value <<= 1
+#    return ((crc<<1)|1)&0xFF
+#def cmd_args_to_bytes( cmd:int, args:int ):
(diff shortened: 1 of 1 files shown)


Josverl:
Thanks for the PR.

It would be nice if you could add some information about these magic values that make things work better.
Either in comments, or by defining it as a const with an explanatory name.

That would make it possible for future maintainers to understand better wat the code's intent is.

scruss:
... or why this is better than the long-pending, mostly-reviewed [SDcard sdcard.py rewrite for efficiency and function by mendenm · Pull Request #765 · micropython/micropython-lib](https://github.com/micropython/micropython-lib/pull/765)

multicoder9:
Sorry, I started this [as an issue here](https://github.com/orgs/micropython/discussions/17039). I'm new to GitHub and didn't realize that the pull request would start another conversation. Is there a way to link the conversations together besides just posting a link?

multicoder9:
> ... or why this is better than the long-pending, mostly-reviewed [SDcard sdcard.py rewrite for efficiency and function by mendenm · Pull Request #765 · micropython/micropython-lib](https://github.com/micropython/micropython-lib/pull/765)

As for that, my thought is simple. This gets the existing code working with SDXC cards by doing nothing more than filling in the correct CRC7 values according to the spec. In the entire file, I've done nothing more than set 11 numbers that were unset (10 were 0, and one was 0xFF). I was hoping this would allow the existing sdcard.py to work for the users having problems with it while the "long-pending, mostly-reviewed" pull request continues its process of getting reviewed. But that is a major rewrite for overall improvement. Mine is simply correcting the fact that the CRC7 values in the existing one are wrong, and while many older cards are okay with this, some newer ones aren't. 

So I hoped it would get approved in the meantime, until the bigger cleanups are approved.  Although one thing that is better about this than the sdcard.py in the PR#765 is that one is computing CRC's dynamically that can't possibly change. All the parameters are hardcoded, so the CRC will always be a constant value. On a desktop, I would have written it that way, but on a microcontroller, I would normally precompute any values that are known at coding time and will never change. 

And hardcoding it has the advantage of not using any memory for storing a globally computed value, and keeping the value next to the hard coded parameters that it was computed from. If anyone changes a hard coded argument, it will be good to have the hard coded CRC next to it to update too.

multicoder9:
FYI I just found two more self.cmd lines in writeblocks() with zero CRC7 values. I changed them to 0x7F and have confirmed that writeblocks now works with my SDXC cards.

dpgeorge:
Thanks @multicoder9 for the detailed explanation above.  I agree that it's good to just have something that works.  But hopefully #765 can be merged very soon, and that will supersede this PR.

dpgeorge:
CRC7 was added in commit 446038da9b03ad2aa67e256134937e19ff8d0a73

PR #995, opened 2025-04-06
unix-ffi/machine/timer: Use libc if librt is not present

basvd:
Newer implementations of libc integrate the functions from librt, for example glibc since 2.17 and uClibc-ng. So if the librt.so cannot be loaded, it can be assumed that libc contains the expected functions.


dpgeorge (approved):
Thanks for the contribution.  This looks good.

Issue #997, opened 2025-04-10
inspect.isgenerator is broken


smurfix:
This doesn't quite match what I'd expect.
```
MicroPython v1.24.0-400.g160c2931f.dirty on 2025-04-10; linux [GCC 14.2.0] version
Use Ctrl-D to exit, Ctrl-E for paste mode
>>> async def fn():
...     pass
...
>>> x=fn()
>>> inspect.isgeneratorfunction(x)
False  ## correct
>>> inspect.isgenerator(x)
False  ## wrong
>>> x
<generator object 'fn' at 7f5b2a9534a0>
>>> inspect.isgeneratorfunction(fn)
True  ## correct
>>> inspect.isgenerator(fn)
True  ## wrong
>>> 
```

dpgeorge:
Yes, you're quite right, it's missing those parenthesis.

See #998 for a fix.

PR #998, opened 2025-04-11
inspect: Fix isgenerator logic.

dpgeorge:
Fix the `isgenerator()` logic.

Also optimise both `isgenerator()` and `isgeneratorfunction()` so they use the same lambda, and don't have to create it each time they are called.

And add some basic unit tests for this package.

Fixes issue #997.

### DIFF
--- a/python-stdlib/inspect/inspect.py
+++ b/python-stdlib/inspect/inspect.py
@@ -25,6 +25,11 @@ def isgenerator(obj):
     return isinstance(obj, type((_g)()))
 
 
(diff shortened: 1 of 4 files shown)


smurfix:
Looks sensible, thanks.

smurfix:
While you're at it, might you be persuaded to add
```
iscoroutinefunction = isgeneratorfunction
iscoroutine = isgenerator
```
as per https://github.com/micropython/micropython/pull/11526 ?

dpgeorge:
> While you're at it, might you be persuaded to add

Good idea, now done.  See what you think.

PR #1019, opened 2025-06-04
Add ENOTCONN to errno

thatbudakguy:
ENOTCONN seems to be present in core micropython (v1.24.1), at least on my unix build, but not here – installing `errno` via `mip` resulted in me being unable to import it anymore.


dpgeorge (approved):
Looks OK, thanks!

PR #1020, opened 2025-06-04
logging: Allow logging.exception helper to handle tracebacks.

thatbudakguy:
Although `Logger.exception` supports passing exception info with `exc_info`, when you use `logging.exception`, keyword arguments are not forwarded to the root logger, which makes passing `exc_info` raise TypeError.

### DIFF
--- a/python-stdlib/logging/logging.py
+++ b/python-stdlib/logging/logging.py
@@ -202,8 +202,8 @@ def critical(msg, *args):
     getLogger().critical(msg, *args)
 
 
-def exception(msg, *args):
-    getLogger().exception(msg, *args)
+def exception(msg, *args, exc_info=True):
+    getLogger().exception(msg, *args, exc_info=exc_info)
 
 
 def shutdown():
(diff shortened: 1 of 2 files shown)


python-stdlib/logging/logging.py
```diff
@@ -202,8 +202,8 @@ def critical(msg, *args):
     getLogger().critical(msg, *args)
 
 
-def exception(msg, *args):
-    getLogger().exception(msg, *args)
+def exception(msg, *args, **kwargs):
```
dpgeorge, line 205: Would it be better to do:
```py
def exception(msg, *args, exc_info=True):
   getLogger().exception(msg, *args, exc_info)
```
?

That way it's clear what args are supported.  In particular, general keyword args are **not** supported, only `exc_info` is.

thatbudakguy: Good idea; I've updated it!

dpgeorge (approved):

PR #1022, opened 2025-06-11
python-ecosys/debugpy: Add VS Code debugging support for MicroPython.

andrewleech:
This implementation provides a Debug Adapter Protocol (DAP) server that enables VS Code to debug MicroPython code with full breakpoint, stepping, and variable inspection capabilities.

Features:
- Manual breakpoints via debugpy.breakpoint()
- Line breakpoints set from VS Code
- Stack trace inspection
- Variable scopes (locals/globals)
- Source code viewing
- Stepping (into/over/out)
- Non-blocking architecture for MicroPython's single-threaded environment
- Conditional debug logging based on VS Code's logToFile setting

Implementation highlights:
- Uses MicroPython's sys.settrace() for execution monitoring
- Handles path mapping between VS Code and MicroPython
- Efficient O(n) fibonacci demo (was O(2^n) recursive)
- Compatible with MicroPython's limited frame object attributes
- Comprehensive DAP protocol support

Files:
- README.md: Setup and usage instructions
- debugpy/: Core debugging implementation
- test_vscode.py: VS Code integration test
- dap_monitor.py: Protocol debugging utility

Usage:
```python
import debugpy
debugpy.listen()          # Start debug server
debugpy.debug_this_thread()  # Enable tracing
debugpy.breakpoint()      # Manual breakpoint
```

---

Currently only tested on unix port with updates to settrace in https://github.com/micropython/micropython/pull/8767
Should work on any network enabled device however?


python-ecosys/debugpy/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+Create a `.vscode/launch.json` file in your project:
+
+```json
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Attach to MicroPython",
+            "type": "python",
```
Josverl, line 61: "type": "debugpy",

python-ecosys/debugpy/debugpy/server/pdb_adapter.py
```diff
@@ -0,0 +1,285 @@
...
+        else:
+            return []
+            
+        for name, value in var_dict.items():
+            # Skip private/internal variables
+            if name.startswith('__') and name.endswith('__'):
+                continue
+                
+            try:
+                value_str = str(value)
```
Josverl, line 238: I think this should be `value_str = repr(value)` . without that strings show without quotes etc.

python-ecosys/debugpy/README.md
andrewleech: Thanks, yeah I fixed that in the examples file, missed it here

python-ecosys/debugpy/README.md
```diff
@@ -0,0 +1,172 @@
...
+    ]
+}
+```
+
+### Testing
+
+1. Build the MicroPython Unix coverage port:
+   ```bash
+   cd ports/unix
+   make CFLAGS_EXTRA="-DMICROPY_PY_SYS_SETTRACE=1"
```
Josverl, line 84: this flag `MICROPY_PY_SYS_SETTRACE` conflicts with the referenced PR where this is already set unconditionally

Josverl:
> I'd be interested to hear any notes about what was confusing / difficult to get going to feed into docs.

I first got in a tangle by 
- checking out the micropython repo on PR#8767
- then checking out ./lib/micropython-lib on this PR 

blocked most attempts at building as `make submodules` breaks 
After reverting ./lib/micropython-lib, and cloning this PR to a separate folder things got better 

Building a firmware with "the updates to settrace in https://github.com/micropython/micropython/pull/8767"
I still ran into build problems , essentially by a double definition of `MICROPY_PY_SYS_SETTRACE` .
In the readme of this PR it is part of the instructions , while in the "settrace  PR" it is unconditionally defined. These two clash.

Matching up the paths vscode / remote paths, was not to difficult.

Open questions/ more play time needed  : 
 - I am confused by the read-only source used by the debugger to step though. I can see it is retrieved from the remote, but not sure why this is needed , or how to configure it. 
 - why JustMycode : false is needed 
 - how debugging would work with a .mpy on the remote  
 - How to Enable VS Code's built-in DAP logging: ( I do not have these settings in my VSCode) 
 - Why the DAP_Monitor does not terminate or reset itself when the remote target exitst

Josverl:
@andrewleech 
I have made some more notes after some additional testing I have done.
I have put them in  [a Gist notes.md](https://gist.github.com/Josverl/72d11a63ab7611296addcc171528e86f) for now as I do not want to drop too many comments  here.
Would it make sense to start a Discussion on this topic , or do you prefer everything here ?

andrewleech:
> > I'd be interested to hear any notes about what was confusing / difficult to get going to feed into docs.
> 
> I first got in a tangle by 
> - checking out the micropython repo on PR#8767
> - then checking out ./lib/micropython-lib on this PR 
> 
> blocked most attempts at building as `make submodules` breaks 
> After reverting ./lib/micropython-lib, and cloning this PR to a separate folder things got better 

Yes the automatic make submodules in mpbuild pretty much breaks development within micropython-lib, I haven't figured out why way to resolve this without a newer/smarter algorithm in the submodules target (which I haven't written yet)

> Building a firmware with "the updates to settrace in https://github.com/micropython/micropython/pull/8767"
> I still ran into build problems , essentially by a double definition of `MICROPY_PY_SYS_SETTRACE` .
> In the readme of this PR it is part of the instructions , while in the "settrace  PR" it is unconditionally defined. These two clash.

Ah yeah, the readme in micropython-lib doesn't necessarily know about the updates in the micropython PR, 


> Matching up the paths vscode / remote paths, was not to difficult.
> 
> Open questions/ more play time needed  : 
>  - I am confused by the read-only source used by the debugger to step though. I can see it is retrieved from the remote, but not sure why this is needed , or how to configure it. 

I haven't figured out the read only src either, it's annoying though. I think it's related to path matching issue between local and remote copies.

>  - why JustMycode : false is needed 

Not sure if it is really needed?

>  - how debugging would work with a .mpy on the remote  

The path matching would need to be smart enough to match PC side py to remote module (should be possible) and the mpy would need to have been made with opt=0 otherwise line numbers are stripped.

>  - How to Enable VS Code's built-in DAP logging: ( I do not have these settings in my VSCode) 

Ah yeah I didn't actually test that!

>  - Why the DAP_Monitor does not terminate or reset itself when the remote target exits

Yeah that would be helpful, I'll add it

>  
>  📚 likely relevant : [What is the Debug Adapter Protocol?](https://microsoft.github.io/debug-adapter-protocol/overview) 

Definitely a good reference to add to the docs thanks

Josverl:
> > Why the DAP_Monitor does not terminate or reset itself when the remote target exits

> Yeah that would be helpful, I'll add it

I did some work to terminate on disconnect 
See the [PR to your fork](https://github.com/andrewleech/micropython-lib/pull/1)

Josverl:
I was able to add (a first rough draft) of [support for `frame.f_locals`](https://github.com/Josverl/micropython/tree/pdb_support) on top of the pdb_support branch.
image

andrewleech:
> I was able to add (a first rough draft) of [support for `frame.f_locals`](https://github.com/Josverl/micropython/blob/c923b54baf53bbf6395d058cbac5683c2d70ee45/py/profile.c#L135-L167) on top of the pdb_support branch.

Oh wow awesome, I wanted to tackle basically exactly that! Looks great :-D
I was only just about to test my latest changes to this micropython-lib branch; hopefully the read-only window no longer pops up...

andrewleech:
Pushed an update folding in the enhancement lineage that had been developed alongside this branch, plus several fixes found while building tooling on top of it. Fast-forward, no history rewritten.

Two are correctness fixes that affect anyone using this today:

**Nested message pumps deadlocked the session.** `process_pending_messages()` set a 1 ms socket timeout and restored blocking mode in its `finally`. The trace function calls it on entry to every new frame, so handling a message re-enters it, and the inner call's `finally` put the socket back into blocking mode underneath the outer pump. That pump's next `recv()` then waited for a message the client won't send until it has seen an event the pump itself produces. When it lost the race the session hung before `wait_for_client()` returned, so the target never ran and no `stopped` event was ever produced. It also cost ~100 ms per traced statement when it didn't deadlock. The nesting is now tracked rather than the timeout saved and restored, since MicroPython sockets have no `gettimeout()`.

**`listen()` now returns the bound endpoint before accepting.** It previously bound, blocked in `accept()`, and handled `initialize` before returning, so a caller could only learn the endpoint after a client had already connected to it — unusable for anything that has to read the address in order to attach. `accept()` and the `initialize` handshake move into `wait_for_client()`, matching CPython debugpy, where `listen()` reports the endpoint and `wait_for_client()` blocks. Relatedly, `port=0` now raises where `getsockname()` can't report the assigned port instead of silently substituting the default and advertising an address nothing is bound to.

Also included: DAP messages split across `recv()` calls are reassembled; a runtime capability probe (`settrace`/`save_names`/`set_local`/`f_back`) so tooling can ask what a firmware actually supports rather than inferring from a build name; a read-only hint on locals where the firmware can't write them back; `evaluate` honouring its `context` field so statements work from the Debug Console; and the pause / set-variable / local-modification work from the parallel lineage.

Exercised by a host-side DAP harness (a fake VS Code client driving real sessions against a built unix firmware): 278 passed. Happy to split any of this out if you'd rather review it separately.

PR #1027, opened 2025-06-29
pathlib: Fix glob() to return Path objects rather than strings.

RotatingLlama:
Fix glob() to return Path objects rather than bare strings, same as CPython https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob

### DIFF
--- a/python-stdlib/pathlib/pathlib.py
+++ b/python-stdlib/pathlib/pathlib.py
@@ -126,7 +126,7 @@ class Path:
         for name, mode, *_ in os.ilistdir(path):
             full_path = path + _SEP + name
             if name.startswith(prefix) and name.endswith(suffix):
-                yield full_path
+                yield Path(full_path)
             if recursive and mode & 0x4000:  # is_dir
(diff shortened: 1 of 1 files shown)


andrewleech:
This discrepancy with cpython caught me out recently too, thanks for submitting the fix. 

Would you be able to update the commit message to match the repo standard needs patterns and force-push to the same branch? That would fix the ci failure.

RotatingLlama:
No probs, it's very minor. This is actually my first pull request and I spent about half an hour yesterday trying to fix the error, but couldn't figure out how to do it. Would you give me a pointer, please?

RotatingLlama:
I've now spent literal hours of my life trying to understand how to fix this pull request, and am still no further forward.  I'm not willing to waste any more time for the sake of such a minor correction.  I'm going to see if I can delete it for the sake of keeping the place tidy, but if it won't let me do that either then I'm out.

dpgeorge:
I redid this in #1120.

RotatingLlama:
Happy days!  Sorry I failed so utterly to get my own PR working 🤦

PR #1031, opened 2025-07-05
micropython/usb/usb-device/usb/device/core.py: Add alternate setting support plus a few fixes and improvements.

HLammers:
### Summary
* Added `bAlternateSetting` as argument to `Descriptor.interface`, because it is used for USB MIDI 2.0
* Fixed the loop reserving other string indexes used by built-in drivers `while len(strs) < builtin_driver.str_max:` to `while len(strs) < builtin\_driver.str\_max - 1:` (row 169) because `str_max` equals to one more than the highest string descriptor index value used by any built-in descriptor according to the [docs](https://docs.micropython.org/en/latest/library/machine.USBDevice.html) and added a comment that this loop is probably unnecessary or even wrong
* Removed `import time` (not used)
* Removed unused/duplicate `_Device.config` argument `config_str`
* Some code optimisations (in particular to the `Buffer` class)
### Testing
* Tested with micropython/usb/examples/device/midi_example.py
* Tested the Buffer class with micropython/usb/usb-device/usb/device/tests/test_core_buffer.py
* Tested with a soon to be made public multi-port MIDI module draft which adds port names to `strs`
None

### DIFF
--- a/micropython/usb/usb-device/usb/device/core.py
+++ b/micropython/usb/usb-device/usb/device/core.py
@@ -3,10 +3,11 @@
 # These contain the classes and utilities that are needed to
 # implement a USB device, not any complete USB drivers.
 #
-# MIT license; Copyright (c) 2022-2024 Angus Gratton
+# MIT license; Copyright (c) 2022-2024 Angus Gratton, 2025 Harm Lammers
 from micropython import const
 import machine
 import struct
+import time
 
 try:
     from _thread import get_ident
@@ -108,7 +109,6 @@ class _Device:
         device_class=0,
         device_subclass=0,
         device_protocol=0,
-        config_str=None,
         max_power_ma=None,
         remote_wakeup=False,
     ):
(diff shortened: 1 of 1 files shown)


HLammers:
The updated version of `core.py` is used in this project, which also shows that it is working: https://github.com/orgs/micropython/discussions/17640

micropython/usb/usb-device/usb/device/core.py
```diff
@@ -3,10 +3,11 @@
 # These contain the classes and utilities that are needed to
 # implement a USB device, not any complete USB drivers.
 #
-# MIT license; Copyright (c) 2022-2024 Angus Gratton
+# MIT license; Copyright (c) 2022-2024 Angus Gratton, 2025 Harm Lammers
 from micropython import const
 import machine
 import struct
+import time
```
projectgus, line 10: Confused by this as PR description says:

> Removed import time (not used)

... and it is indeed not used, from what I can see?

micropython/usb/usb-device/usb/device/core.py
```diff
@@ -166,7 +166,11 @@ def maybe_set(value, idx):
         # Keep track of the interface and endpoint indexes
         itf_num = builtin_driver.itf_max
         ep_num = max(builtin_driver.ep_max, 1)  # Endpoint 0 always reserved for control
-        while len(strs) < builtin_driver.str_max:
+        while (
+            len(strs) < builtin_driver.str_max - 1
+        ):  # This is possibly unnecessary or wrong because
+            # https://docs.micropython.org/en/latest/library/machine.USBDevice.html
+            # states all string values except index 0 should be plain ASCII
```
projectgus, line 173: What's happening here is that we're expanding the `strs` list to cover the built-in driver indexes, because these str index values are actually handled by C code and never read by Python code. By putting something here we make sure that the indexes elsewhere in the driver match what the host sees.

This is why the documentation says the strings arguments should be strings - the values provided by the Python caller should be strings, these values are internal placeholders to pad the index out.

I assume the off-by-one is a real bug, though - I'm guessing you found that strings were reported wrong without this?

projectgus (commented):
Hey @HLammers,

Sorry I never saw this when you originally submitted it. Appreciate you submitting these improvements, thank you!

I have a couple of questions but generally this looks good to me.

PR #1032, opened 2025-07-14
tarfile: Make it possible to extract nested tarfiles in memory

woju:
Make it possible to extract nested archives, which are used in e.g. Mender artifacts. See commit message for details and a (simplified) example.


dpgeorge:
For reference, this used to work, but commit 2ca1527321d7e6b65bcab45d304be82d65f3c4f4 optimised `skip` to not use too much memory.

dpgeorge:
> I thought this was already the case, because as it is, it doesn't work without 2-argument `readinto` at all, even the outer tarfile can't be extracted.

Yes, you're right, all fileobj's that it uses must support 2-arg `readinto`.  For the most part that's OK, all C-based MicroPython streams will implement that.  But Python-based files/stream won't.

For example, if there's a tar file on the host PC and you use `mpremote mount .` and then at the REPL try to use `tarfile.TarFile` to open the tar file on the mounted host PC, it'll fail with the same problem as this PR is addressing:
```
  File "tarfile/__init__.py", line 129, in __next__
  File "tarfile/__init__.py", line 106, in next
  File "tarfile/__init__.py", line 59, in skip
TypeError: function takes 2 positional arguments but 3 were given
```

So, instead of trying to add the 2-arg form to all streams/files, better to fix it once here so that it doesn't use the 2-arg form.

PR #1035, opened 2025-07-30
usb: propagate submit_xfer result

hyx0329:
The result is used in the hid keyboard module but was not actually propagated. With this result, the report double buffering will work correctly.


Josverl:
Could you add a test to verify this in CI?

Or is this already covered by another test?

hyx0329:
I've updated the PR with the proposed changes. It should work now.

As to the test, I have no idea how to do it with software only. Core depends on `machine.USBDevice`.

projectgus (approved):
Thanks @hyx0329, this fix looks good to me.

We don't currently have any way to test the micropython-lib usb classes in CI. Running the examples manually is the best we can do. :face_with_diagonal_mouth:

PR #1038, opened 2025-07-31
aiohttp: ws: fix truncated read of payload

mirko:
aiohttp: fix partial read on websocket resulting in truncated payload

Attempting to read N bytes from a socket does not guarantee to actually
read N bytes, even if >= N bytes were written onto the peer socket.

This especially becomes an issues when attempting to read larger
messages at once, which then potentially can't be read all at once,
resulting in truncated payloads.

Fix that by reading as long / often until expected length was actually
received.


dpgeorge:
This should be fixed by #1034 (that was merged yesterday).

PR #1040, opened 2025-08-04
python-ecosys/uminio: Add uminio package.

paluigi:
uminio is a MicroPython client for MinIO object storage.
This commit add the uminio folder and all its content in the python-ecosys folder.
Original implementation for AWS S3:
https://github.com/DanielMilstein/uboto3/

### DIFF
--- a/python-ecosys/uminio/README.md
+++ b/python-ecosys/uminio/README.md
@@ -1,0 +1,129 @@
+# uminio
+
+`uminio` is a MicroPython library designed to facilitate uploading files directly from a MicroPython-enabled device (like an ESP32 or ESP8266) to MinIO object storage. It implements the necessary AWS Signature Version 4 for an S3 PUT Object request. This allows you to store data, sensor readings, images, or any other files from your microcontroller projects in the cloud.
+
+Forked from `uboto3` [https://github.com/DanielMilstein/uboto3](https://github.com/DanielMilstein/uboto3) 
+
+## Features
+
+* **Direct S3 Upload:** Upload files directly to an MinIO bucket without needing an intermediary server.
+* **AWS Signature V4:** Implements the required request signing process.
+* **HMAC-SHA256:** Includes a MicroPython-compatible HMAC-SHA256 implementation for signing.
+* **Time Synchronization:** Includes a helper function to synchronize the device's time using NTP, which is crucial for MinIO request signing.
+* **Minimal Dependencies:** Built with standard MicroPython libraries like `urequests`, `uhashlib`, `ubinascii`, `utime`, and `network`.
+
+## Requirements
+
+* MicroPython firmware flashed on your device.
+* Network connectivity (WiFi) configured on the device.
+* The following MicroPython libraries:
+    * `urequests`
+    * `uhashlib`
+    * `ubinascii`
+    * `utime`
+    * `network`
+    * `ntptime` (for time synchronization)
+
+## Setup
+
+1.  **Copy `__init__.py`:** Create a `uminio` folder in the `/lib` directory of your MicroPython device and copy the `__init__.py` file into it.
+2.  **MinIO Credentials & Configuration:**
+    Import the `MinioClient` class in your MicroPython script and configure it with your MinIO server details. You can do this by setting the following variables in your script:
+    ```python
+    from uminio import MinioClient
+    # --- MinIO Client Configuration ---
+    MINIO_ENDPOINT = "192.168.1.100:9000"  # Your MinIO server IP address and port
+    MINIO_ACCESS_KEY = "YOUR_ACCESS_KEY"      # Your MinIO access key
+    MINIO_SECRET_KEY = "YOUR_SECRET_KEY"      # Your MinIO secret key
+    MINIO_REGION = "eu-east-1"                # The region for your MinIO server
+    MINIO_USE_HTTPS = False               # Set to True if your MinIO server uses HTTPS
+
+    mc = MinioClient(
+        endpoint=MINIO_ENDPOINT,
+        access_key=MINIO_ACCESS_KEY,
+        secret_key=MINIO_SECRET_KEY,
+        region=MINIO_REGION,
+        use_https=MINIO_USE_HTTPS,
+    )
+    ```
+    **Important Security Note:** Hardcoding credentials directly into the script is generally not recommended for production environments. Consider alternative methods for managing secrets on your device if security is a major concern.
+
+3.  **IAM Permissions:**
+    Ensure the MinIO user associated with the `MINIO_ACCESS_KEY` and `MINIO_SECRET_KEY` has the necessary permissions to put objects into the specified bucket.
+
+
+## Usage Example
+
+Here's how to use `uminio` to upload a local file from your MicroPython device to MinIO:
+
+```python
+import network
+import time
+from uminio import MinioClient
+# --- MinIO Client Configuration ---
+MINIO_ENDPOINT = "192.168.1.100:9000"  # Your MinIO server IP address and port
+MINIO_ACCESS_KEY = "YOUR_ACCESS_KEY"      # Your MinIO access key
+MINIO_SECRET_KEY = "YOUR_SECRET_KEY"      # Your MinIO secret key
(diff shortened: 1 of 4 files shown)


paluigi:
I tried to figure our why the lint/formatting fails...but I'm lost.

dpgeorge:
If you look here: https://github.com/micropython/micropython-lib/actions/runs/16717239939/job/47313198532?pr=1040 it shows you the problems.  Basically you need to put a few things on one line, instead of over many lines.

python-ecosys/uminio/uminio/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+# Based on the uboto3 library for AWS S3 by DanielMilstein.
+# Modified to support self-hosted MinIO endpoints.
+#
+# MinIO Python client: https://github.com/minio/minio-py (for reference)
+# Original uboto3: https://github.com/DanielMilstein/uboto3
+
+import urequests
+import uhashlib
+import ubinascii
+import utime
```
agatti, line 12: Most, if not all, of these imports can now be accessed without the `u` prefix.

python-ecosys/uminio/uminio/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+            k_secret_bytes, date_stamp_string.encode("utf-8")
+        )
+        k_region_bytes = self._hmac_sha256(k_date_bytes, self.region.encode("utf-8"))
+        k_service_bytes = self._hmac_sha256(
+            k_region_bytes, service_name_string.encode("utf-8")
+        )
+        k_signing_bytes = self._hmac_sha256(k_service_bytes, b"aws4_request")
+        return k_signing_bytes
+
+    def sync_time(self) -> None:
```
agatti, line 100: If knowing the current time as retrieved by the NTP server is important, this function could be changed to return the `time.gmtime()` tuple instead.

python-ecosys/uminio/uminio/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+        """
+        Synchronizes the device's real-time clock with an NTP server.
+        This is crucial for generating a valid signature.
+        """
+        print("Synchronizing time with NTP server...")
```
agatti, line 105: Maybe these `print()` statements could be taken out.  Same applies to `upload_file` too.

python-ecosys/uminio/uminio/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+                "Current UTC from device: {:04d}-{:02d}-{:02d} {:02d}:{:02d}:{:02d}".format(
+                    now_utc[0],
+                    now_utc[1],
+                    now_utc[2],
+                    now_utc[3],
+                    now_utc[4],
+                    now_utc[5],
+                )
+            )
+        except Exception as e:
```
agatti, line 120: I'd let the exception propagate rather than ignoring it here.

If the lack of exception propagation is critical then this bit could be either a simple `pass` statement or, if the function was modified to return the new timestamp, make it `return None` instead.

python-ecosys/uminio/uminio/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+        """
+        Uploads a file to a MinIO bucket using AWS Signature V4.
+
+        :param local_file_path: The path to the local file to upload.
+        :param bucket: The name of the MinIO bucket to upload to.
+        :param object_name: The name of the object as it will be stored in MinIO.
+        :param content_type: The MIME type of the file.
+        :return: True if upload was successful (HTTP 200), False otherwise.
+        """
+        try:
```
agatti, line 139: These exceptions could also be let to propagate (file I/O and network).  As far as I can see there's no recovery mechanism besides a print statement and returning a value, that can be handled by catching the exception upstream.

It also saves quite some space when compiled into bytecode :)

python-ecosys/uminio/uminio/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+        canonical_headers_list.sort(key=lambda item: item[0])
+
+        canonical_headers_str = ""
+        signed_headers_list = []
+        for key, value in canonical_headers_list:
+            canonical_headers_str += f"{key}:{value.strip()}\n"
+            signed_headers_list.append(key)
+        signed_headers_str = ";".join(signed_headers_list)
+
+        canonical_request = (
```
agatti, line 179: If these are all guaranteed to be strings, `"\n".join(method, canonical_uri, canonical_querystring, canonical_headers_str, signed_headers_str, payload_hash_hex)` is actually shorter by ~19 bytes when compiled into bytecode.

Even with `"\n".join(map(str, (method, ...)))` to pre-convert variables to strings, it is still shorter by 7 bytes.

python-ecosys/uminio/uminio/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+        credential_scope = f"{datestamp}/{self.region}/{service}/aws4_request"
+
+        hashed_canonical_request_bytes = uhashlib.sha256(
+            canonical_request.encode("utf-8")
+        ).digest()
+        hashed_canonical_request_hex = ubinascii.hexlify(
+            hashed_canonical_request_bytes
+        ).decode()
+
+        string_to_sign = (
```
agatti, line 199: This one could be converted into `"\n".join(...)` too.

python-ecosys/uminio/README.md
```diff
@@ -0,0 +1,129 @@
...
+    ```
+    **Important Security Note:** Hardcoding credentials directly into the script is generally not recommended for production environments. Consider alternative methods for managing secrets on your device if security is a major concern.
+
+3.  **IAM Permissions:**
+    Ensure the MinIO user associated with the `MINIO_ACCESS_KEY` and `MINIO_SECRET_KEY` has the necessary permissions to put objects into the specified bucket.
+
+
+## Usage Example
+
+Here's how to use `uminio` to upload a local file from your MicroPython device to MinIO:
```
agatti, line 57: Maybe you can just point to the example file in the same directory rather than reproducing it verbatim here.

agatti (commented):
Thanks for the PR.

I've had a quick look at the code and there are a couple of places where I believe this could be improved.  It could also help if you'd squash the two commits into just one, since the second one is just minor changes to the source formatting.

python-ecosys/uminio/README.md
```diff
@@ -0,0 +1,129 @@
...
+* **HMAC-SHA256:** Includes a MicroPython-compatible HMAC-SHA256 implementation for signing.
+* **Time Synchronization:** Includes a helper function to synchronize the device's time using NTP, which is crucial for MinIO request signing.
+* **Minimal Dependencies:** Built with standard MicroPython libraries like `urequests`, `uhashlib`, `ubinascii`, `utime`, and `network`.
+
+## Requirements
+
+* MicroPython firmware flashed on your device.
+* Network connectivity (WiFi) configured on the device.
+* The following MicroPython libraries:
+    * `urequests`
```
Josverl, line 20: The u prefix for most modules has been deprecated, 

But more importantly, they should be part of a manifest.py, rather than just documented.

python-ecosys/uminio/README.md
```diff
@@ -0,0 +1,129 @@
...
+    * `uhashlib`
+    * `ubinascii`
+    * `utime`
+    * `network`
+    * `ntptime` (for time synchronization)
+
+## Setup
+
+1.  **Copy `__init__.py`:** Create a `uminio` folder in the `/lib` directory of your MicroPython device and copy the `__init__.py` file into it.
```
Josverl, line 29: Installation for a package should be using mop, not copy

python-ecosys/uminio/manifest.py
```diff
@@ -0,0 +1,3 @@
+metadata(version="0.0.1", description="MIcroPython client for MinIO object storage.")
+
+package("uminio")
```
Josverl, line 3: What about the preqeq packages mentioned in the readme file?
The should be included in the manifest, so they are installed automatically

Josverl (changes requested):

PR #1043, opened 2025-09-19
inspect: Implement a very basic signature function.

dpgeorge:
This implements a very basic `inspect.signature()` function.

At the moment it returns only a simple `Signature` instance with a `parameters` attribute that holds an `OrderedDict` whose length matches the arity of the input function (the number of arguments it takes).

So, the following code works and is compatible with CPython:
```py
def f(a, b, *, c):
    pass

print(len(inspect.signature(f).parameters))
```

That should print `3`.

I tested this on unix 64-bit, unix nanbox (32-bit), webassembly (32-bit) and a 32-bit bare-metal target.

-------

The main aim here is to support PyScript which is currently using a very unreliable hack to work out if a Python function takes arguments or not: https://github.com/pyscript/pyscript/blob/main/core/src/stdlib/pyscript/events.py#L96-L126

### DIFF
--- a/python-stdlib/inspect/inspect.py
+++ b/python-stdlib/inspect/inspect.py
@@ -80,3 +80,60 @@ def currentframe():
 
 def getframeinfo(frame, context=1):
     return ("<unknown>", -1, "<unknown>", [""], 0)
+
+
+class Signature:
+    pass
+
+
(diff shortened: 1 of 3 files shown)


dpgeorge:
@WebReflection @ntoll this will allow you to use `inspect.signature()` in your "when" event callback code.

Issue #780, opened 2023-12-21
popen leaves zombie processes


protectivedad:
I discuss this [here](https://github.com/orgs/micropython/discussions/13239#top). The zombie processes make it hard to use if it is called continuously. I reverted to piping to file and reading the output.

PR #1046, opened 2025-09-25
unix-ffi: Fixing behavior to avoid zombie threads.

klukonin:
This commit is fixing  this issue:
https://github.com/micropython/micropython-lib/issues/780
As discussed here https://github.com/orgs/micropython/discussions/13239

Now the pid of child thread will be stored and closed properly. Even with context manager.

### DIFF
--- a/unix-ffi/os/os/__init__.py
+++ b/unix-ffi/os/os/__init__.py
@@ -45,6 +45,7 @@ if libc:
     write_ = libc.func("i", "write", "iPi")
     close_ = libc.func("i", "close", "i")
     dup_ = libc.func("i", "dup", "i")
+    dup2_ = libc.func("i", "dup2", "ii")
     access_ = libc.func("i", "access", "si")
     fork_ = libc.func("i", "fork", "")
     pipe_ = libc.func("i", "pipe", "p")
@@ -207,6 +208,10 @@ def dup(fd):
     check_error(r)
     return r
 
+def dup2(oldfd, newfd):
+    r = dup2_(oldfd, newfd)
+    check_error(r)
+    return r
 
 def access(path, mode):
     return access_(path, mode) == 0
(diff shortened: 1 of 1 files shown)


hodan121:
bump up

unix-ffi/os/os/__init__.py
```diff
@@ -293,24 +298,53 @@ def urandom(n):
     with builtins.open("/dev/urandom", "rb") as f:
         return f.read(n)
 
+class _PopenStream:
+    def __init__(self, fd, pid, mode):
+        self._fd = open(f"/dev/fd/{fd}", mode)
```
dpgeorge: This needs to use `builtins.open` so that the returned object is a stream.

unix-ffi/os/os/__init__.py
```diff
@@ -293,24 +298,53 @@ def urandom(n):
...
-        i, o = o, i
+    def close(self):
+        if not self._closed:
+            try:
+                self._fd.close()
+            finally:
+                pid, status = waitpid(self._pid, 0)
+                self._exitcode = status
+            self._closed = True
+        return self._exitcode
```
dpgeorge: This `_PopenStream` object will need to implement read/write/etc operations, so that you can communicate with the process.

Eg prior to this patch it's possible to do:
```py
import os
ls = os.popen("/bin/ls")
print(ls.read())
ls.close()
```
and that still needs to work.

dpgeorge:
@klukonin do you have any time to update this PR based on the feedback?

klukonin:
Yes. Sorry for the delay.

klukonin:
@dpgeorge 
Seems OK for me =)

unix-ffi/os/os/__init__.py
```diff
@@ -294,23 +301,84 @@ def urandom(n):
...
 
+class _PopenStream:
+    # Wraps the pipe to a process started by popen().  Stream operations are
+    # forwarded to the underlying file, and closing the stream also waits for
+    # the process to finish, so that it does not stay around as a zombie.
+
+    def __init__(self, f, pid):
+        self._f = f
+        self._pid = pid
+        self._status = None
```
agatti, line 312: If you want, you can rename these two variables into something shorter.

Non-local variable names (as in this case) are stored verbatim in the compiled MPY string table.  Since these two variables are clearly meant to not be accessed from outside, picking a short name is usually preferable.

To see that in action for yourself, if you built `mpy-cross` (`make -C mpy-cross` from MicroPython's root directory):

```bash
$ cd $MICROPYTHON_ROOT
$ ./mpy-cross/build/mpy-cross -o compiled.mpy source.py
$ ./tools/mpy-tool.py -d compiled.mpy
```

That will show you the string table for that particular MPY file (along with a bytecode dump, but that's not important in this case).

klukonin: I Agree)
Fixed)

PR #1047, opened 2025-09-30
[lora-sx126x,lora-sx127x]: Add dependencies for sync/async modems.

brenordv:
## Problem
When installing `lora-sx126x` or `lora-sx127x` packages via mip, users will generally install only one of them (either SyncModem or AsyncModem wrapper classes), and this can lead to runtime errors.

When handling import in `lora/__init__.py`, an `ImportError` would can be incorrectly raised when the error message contained `lib.lora` instead of just `lora`.

I noticed this after installing the `lora-sx1262` package via `Thonny` package manager.

## Solution
To fix the import problem, I added an extra check to prevent this from happening by checking if the error message contains both "no module named", and "lora". Since the check was the same for all imports, I centralized the check in a function.

## Testing
Install the package and verify the modem classes are available:
```python
import mip
mip.install("lora-sx126x")
from lora import SX1262, AsyncSX1262  # Should work now
```

### DIFF
--- a/micropython/lora/lora/lora/__init__.py
+++ b/micropython/lora/lora/lora/__init__.py
@@ -5,6 +5,12 @@ from .modem import RxPacket  # noqa: F401
 
 ok = False  # Flag if at least one modem driver package is installed
 
+
+def _can_ignore_error(e):
+    """Check if ImportError can be ignored due to missing module."""
+    return all(x in str(e) for x in ["no module named", "lora"])
+
+
 # Various lora "sub-packages"
 
 try:
@@ -12,7 +18,7 @@ try:
 
     ok = True
 except ImportError as e:
-    if "no module named 'lora." not in str(e):
+    if not _can_ignore_error(e):
         raise
 
 try:
@@ -20,7 +26,7 @@ try:
 
     ok = True
 except ImportError as e:
-    if "no module named 'lora." not in str(e):
+    if not _can_ignore_error(e):
         raise
 
 try:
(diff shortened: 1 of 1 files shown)


projectgus:
Hi @brenordv,

Thanks for digging into this and submitting this PR. Not depending on either `lora-sync` or `lora-async` is an intentional design choice to save flash size - both these packages are biggish and no one is likely to need both.

We tried to smooth over the usability of this in the [installation instructions](https://github.com/micropython/micropython-lib/tree/master/micropython/lora#installation) and by producing a helpful error message if neither package was installed. It seems like your PR also fixes a case where that message wasn't being generated, thanks!

micropython/lora/lora/lora/__init__.py
```diff
@@ -38,3 +42,6 @@
     )
 
 del ok
+
+
+__version__ = '0.2.1'
```
projectgus, line 47: This line is automatically added by `mip` when the package is installed.

brenordv: @projectgus, sorry about that. I removed that line.

projectgus:
@brenordv Thanks for updating the PR.

Maybe I wasn't clear enough in my review: I'm happy to improve the error reporting if both lora-sync and lora-async aren't installed (thanks for that fix).

However, adding both packages as dependencies by default will increase the code size impact of using lora significantly (and almost no one will use both the sync and async APIs on the same board). I'm not inclined to make this change, unless can be convinced it's too difficult to use otherwise.

projectgus:
(Tossed up whether to bump the bugfix version of the package on this, but as it's only a problem for imports it didn't seem worth updating existing installs.)

PR #1052, opened 2025-10-29
sdcard: Updating sector calculation for SDXC.

benwynn:
The current code only calculates sectors for SDHC, with a max of 32G
The new code will calculate sectors for SDXC, with a max of 2T

C_SIZE is 22 bits[69:48] and the high 2 bits of csd[7] are required
to be 0, currently, so i'm not masking them off.

The max value for C_SIZE is 0x3FFEFF, so C_SIZE+1*1024 is a 32-bit number.


projectgus:
Thanks!

PR #1057, opened 2025-11-11
fix: correctly handle WebSocket message fragmentation

ThomasFarstrike:
Handle WebSocket fragmentation was properly by taking into account the "fin" flag to know if a frame is "final" or whether there will be continuations before it's final.


dpgeorge (approved):
This looks like a good fix, thanks.

Issue #952, opened 2024-12-27
minor typo in deep copy


LokiMetaSmith:
I came across this when compiling against copy.py, looking to use the deepcopy function

File "/lib/copy.py", line 194, in deepcopy
NameError: name 'dispatch_table' is not defined

I believe it should be _deepcopy_dispatch

https://github.com/micropython/micropython-lib/blame/e4cf09527bce7569f5db742cf6ae9db68d50c6a9/python-stdlib/copy/copy.py#L96

Issue #699, opened 2023-07-07
aioble examples use incorrect module constant name (module attribute)


gnbl:
Examples in https://github.com/micropython/micropython-lib/blob/master/micropython/bluetooth/aioble/README.md use `aioble.PUBLIC` which, according to https://github.com/micropython/micropython-lib/blob/master/micropython/bluetooth/aioble/aioble/__init__.py should probably be `aioble.ADDR_PUBLIC`.

PR #1058, opened 2025-11-13
Fix constant names and typos in aioble examples and deepcopy function

### LINKED ISSUE #1055: Example should say "..._US".
Instead of "_ADV_INTERVAL_MS" here:

https://github.com/micropython/micropython-lib/blob/3eaf0279f350855f0ee3a8a65774a49afdbf4e3d/micropython/bluetooth/aioble/examples/temp_sensor.py#L23

Josverl:
This PR combines a few rather minor fixes of typos in documentation and scripts.

- Update the aioble example to use the correct advertising interval constant name. 
- Correct the README to reference the appropriate public address constant. 
- Fix a typo in the deepcopy function to reference the correct dispatch table.

Fixes #1055, #699, #952


python-stdlib/cmd/cmd.py
```diff
@@ -23,32 +23,33 @@
...
-`self.undoc_header' set the headers used for the help function's
+The data members ``self.doc_header``, ``self.misc_header``, and
+``self.undoc_header`` set the headers used for the help function's
 listings of documented functions, miscellaneous topics, and undocumented
 functions respectively.
 
-----------------------------------------------------------------------------
-This is a copy of python's Cmd, but leaves out features that aren't relevant
-or can't currently be implemented for MicroPython.
+.. caution::
```
dpgeorge: I really don't think we should be using RST formatting here.

python-stdlib/heapq/heapq.py
```diff
@@ -1,19 +1,20 @@
...
+Usage::
 
-heap = []            # creates an empty heap
-heappush(heap, item) # pushes a new item on the heap
-item = heappop(heap) # pops the smallest item from the heap
-item = heap[0]       # smallest item on the heap without popping it
-heapify(x)           # transforms list into a heap, in-place, in linear time
-item = heapreplace(heap, item) # pops and returns smallest item, and adds
-                               # new item; the heap size is unchanged
+  heap = []            # creates an empty heap
```
dpgeorge: Why not 4 space indent like the other example code reformatted in this PR?

dpgeorge (commented):
Thanks for this, there are some good fixes here.

I approve all commits except the RST formatting one.

Josverl:
The rest formatting was a side effect of the work I started with @jimmo to see if/how it would be possible to generate documentation pages from the source files. 

With these changes that works, although de demo i built for that is no longer live.

It ( sphinx +autoapi) does require consistent formatting, and sphinx and the rest of the project use Rst. 

It could be simpler markdowns or plain docstrings as well, but not RST with formatting errors.
And there needs to be some degree of consistency, documented in contributed guidelines for the -lib repo

This was discussed in 

https://github.com/orgs/micropython/discussions/15463

micropython/drivers/imu/bmi270/manifest.py
```diff
@@ -1,2 +1,2 @@
-metadata(description="BOSCH BMI270 IMU driver.", version="1.0.0")
+metadata(description="BOSCH BMI270 IMU driver.", version="1.0.1")
```
dpgeorge: For code that just has a change to the comments, I don't think we should update the version number.  It uses up a small amount of storage to store each version, and the compiled .mpy files anyway are not changed.

Josverl: reverted.

dpgeorge (approved):

PR #1062, opened 2025-11-15
sdcard: Adding tests for csd parsing.

benwynn:
On my last PR that updated the parsing of the CSD register,
I was asked if a test could be provided.

Here's a test for the csd parsing, with csd values
recorded from the two cards I currently have available.

I've ordered some smaller cards so I can record and test
additional values, but happy to have others contribute
test-cases as well.


projectgus (commented):
Hi @benwynn, sorry it's taken a while for someone to get back to you with this change - appreciate you putting in the time to improve test coverage.

I have some quite small SD cards kicking around somewhere, so I'll try to get some additional CSD values - although I think merging with the two you already have would be a benefit by itself.

dpgeorge:
Thanks @benwynn for your efforts here.  In "normal" Python this kind of change might be OK.  But for MicroPython we prioritize efficiency over almost everything else, and adding new functions makes it less efficient (and take up more RAM/ROM).

A better way to make a test would be to fully mock out an SPI object and pass that into the `SDCard` constructor.  Then that mocked SPI object would return the test CSD data.  That's what @projectgus was saying in his comment https://github.com/micropython/micropython-lib/pull/1062#discussion_r2710958137

benwynn:
ok, thank you for explaining gus' comment.  i'll retract this; but mocking out the SPI object seems like a lot more work

projectgus:
Thanks @benwynn, sorry for the unclear review comment earlier.

PR #1064, opened 2025-11-18
argparse: Add support for custom argument types.

agatti:
This PR adds support for optional custom argument type validation to argparse.ArgumentParser, allowing for shorter argument validation code for both simple builtins and complex types, as supported by CPython.

CPython also provides the `argparse.FileType` class to simplify argument file/directory handling, but that was not added to the MicroPython module.  Its behaviour is rather easy to manually replicate if needed anyway.

This change increases the size of the `argparse` module by 291 bytes, and that's mostly due to the extra error handling and related text strings.  Probably those new strings can be shortened, but given that this module isn't usually included in production code, the size increase shouldn't hurt too much I guess.

The unit tests unfortunately do not cover the case of a parameter failing validation, as the module will automatically issue a `sys.exit` call on error, after printing the relevant exception message.  If there are cleaner ways to test that without hijacking `sys.exit`, let me know so I can update the tests accordingly.

### DIFF
--- a/python-stdlib/argparse/argparse.py
+++ b/python-stdlib/argparse/argparse.py
@@ -10,8 +10,12 @@ class _ArgError(BaseException):
     pass
 
 
+class ArgumentTypeError(BaseException):
+    pass
+
+
 class _Arg:
-    def __init__(self, names, dest, action, nargs, const, default, help):
+    def __init__(self, names, dest, action, nargs, const, default, help, type):
         self.names = names
         self.dest = dest
         self.action = action
@@ -19,20 +23,26 @@ class _Arg:
         self.const = const
         self.default = default
         self.help = help
+        self.type = type
+
+    def _apply(self, optname, arg):
+        if self.type:
+            try:
+                return self.type(arg)
+            except (ArgumentTypeError, TypeError, ValueError) as e:
+                raise _ArgError("invalid value for %s: %s (%s)" % (optname, arg, str(e)))
+        return arg
 
     def parse(self, optname, args):
         # parse args for this arg
         if self.action == "store":
             if self.nargs is None:
                 if args:
-                    return args.pop(0)
+                    return self._apply(optname, args.pop(0))
                 else:
                     raise _ArgError("expecting value for %s" % optname)
             elif self.nargs == "?":
-                if args:
-                    return args.pop(0)
-                else:
-                    return self.default
+                return self._apply(optname, args.pop(0) if args else self.default)
             else:
                 if self.nargs == "*":
                     n = -1
@@ -52,7 +62,7 @@ class _Arg:
                         else:
                             break
                     else:
-                        ret.append(args.pop(0))
+                        ret.append(self._apply(optname, args.pop(0)))
                         n -= 1
                 if n > 0:
                     raise _ArgError("expecting value for %s" % optname)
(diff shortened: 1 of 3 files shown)


python-stdlib/argparse/test_argparse.py
```diff
@@ -66,3 +66,37 @@
...
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-a", type=int)
+parser.add_argument("-b", type=str)
+parser.add_argument("-c", type=CustomArgType(1))
+try:
+    parser.add_argument("-d", type=())
+    assert False
+except ValueError as e:
+    assert str(e) == "type is not callable"
```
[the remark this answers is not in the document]
agatti: Thanks, this should be sorted out now.

python-stdlib/argparse/argparse.py
```diff
@@ -103,6 +115,12 @@ def add_argument(self, *args, **kwargs):
                 dest = args[0]
             if not args:
                 args = [dest]
+        arg_type = kwargs.get("type", None)
+        if arg_type is not None:
+            if not callable(arg_type):
+                raise ValueError("type is not callable")
+            if default is not None and not isinstance(default, str):
```
stinos: This is not correct: it skips conversion when an argument is passed. Test like

```
parser.add_argument("-x", type=CustomArgType(1), default=1)
parser.add_argument("-g", type=CustomArgType(1), default=1)
parser.add_argument("-i", type=CustomArgType(1), default="1")
args = parser.parse_args(["-x", "3"])
print(args)
assert args.x == 4
assert args.g == 1
assert args.i == 2
```

agatti: That was it, missed that bit.  Thanks for sticking with this, it was indeed an issue on my end, code-wise.

I'd partially chalk it up to CPython's documentation being not really clear for that.  It should probably read (emphasis mine) "If the type keyword is used with the default keyword, the type converter is only applied *to the default value* if the default is a string.".

Still, the test file's behaviour in isolation under CPython now matches MicroPython's when using its own `argparse` implementation.  Hope this should be it :)

python-stdlib/argparse/argparse.py
```diff
@@ -10,29 +10,41 @@ class _ArgError(BaseException):
...
+        self.type = type
+
+    def _apply(self, optname, arg):
+        if self.type:
+            try:
+                return self.type(arg)
+            except Exception as e:
+                if isinstance(e, (ArgumentTypeError, TypeError, ValueError)):
+                    raise _ArgError("invalid value for %s: %s (%s)" % (optname, arg, str(e)))
+                raise
```
dpgeorge: How about writing this as more idiomatic Python:
```py
except (ArgumentTypeError, TypeError, ValueError) as e:
    raise _ArgError(...)
```
I think that's equivalent?  It also saves 12 bytes in the resulting .mpy.

agatti: I thought this syntax wasn't supported by MicroPython, good to know I was wrong all along :)

It is indeed equivalent.  Thanks for pointing this out, I've changed the code to reflect that.

dpgeorge (approved):

python-stdlib/argparse/manifest.py
```diff
@@ -1,4 +1,4 @@
-metadata(version="0.4.0")
+metadata(version="0.5.0")
```
dpgeorge: I bumped the minor version, because this change is adding a feature.

PR #1077, opened 2026-01-26
fix(logging): allow child loggers to set independent log levels

ThomasFarstrike:
The handler's level was being set to the root logger's level in basicConfig(), which prevented child loggers from setting their own levels. This violated the standard logging architecture where handlers should not filter by level - that's the logger's responsibility.

Changed handler.setLevel(level) to handler.setLevel(NOTSET) in basicConfig() so that handlers pass through all messages and let loggers control filtering.

This allows code like:
  logger = logging.getLogger("myapp")
  logger.setLevel(logging.INFO)
  logger.info("message")  # Now works correctly

Previously, INFO messages would be filtered out if the root logger was at WARNING.

Fixes: Child loggers unable to log at levels lower than root logger's level

### DIFF
--- a/python-stdlib/logging/logging.py
+++ b/python-stdlib/logging/logging.py
@@ -242,7 +242,7 @@ def basicConfig(
         else:
             handler = FileHandler(filename, filemode, encoding)
 
-        handler.setLevel(level)
+        handler.setLevel(NOTSET)
         handler.setFormatter(Formatter(format, datefmt))
 
         logger.setLevel(level)


agatti:
Thanks for the PR.

Are there any chances the commit message will get updated to prevent CI from complaining? :)

ThomasFarstrike:
> Are there any chances the commit message will get updated to prevent CI from complaining? :)

I don't see the error anymore and also don't see how to change it easily as this was just done in github, not on a clone...

If you don't mind, I would just leave the commit message as it is... it's not that bad :-D

agatti:
Indeed the error report has been purged due to its age, however I can quickly notice two issues with this.  The subject line format doesn't follow the repo guidelines (https://github.com/micropython/micropython-lib/blob/master/CONTRIBUTING.md#pull-requests - first bullet point), and the commit lines are too long (should be max 75 characters) [1].

Committing through GitHub's editor unfortunately does bypass all the pre-commit checks we have set up, so that should have been caught at commit time when adding this to your repo first [2].

Anyway, something like this goes through the checks (I've simply reformatted the subject line and wrapped the commit message body at column 74):

```
logging: Allow child loggers to set independent log levels.

The handler's level was being set to the root logger's level in
basicConfig(), which prevented child loggers from setting their own
levels. This violated the standard logging architecture where handlers
should not filter by level - that's the logger's responsibility.

Changed handler.setLevel(level) to handler.setLevel(NOTSET) in
basicConfig() so that handlers pass through all messages and let loggers
control filtering.

This allows code like:
  logger = logging.getLogger("myapp")
  logger.setLevel(logging.INFO)
  logger.info("message")  # Now works correctly

Previously, INFO messages would be filtered out if the root logger was at
WARNING.

Fixes: Child loggers unable to log at levels lower than root logger's
level

Signed-off-by: Thomas Farstrike <111072251+ThomasFarstrike@users.noreply.github.com>
```

[1] This isn't mentioned in the guidelines, as it is assumed all code goes through pre-commit hooks first.  I'll send a PR to update this.

[2] Guess that wasn't considered as an option back when the guidelines were drafted, time to update that too!

python-stdlib/logging/logging.py
```diff
@@ -242,7 +242,7 @@ def basicConfig(
         else:
             handler = FileHandler(filename, filemode, encoding)
 
-        handler.setLevel(level)
+        handler.setLevel(NOTSET)
```
dpgeorge, line 245: The default level is NOTSET (see `Handler` constructor), so I think a simpler and equivalent fix would be to just remove this line.

Issue #1083, opened 2026-02-18
unix-ffi re module throws when looking at match containing optional groups


dhouck:
## Problem description
When a regex contains optional capture groups, for example `(a)?b`, the `PCREMatch.group()` method throws an overflow error for that group.  I believe this is because PCRE is representing a nonexistent group as `SIZE_MAX` and `re` isnʼt checking for that.

## Additional fallout
The unix-ffi `json` library requires the unix-ffi `re` library, and currently cannot parse numbers unless they have an integer part, a fractional part, *and* and exponential part; instead, it throws this same error.

## To reproduce
```py
>>> import re
>>> r = re.compile(r'(a)?b')
>>> m = r.match('b')
>>> (m.group(0), m.group(1))
```

### Expected (cpython, and micropython built-in `re` library):
```py
('b', None)
```

### Actual (micropython `re` library)
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/micropython/re.py", line 65, in group
OverflowError: overflow converting long int to machine word
```

dhouck:
I believe this is fixed in d8d362e but I have not yet tested.

dhouck:
Yep this is fixed in v1.28.0

PR #1084, opened 2026-02-18
unix-ffi/json: Fix json.loads() of a bytes object.

dhouck:
The conversion from bytes to string previously worked only if some other argument was also given to `loads`.  This makes even a basic call with no other arguments work.

NOTE: The `test_json.py` file fails, and failed before this PR, because of #1083.  If I patch `scanner.py` to only parse ints (and thus not need optional groups), the tests pass.

### DIFF
--- a/unix-ffi/json/json/__init__.py
+++ b/unix-ffi/json/json/__init__.py
@@ -391,6 +391,8 @@ def loads(
     The ``encoding`` argument is ignored and deprecated.
 
     """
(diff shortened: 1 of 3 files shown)


dhouck:
Now that #1083 is fixed, the updated `test_json.py` now indeed works.

dpgeorge (approved):
Thanks for the fix. 

For reference, the original code to support bytes/bytearray was added in d1a74360a20dadfaae717d4d8c8bd3531672362f

PR #1086, opened 2026-02-19
umqtt.async: Initial implementation based on umqtt.simple

dketov:
The async mqtt.simple adaptation solely based on umqtt.simple code


dketov:
@pfalcon  please have a look

dpgeorge:
This reimplements all the MQTT logic from scratch.  Is there a way to factor out the logic (eg for creating packets) so it can be shared by `umqtt.simple` and `umqtt.async`?

And see also #722 which does a very similar thing.

dketov:
> This reimplements all the MQTT logic from scratch. Is there a way to factor out the logic (eg for creating packets) so it can be shared by `umqtt.simple` and `umqtt.async`?
Of course it is :) I'll have a look later

> And see also #722 which does a very similar thing.
Sure

PR #1092, opened 2026-03-06
python-ecosys/dotenv: Add .env file loader package.

Holdrulff:
This package provides a lightweight .env file loader for MicroPython, allowing secure storage of credentials and configuration outside of source code.

Features:
- Zero dependencies
- ~1.9KB size
- Works on ESP32, ESP8266, RP2040
- Tested on ESP32-C3
- Simple 2-function API: load_dotenv() and get_env()

### DIFF
--- a/python-ecosys/dotenv/README.md
+++ b/python-ecosys/dotenv/README.md
@@ -1,0 +1,20 @@
+# dotenv
+
+Lightweight .env file loader for MicroPython.
+
+## Installation
+
+```python
+import mip
+mip.install("dotenv")\
+```
+
+## Usage
+
+```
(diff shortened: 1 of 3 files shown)


dpgeorge:
Thanks for the contribution!

Usually a package in `python-ecosys` would be a subset of the corresponding "big" Python package.  But I see that `dotenv.get_env()` doesn't exist in big dotenv.  Instead it's `dotenv.get_key(file, key)`.

What do you think about making the API here a subset of the standard dotenv?

dpgeorge:
Closing due to inactivity.  Feel free to reopen with a response to the above.

PR #1094, opened 2026-03-10
aiohttp: Reimplement ChunkedClientResponse to correctly handle chunked responses

### LINKED ISSUE #1093: Chunked responses in aiohttp result in an attempt to allocate a 4 gigabyte buffer… and maybe it doesn't actually work anyway?
Commit b95ca24 replaced calls to `asyncio.StreamReader.read` with calls to `asyncio.StreamReader.readexactly`. One important distinction between these two methods is that `read` allows a value of -1 to be passed, which results in it reading until EOF is reached, whereas `readexactly` requires (as its name implies) an exact size. This change breaks "chunked" transfers, as they don't supply a "Content-Length" header. When that header isn't present, `aiohttp` defaults to passing -1 to the read method, resulting in it attempting to allocate a buffer of 4,294,967,126 bytes.

Interestingly, `ChunkedClientResponse.read` correctly decodes the content length from the first line of the response data (which is consistent with the HTTP 1.1 spec), but ignores that value in favor of the -1 passed from `ClientResponse.text` or `ClientResponse.json`. This is handled in `ClientResponse.read` by checking specifically for a value of -1 and using the `asyncio.StreamReader.read` method, but not in `ChunkedClientResponse.read`.

## Upon further inspection…

After looking closer at `ChunkedClientResponse` and the HTTP 1.1 spec, I think it just doesn't work correctly at all. A chunk-encoded response can require multiple reads from the response stream, each of a length determined by a value encoded in its first line. The way the class is written assumes that the first chunk's size is the size of the full data (chunked encoding is specifically intended for situations where the full size isn't known when the transfer starts) and there is only a single chunk.

So hey, I'm going to rewrite it and do a PR.

rspeed:
A chunk-encoded response can (and often will) require multiple reads from the response stream, each of a length determined by a value encoded in its first line. `aiohttp.ChunkedClientResponse` currently treats the first chunk's size as the size of the full data (chunked encoding is specifically intended for situations where the full size isn't known when the transfer starts) and there is only a single chunk.

This PR reimplements `aiohttp.CHunkedClientResponse` based on the HTTP/1.1 specification. Chunked responses with more than one chunk will now work correctly. As a side-effect it also fixes #1093.

### DIFF
--- a/python-ecosys/aiohttp/aiohttp/__init__.py
+++ b/python-ecosys/aiohttp/aiohttp/__init__.py
@@ -57,25 +57,23 @@ class ClientResponse:
 
 
 class ChunkedClientResponse(ClientResponse):
-    def __init__(self, reader):
-        self.content = reader
-        self.chunk_size = 0
+    async def read(self, sz=-1):
+        data = bytearray(b"")  # response data accumulator
 
-    async def read(self, sz=4 * 1024 * 1024):
-        if self.chunk_size == 0:
+        while True:
             l = await self.content.readline()
-            l = l.split(b";", 1)[0]
-            self.chunk_size = int(l, 16)
-            if self.chunk_size == 0:
-                # End of message
-                sep = await self.content.readexactly(2)
-                assert sep == b"\r\n"
-                return b""
-        data = await self.content.readexactly(min(sz, self.chunk_size))
-        self.chunk_size -= len(data)
-        if self.chunk_size == 0:
-            sep = await self.content.readexactly(2)
-            assert sep == b"\r\n"
+            l = l.strip(b"\r\n")  # remove CRLF
+            l = l.split(b";", 1)[0]  # ignore chunk-extensions
+            chunk_size = int(l, 16)
+            if chunk_size == 0:
+                break
+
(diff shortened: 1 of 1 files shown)


dpgeorge:
> `aiohttp.ChunkedClientResponse` currently treats the first chunk's size as the size of the full data (chunked encoding is specifically intended for situations where the full size isn't known when the transfer starts) and there is only a single chunk.

I don't think that's correct.  The current code does support multiple chunks, using the `self.chunk_size` variable to keep track of how much is remaining in the current chunk.  When that becomes 0 then it reads in the next chunk header and start the process again.

This allows you to read in a chunked response in separate chunks, where the chunk sizes are different.  Eg the incoming data could be chunked every 256 bytes, but the application reads in 16 byte blocks:
```py
resp = session.get(...)
while True:
    data = await resp.read(16)
    if not data:
        # end of stream
        break
    process_data(data)
```

IMO that's a useful extension to standard aiohttp because most microcontroller applications can't read in megabytes at once.

I think the only issue to solve here is handling `-1` to indicate that you want to read everything.

Issue #1090, opened 2026-03-02
Deepcopy does not work , copied Copy from here.  Can not copy Instances - Error: un(deep)copyable object of type <class 'counterObject'>. Simple examples  do NOT work.


jlsilicon:
Deepcopy does not work.
Using Thonny Micropython on PICO.
Copy.py is NOT on the UF2 file for PICO - as much as the docs seem to reflect.
Copied Copy.py from site here.  

Can not copy Instances : 
Error: un(deep)copyable object of type <class 'counterObject'>

Simple examples  do NOT work.

If Standard documented Features / Cmds don't work here , then what good is the MicroPython ???

I have invested in using this for Programs for robots that I build.
So far the OOPS Objects Structuring here - is pretty much Trash - without the ability to copy Objects.
- JAVA has no problem with this, what's the Problem here ???

---

StackOverflow ex :
https://stackoverflow.com/questions/42143461/make-copy-of-object-instance-in-python

```
class counterObject:

def __init__(self):
self.Value = 0

def incrementValue(self):
self.Value += 1

def printValue(self):
print(self.Value)

A = counterObject()
A.incrementValue()
A.printValue() ##Prints 1

import copy

B = copy.deepcopy(A)

## => Error: un(deep)copyable object of type <class 'counterObject'>

print( id(A), id(B) )

```

---

Even the Google example does not work :
https://www.google.com/search?q=micropython+howto+create+object+then+copy+object

This returns :
 => Error: un(deep)copyable object of type <class 'MyClass'>

How to Create an ObjectYou define a class and then create an instance (object) of that class by calling the class name as a function.python

```
class MyClass:
def __init__(self, name, value):
[self.name](http://self.name/) = name
self.value = value

## Create an object (instance of the class)
original_object = MyClass("example", 100)
```

2. Deep Copy (copy.deepcopy())A deep copy creates a completely independent clone of the original object, including recursively copying all nested objects and data structures. Changes to the copy will not affect the original. 

```
import copy

## Create a deep copy
deep_copied_object = copy.deepcopy(original_object)

## => Error: un(deep)copyable object of type <class 'MyClass'>

```

---

agatti:
More documentation about this issue is available in 28fe4b70da64d70ec45cd2ccf8bf82a2c78f977d.

Closing this as completed, feel free to reopen if you think this needs more clarifications.

PR #1098, opened 2026-03-22
copy: Provide top-level documentation.

agatti:
### Summary

This PR adds documentation for the `copy` module, explaining why the module may require some changes in the objects being copied to behave as CPython.

Unlike CPython, MicroPython does not provide default implementations for either `__reduce__` or `__reduce_ex__`, which are used by the copy module as the fallback methods to create copies of non-trivial objects.  These methods aren't provided by default to minimise both the footprint of the interpreter binary and the RAM taken by object instances.

Users who want simpler ways to allow object copies can implement `__copy__` and `__deepcopy__` instead, as mentioned in CPython's documentation for the `copy` module.

Arguably the lack of `__reduce__` and `__reduce_ex__` could also be mentioned in MicroPython's own documentation as part of the differences from CPython, but it doesn't hurt to mention more module-specific caveats in here too.  This should, in theory, keep user expectations in check w.r.t. the behaviour of this specific module.

This should clarify scenarios like the ones mentioned in #1090.

### Generative AI

### DIFF
--- a/python-stdlib/copy/README.md
+++ b/python-stdlib/copy/README.md
@@ -1,0 +1,23 @@
+# copy
+
+Non-trivial objects (eg. classes) instantiated by CPython are usually able to
+be copied by methods of this module without any modification, as the copying
+mechanism makes use of two functions (`__reduce__` and `__reduce_ex__`) that
+are added to instances by CPython itself.
+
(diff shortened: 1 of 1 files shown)


Josverl (changes requested):

dpgeorge (approved):
Thanks!

Issue #400, opened 2020-10-22
OSError: -40


mradovan:
ESP-8266, urequests installed manually.

Code: 

import urequests as u
req = u.get("https://api.telegram.org/bot<my_token>/ ... my_chat_id>")
, i get this:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/lib/urequests.py", line 112, in get
  File "/lib/urequests.py", line 60, in request
OSError: -40

But for example, this works:
req = u.get("https://telegram-webhook-sample.herokuapp.com/")

JasonLearmouth:
I'm getting the same error when I try posting to Slack.

Traceback (most recent call last):
  File "main.py", line 16, in <module>
  File "slackService.py", line 10, in send_toban_status
  File "urequests.py", line 166, in post
  File "urequests.py", line 104, in request
OSError: -40

MicroPython v1.13 on 2020-09-11; ESP module with ESP8266

```python
URL = 'https://hooks.slack.com/services/TTL3HSK...4pbNKt9'
msg = dict([("text","Hello, World!")])
resp = urequests.post(URL, data=msg)
```

I have a get request that is working fine, but it was not HTTPS. When I changed it to HTTPS (which is supported by the server), I now get an error on that request:

TLS buffer overflow, record size: 9040 (+5)
Traceback (most recent call last):
  File "main.py", line 12, in <module>
  File "rfsService.py", line 10, in get_sydney_toban_today
  File "urequests.py", line 162, in get
  File "urequests.py", line 128, in request
OSError: -257

Not the same error in each case, so maybe not that helpful...

I also tried the URL that @mradovan said worked, and it worked for me as well.

gretel:
have the same case using the `telegram` api. as a side note - it's working fine using the `pycom` fork.

mrkprdo:
I encountered same error with urequests when trying to fetch HTTPS url.
As an experiment, i tried it with HTTP (http://www.example.com/) and surprisingly it is working.
Checking urequests.py line 59-60, as expected it is related to ussl module
```
        if proto == "https:":
            s = ussl.wrap_socket(s, server_hostname=host)
```

I found somewhere that socket must be closed first, just dont know how to make it work with this one...

gretel:
@mrkprdo i got the same idea.. and on `wrap_socket()` the `esp8266` resets. i did enable the emergency buffer but don't get anything but garbled output until it resets. i'm going to read the related `c` code soon i hope.

[...]

PR #1101, opened 2026-03-25
micropython/mip/mip: Add an alias for codeberg repos.

agatti:
### Summary

This PR introduces an alias to access codeberg repos from within the mip embedded package manager.

Right now packages hosted on codeberg could only be referenced by their full URL, unlike other packages hosted on either GitHub or GitLab.  To make access those packages easier, now they can be referenced as "codeberg:org/repo".

This is the `micropython-lib` counterpart of https://github.com/micropython/micropython/pull/18989.

### Testing

The same package mentioned in the other repo's PR (https://codeberg.org/agatti/micropython-package-test) was installed successfully via `import mip; mip.install("$URL")` on an ESP32C3.

The ESP8266 is not able to access codeberg repos due to a SSL cipher incompatibility (see #400), but there isn't much that can be done for that right now.

The same remarks about whether adding a new provider or not, mentioned in the micropython PR still apply.

A similar refactoring has been applied to the on-device python code - albeit with a few more aggressive size optimisations, having a final size impact of -6 bytes.  Unfortunately codeberg using a different parameter order than the other hosts forces having some named parameters in the string template so some more space could have been saved there.  A further size reduction could be had by renaming `_HOSTS` into something shorter.  Suggestions for the new name are welcome :)

### Generative AI

### DIFF
--- a/micropython/mip/mip/__init__.py
+++ b/micropython/mip/mip/__init__.py
@@ -10,7 +10,19 @@ from micropython import const
 _PACKAGE_INDEX = const("https://micropython.org/pi/v2")
 _CHUNK_SIZE = const(128)
 
-allowed_mip_url_prefixes = ("http://", "https://", "github:", "gitlab:")
+# Since all URLs are accessed via HTTPS, the URL scheme is added by _rewrite_url.
+# The first three format parameters are assumed to be the organisation, the
+# repository names, and the branch/tag name, in this order.
+_HOSTS = {
+    # https://codeberg.org/api/v1/repos/{org}/{repo}/raw/{path}?ref={branch}
+    "codeberg:": "codeberg.org/api/v1/repos/{}/{}/raw/{p}?ref={}",
+    # https://raw.githubusercontent.com/{org}/{repo}/{branch}/{path}
+    "github:": "raw.githubusercontent.com/{}/{}/{}/{p}",
+    # https://gitlab.com/{org}/{repo}/-/raw/{branch}/{path}
+    "gitlab:": "gitlab.com/{}/{}/-/raw/{}/{p}",
+}
+
+_ALLOWED_MIP_URL_PREFIXES = const(("http://", "https://", "codeberg:", "github:", "gitlab:"))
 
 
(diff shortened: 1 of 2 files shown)


Josverl (approved):
tested with package, package@branch and  package@version,
All work fine

micropython/mip/mip/__init__.py
```diff
@@ -10,7 +10,20 @@
...
+    # https://codeberg.org/api/v1/repos/{org}/{repo}/raw/{path}?ref={branch}
+    "codeberg:": "codeberg.org/api/v1/repos/{}/{}/raw/{p}?ref={}",
+    # https://raw.githubusercontent.com/{org}/{repo}/{branch}/{path}
+    "github:": "raw.githubusercontent.com/{}/{}/{}/{p}",
+    # https://gitlab.com/{org}/{repo}/-/raw/{branch}/{path}
+    "gitlab:": "gitlab.com/{}/{}/-/raw/{}/{p}",
+}
+
+# ruff: noqa: RUF005 - tuple construction with list destructuring is not supported.
+allowed_mip_url_prefixes = tuple(["http://", "https://"] + list(_HOSTS.keys()))
```
dpgeorge: I suggest leaving this as a literal tuple of strings.  That might actually produce smaller bytecode, and it'll definitely be faster to import because it's just a constant.

Actually, this could even be a true const:
```py
_ALLOWED_MIP_URL_PREFIXES = const(("http://", ...))
```
which should reduce code size.

agatti: Making the tuple `const` did save more than enough bytes on its own.  Now the new version is actually 6 bytes *shorter* than the original.

dpgeorge (approved):
Sorry I didn't see that you updated this.  Looks very good now!

micropython/mip/mip/__init__.py
dpgeorge: It's probably shorted because the `allowed_mip_url_prefixes` global variable name no longer exists.

PR #1102, opened 2026-04-02
lora/lora-sx127x-pycom: Add unified SX1272/SX1276 LoRa driver

ddtdanilo:
## Summary

Add `lora-sx127x-pycom`: a unified SX1272/SX1276 LoRa radio driver for `micropython-lib`.

The existing `lora-sx127x` package only supports SX1276 (version register `0x12`); this driver also handles SX1272 (`0x22`), which has a different register layout for bandwidth encoding, coding rate bits, CRC location, and RSSI calculation offsets.

This package is consumed by the Pycom LoPy (SX1272) and LoPy4 (SX1276) board definitions in micropython/micropython#19026.

## Testing

- All files pass `ruff check` and `ruff format`
- Hardware tests (driver init, config, TX power, validation) available in micropython/micropython#19026

## Related

- LoRaWAN MAC layer: micropython/micropython-lib#1111 (split out per @projectgus's request)
- Boards consuming this package: micropython/micropython#19026 (PYCOM_LOPY, PYCOM_LOPY4)

### DIFF
--- a/micropython/lora/lora-lorawan/lora/lorawan.py
+++ b/micropython/lora/lora-lorawan/lora/lorawan.py
@@ -1,0 +1,397 @@
+"""
+LoRaWAN 1.0.x MAC layer for MicroPython.
+
+Implements Class A device operation with OTAA and ABP activation.
+Runs on top of the SX127x raw LoRa driver.
+
+Usage (OTAA):
+    from lora import LoRa
+    from lorawan import LoRaWAN
+
+    radio = LoRa(frequency=868100000, sf=7, bw=125000)
+    wan = LoRaWAN(radio, mode=LoRaWAN.OTAA,
+                  dev_eui=bytes.fromhex('...'),
+                  app_eui=bytes.fromhex('...'),
+                  app_key=bytes.fromhex('...'))
+    wan.join()
+    wan.send(1, b'Hello')
+
+Usage (ABP):
+    wan = LoRaWAN(radio, mode=LoRaWAN.ABP,
+                  dev_addr=0x01234567,
+                  nwk_skey=bytes.fromhex('...'),
+                  app_skey=bytes.fromhex('...'))
+    wan.send(1, b'Hello')
+"""
+
+from micropython import const
+import struct
+import time
+import os
+from lora.lorawan_crypto import (
+    aes_cmac,
+    aes_ctr_encrypt,
+    aes_ctr_decrypt,
+    compute_mic,
+    derive_session_keys,
+)
+
+# LoRaWAN MHDR message types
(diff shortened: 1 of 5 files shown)


projectgus (commented):
This looks really exciting, thanks @ddtdanilo! Had always hoped the LoRa micropython-lib support would be enough for someone to build a LoRaWAN driver.

Do you mind moving the lorawan support into its own PR, please? There's a lot of new code here, and it will be easier to review and merge these independently.

ddtdanilo:
Hi @projectgus — thanks for the review!

I've split the LoRaWAN MAC layer into its own PR as you suggested:

➡️ micropython/micropython-lib#1111 (`lora-lorawan`)

This PR (#1102) now contains only the `lora-sx127x-pycom` driver. The two are independently reviewable and mergeable. Force-pushed `add-pycom-lora` to keep just the driver commit (`c9c9d04`); the MAC commit is on `add-pycom-lorawan` for #1111.

Let me know if you'd like any further changes to either one.

ddtdanilo:
Hi @projectgus — just a friendly check-in on this one. Since the split a week ago, #1102 carries only the `lora-sx127x-pycom` driver (the smaller, focused unit you asked for) and CI has stayed green. Happy to apply any tweaks whenever you have a moment. Companion PR for the MAC layer is #1111. No rush — thanks for your time!

projectgus (commented):
Hi @ddtdanilo,

Thanks for splitting these out and sorry it's taken me a little while to respond.

My main concern with this driver is that it sits independent from the existing lora-sx126x and lora-sx127x drivers, which both share a common base class for their documented public API, and support both synchronous and asynchronous driver classes on top.

Particularly as we already have a SX1276 driver, merging a second slightly different SX1276 driver seems like a mis-step.

So before going into more details, my high-level questions:

* Is this driver API-compatible with the existing SX1276 driver in micropython-lib?
* Did you investigate how hard it would be to add an SX1272 driver with the same architecture as the existing lora-sx127x driver, and/or by extending the existing lora-sx126x driver? If so, what did you find?

(This may go without saying, and I apologise if this seems abrupt, but I'm interested in what you the human contributor thinks, as opposed to what text an LLM outputs when asked these questions. Although you're of course welcome to consult an LLM yourself.)

projectgus:
> Is this driver API-compatible with the existing SX1276 driver in micropython-lib?

Sorry, answering my own question here as I started looking at the LoRaWAN PR which has some sample code - the APIs aren't compatible. I'm afraid that's a non-negotiable, we can't support two different LoRa APIs in micropython-lib (particularly if we end up with two drivers for one chip).

PR #1104, opened 2026-04-08
tools/ci.sh: Allow CI tests to run locally.

tsukasa-au:
When working on code in this repo, it can be beneficial to run the same tests that run as part of the CI workflow on GitHub.

This change makes it possible to run `tools/ci.sh` without it messing with your user-wide micropython install (`~/.micropython/lib`). It now defaults to creating a virtual environment in `/tmp/micropython-venv`.

Expected usage of this is:
```bash
# Setup tests (run once)
$ source tools/ci.sh
$ ci_package_tests_setup_micropython
# Run tests (run many times)
$ ci_package_tests_setup_lib && ci_package_tests_run || echo 'Failed'
```

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -17,7 +17,9 @@ function ci_commit_formatting_run {
 ########################################################################################
 # package tests
 
-MICROPYTHON=/tmp/micropython/ports/unix/build-standard/micropython
+VIRTUAL_ENV="${VIRTUAL_ENV:-/tmp/micropython-venv}"
+MICROPYTHON="${MICROPYTHON:-${VIRTUAL_ENV}/bin/micropython}"
+MICROPYPATH="${VIRTUAL_ENV}/lib:.frozen"
 
 function ci_package_tests_setup_micropython {
     git clone https://github.com/micropython/micropython.git /tmp/micropython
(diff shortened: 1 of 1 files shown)


tools/ci.sh
```diff
@@ -17,7 +17,9 @@ function ci_commit_formatting_run {
 ########################################################################################
 # package tests
 
-MICROPYTHON=/tmp/micropython/ports/unix/build-standard/micropython
+VIRTUAL_ENV="${VIRTUAL_ENV:-/tmp/micropython-venv}"
```
agatti: Hardcoding `/tmp` should probably be left as a last resort - POSIX specs say there should be an environment variable called `TMPDIR` pointing to the temporary directory.

See also https://en.wikipedia.org/wiki/TMPDIR

tsukasa-au: I agree that it would be better to put these somewhere other than `/tmp`, but the existing script uses /tmp a lot.

Would you prefer I switch just _this_ (new) usage of `/tmp` to `${TMPDIR}`, or should I switch the other usages of `/tmp` in this script as well?

agatti: I'd be easier to read if there's one commit that replaces all existing instances of `/tmp` to `${TMPDIR:-/tmp}` first, and then have the existing changes into a follow-up commit.  Whether the commits should be split into two separate PRs or not is a bit of personal preference IMHO (technically the first commit could be a valid PR in its own right).

As far as I know GitHub doesn't allow declaring dependencies on PRs, so let's keep both commits in here to make sure nothing gets lost if this gets merged.  Thanks!

tsukasa-au: Done.

agatti (approved):

tsukasa-au:
I noticed that you approved this change.

Is there anything else I need to do? I don't believe I have commit access to this repo, so I am unable to merge/push this change myself.

dpgeorge (approved):
A nice improvement, thanks!

PR #1105, opened 2026-04-08
unittest: Make more CPython like, add more tests, and fix bugs

tsukasa-au:
NOTE: This code is based on the CPython unittest implementation, and probably falls under the Python License.
    
The main things I wanted to accomplish with my changes to unittest were:
- Add `TestCase.run(test_result)`, so that users can (easily) wrap their tests, allowing them to output more information in the case of a test failure.
  + See the `test_run_method_overridable` test in `test_basics.py` for an example.
- Ensure that each test runs in its own `TestCase` instance
  + This makes it possible to move some per-test setup code from `setUp` to `__init__`.
- Ensure that the `TestCase` class had `setUp`, `tearDown`, `setUpClass`, and `tearDownClass` methods
  + This allows all subclasses to blindly call their super method, without having to check that it exists.

While working through this code I ended up adding more tests, then fixing the issues identified.

As part of the (partial)-rewrite, I ended up using `ucontextlib.contextmanager` in 3 places. I found this helper made it easier to implement `unittest.subTest` (which was previously implemented with a hand-written context manager), and the `_Outcome` helper class (the implementation of which was previously spread across `SubtestContext`, `_handle_test_exception`, and `_run_suite`).

From what I can see, `ucontextlib` is quite small when compiled (it is mostly comments and doc strings).


dpgeorge:
Thanks for the contribution.  And thanks for clearly explaining the reasoning/goals for making these changes.

> This code is based on the CPython unittest implementation, and probably falls under the Python License.

That is probably a show stopper, since the existing unittest code here is a clean-room implementation with MIT license.

Also, we need to keep the `unittest` module relatively light weight, so it can run on very constrained devices.  See #1125 which tries to optimise it further, based on a real use case running it on a device with 20k heap RAM.

For comparison, this PR increases `__init__.mpy` from 5795 bytes to 7711 bytes, which is quite a significant increase.  It also increases runtime memory use (measured using `import unittest`) from 9648 bytes to 14640 bytes.  That makes it almost unusable on boards with 20k heap RAM (because they still need headroom to load and execute the actual test).

Is there a way to achieve your goals by (1) not copying CPython code, and (2) only modifying the existing `unittest` in a minimal way?  The other option is to provide an extension module which adds the extra functionality, like `unittest-discover` does.

PR #1111, opened 2026-05-09
lora/lora-lorawan: Add LoRaWAN 1.0.x MAC layer

ddtdanilo:
## Summary

Add `lora-lorawan`: a LoRaWAN 1.0.x MAC layer for MicroPython.

- OTAA and ABP activation modes
- Class A device operation (TX → RX1 → RX2)
- AES-128-CMAC (RFC 4493) and CTR-mode crypto built on `ucryptolib`
- MIC calculation, payload encryption, session-key derivation

The MAC layer is radio-agnostic: it consumes a `radio` object that exposes
the `lora` send/recv API. Tested with [lora-sx127x-pycom](https://github.com/micropython/micropython-lib/pull/1102) on Pycom LoPy/LoPy4 boards.

Split out of micropython/micropython-lib#1102 per @projectgus's request to
make the driver and the MAC layer reviewable independently.

## Testing

- All files pass `ruff check` and `ruff format`
- LoRaWAN crypto verified against RFC 4493 test vectors
- End-to-end OTAA join + uplink against The Things Network on a LoPy4 (Pycom SX1276)

## Related

- Driver: micropython/micropython-lib#1102 (`lora-sx127x-pycom`)
- Boards consuming this package: micropython/micropython#19026 (PYCOM_LOPY, PYCOM_LOPY4)

### DIFF
--- a/micropython/lora/lora-lorawan/lora/lorawan.py
+++ b/micropython/lora/lora-lorawan/lora/lorawan.py
@@ -1,0 +1,397 @@
+"""
+LoRaWAN 1.0.x MAC layer for MicroPython.
+
+Implements Class A device operation with OTAA and ABP activation.
+Runs on top of the SX127x raw LoRa driver.
+
+Usage (OTAA):
+    from lora import LoRa
+    from lorawan import LoRaWAN
+
+    radio = LoRa(frequency=868100000, sf=7, bw=125000)
+    wan = LoRaWAN(radio, mode=LoRaWAN.OTAA,
+                  dev_eui=bytes.fromhex('...'),
+                  app_eui=bytes.fromhex('...'),
+                  app_key=bytes.fromhex('...'))
+    wan.join()
+    wan.send(1, b'Hello')
+
+Usage (ABP):
+    wan = LoRaWAN(radio, mode=LoRaWAN.ABP,
+                  dev_addr=0x01234567,
+                  nwk_skey=bytes.fromhex('...'),
+                  app_skey=bytes.fromhex('...'))
+    wan.send(1, b'Hello')
+"""
+
+from micropython import const
+import struct
+import time
+import os
+from lora.lorawan_crypto import (
+    aes_cmac,
+    aes_ctr_encrypt,
+    aes_ctr_decrypt,
+    compute_mic,
+    derive_session_keys,
+)
+
+# LoRaWAN MHDR message types
+_MTYPE_JOIN_REQUEST = const(0x00)
+_MTYPE_JOIN_ACCEPT = const(0x20)
+_MTYPE_UNCONFIRMED_UP = const(0x40)
+_MTYPE_UNCONFIRMED_DOWN = const(0x60)
+_MTYPE_CONFIRMED_UP = const(0x80)
+_MTYPE_CONFIRMED_DOWN = const(0xA0)
+
+# Direction constants
+_DIR_UP = const(0)
+_DIR_DOWN = const(1)
+
+# LoRaWAN data rate table for EU868
+_DR_TABLE_EU868 = {
+    0: (12, 125000),
+    1: (11, 125000),
+    2: (10, 125000),
+    3: (9, 125000),
+    4: (8, 125000),
+    5: (7, 125000),
+}
+
+# Default RX2 parameters (EU868)
+_RX2_FREQ = 869525000
+_RX2_DR = 0  # SF12/125kHz
+
+# Class A timing
+_JOIN_ACCEPT_DELAY1 = 5000  # ms
+_JOIN_ACCEPT_DELAY2 = 6000  # ms
+_RECEIVE_DELAY1 = 1000  # ms
+_RECEIVE_DELAY2 = 2000  # ms
+
+
+class LoRaWAN:
+    """LoRaWAN 1.0.x Class A device."""
+
+    OTAA = 0
+    ABP = 1
+
+    def __init__(
+        self,
+        radio,
+        mode=0,
+        dev_eui=None,
+        app_eui=None,
+        app_key=None,
+        dev_addr=None,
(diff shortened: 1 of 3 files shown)


ddtdanilo:
Hi @projectgus — gentle ping on the LoRaWAN side of the split. #1111 holds the `lora-lorawan` MAC layer on its own branch, CI green, no dependency edge with #1102 so they can be reviewed independently and in any order. Let me know if you'd like the scope or structure tweaked. Thanks!

micropython/lora/lora-lorawan/lora/lorawan.py
```diff
@@ -0,0 +1,397 @@
...
+
+            # RX1 window: JoinAcceptDelay1
+            accept = self._rx_window(tx_end, _JOIN_ACCEPT_DELAY1)
+            if accept:
+                if self._process_join_accept(accept, dev_nonce):
+                    self._joined = True
+                    return True
+
+            # RX2 window: JoinAcceptDelay2 at RX2 frequency/DR
+            sf, bw = _DR_TABLE_EU868[_RX2_DR]
```
projectgus, line 176: Need a way to make data rates configurable rather than hardcoded to one region (for example, I'm in Australia so I legally cannot even test EU868 settings!)

micropython/lora/lora-lorawan/lora/lorawan.py
```diff
@@ -0,0 +1,397 @@
...
+LoRaWAN 1.0.x MAC layer for MicroPython.
+
+Implements Class A device operation with OTAA and ABP activation.
+Runs on top of the SX127x raw LoRa driver.
+
+Usage (OTAA):
+    from lora import LoRa
+    from lorawan import LoRaWAN
+
+    radio = LoRa(frequency=868100000, sf=7, bw=125000)
```
projectgus, line 11: This will need to work with the existing LoRa modem drivers as well. (See related discussion at https://github.com/micropython/micropython-lib/pull/1102#issuecomment-4550018042)

Specifically here: The existing drivers pass the hardware-specific config as constructor kwargs, but the LoRa radio settings are passed in a `lora_cfg` dictionary which is non-hardware-specific so it's easy to factor code to work with different modem drivers.

micropython/lora/lora-lorawan/lora/lorawan_crypto.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def _cmac_subkeys(key):
+    """Derive CMAC sub-keys K1, K2 from AES key (RFC 4493 section 2.3)."""
+    L = _aes_ecb(key, b"\x00" * _BLK)
+    K1 = _shift_left_1(L)
+    if L[0] & 0x80:
+        K1 = _xor(K1, b"\x00" * 15 + b"\x87")
+    K2 = _shift_left_1(K1)
+    if K1[0] & 0x80:
```
projectgus, line 43: I have some mild concern about whether any of these kind of branches create timing side channels.

Naive way to make this "constant time" would be perhaps

```py
K2 = _xor(K2, b"\x00" * 15 + (b"\x87" if (K1[0] & 0x80) else b"\x00"))
```

Whether this is actually better for timing would need to be measured.

Alternatively, mbedTLS has AES-CMAC support so we could expose the primitives from there via cryptolib. Pros: faster, not rolling own crypto. Cons: binary size would probably be higher, and on all boards... :/

micropython/lora/lora-lorawan/lora/lorawan.py
```diff
@@ -0,0 +1,397 @@
...
+            return self._process_downlink(rx_data)
+
+        return None
+
+    def _rx_window(self, tx_end, delay_ms):
+        """Open a receive window at tx_end + delay_ms."""
+        now = time.ticks_ms()
+        wait = time.ticks_diff(tx_end + delay_ms, now)
+        if wait > 0:
+            time.sleep_ms(wait)
```
projectgus, line 256: Not asking for it in this PR, it would be interesting to see if this driver could be made async so an async task is blocking for the correct window to open, while the code is doing something else. (As async is non-preemptive this might turn out to not be tight enough on timing, though.)

micropython/lora/lora-lorawan/lora/lorawan.py
```diff
@@ -0,0 +1,397 @@
...
+        """Process and validate a join-accept message."""
+        if len(data) < 17:
+            return False
+
+        # Decrypt join-accept (the entire payload after MHDR is encrypted)
+        mhdr = data[0]
+        if (mhdr & 0xE0) != _MTYPE_JOIN_ACCEPT:
+            return False
+
+        from ucryptolib import aes
```
projectgus, line 277:
```suggestion
        from cryptolib import aes
```

(We deprecated the u-prefix a couple of versions ago.)

Also please move all imports to the top, unless there's a runtime reason why this code path should be optional.

micropython/lora/lora-lorawan/lora/lorawan_crypto.py
```diff
@@ -0,0 +1,153 @@
+"""
+LoRaWAN cryptographic primitives (AES-128-CMAC, CTR, key derivation).
+
+Uses ucryptolib.aes (AES-128-ECB) as the only hardware dependency.
```
projectgus, line 4:
```suggestion
Uses cryptolib.aes (AES-128-ECB) as the only hardware dependency.
```

(and same below)

projectgus (commented):
This is neat! A lot less code than I expected for a functional LoRaWAN Class A node as well, nice work.

I gave this a quick skim and put a few notes below. My main concern at this point is the same one raised in #1102 - we already have a LoRa modem driver API and we can't merge a second incompatible API.

If we can resolve that in #1102, and then this PR is made to work with that, then I think we'd be on a good path to getting this merged as well.

Issue #838, opened 2024-04-01
SECURITY: Requests module HTTPS - no server certificate verification.


jonfoster:
While looking at the MicroPython Requests module (on the git HEAD), I noticed this nightmare:

```
            context = tls.SSLContext(tls.PROTOCOL_TLS_CLIENT)
            context.verify_mode = tls.CERT_NONE
            s = context.wrap_socket(s, server_hostname=host)
```

Assuming that it has the same meaning in MicroPython as cPython (I haven't checked), that line in the middle totally disables TLS security.  The attacker pretending to be the server can send any certificate they like, and the client will blindly accept it.

If people are using HTTPS as "the new HTTP", and are happy with the HTTP you-get-no-security model, that's fine.  But anyone relying on HTTPS for security, and expecting the normal level of security you'd get from HTTPS, is going to be in trouble.

At a minimum this should be documented clearly on the MicroPython requests documentation ... which doesn't seem to exist anywhere?

Ideally, MicroPython should default to a proper secure HTTPS implementation, including certificate verification, and have a way to opt-out.

sosi-deadeye:
The problem is, that a `chain.der` is required.
It must be loaded before the check could be done.

```
esp32@ganymed:~/micropython/ports/unix/build-standard$ ./micropython c.py
Traceback (most recent call last):
  File "c.py", line 18, in <module>
  File "ssl.py", line 1, in wrap_socket
OSError: (-30336, 'MBEDTLS_ERR_SSL_CA_CHAIN_REQUIRED')
```

With `ctx.load_verify_locations(cafile="chain.der")` I'm able to load the cafile and the check works.
The `chain.der` has a size of ~1 kB. I guess this is why this option is set to `None`.
Flash is limited on microcontrollers.

But it should be documented, that the default behavior is unsafe and the why.

jonfoster:
Yes, you need root certificates.  Some embedded platforms have root certificates built-in as part of the standard manufacturer-supplied drivers.  ([Example for ESP32](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/protocols/esp_crt_bundle.html)).  Ideally, MicroPython should use those, if available.  In cPython, that's done with the [SSLContext.load_default_certs()](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_default_certs) method.

sosi-deadeye:
I used this command to convert the existing **ca-bundle** on my system into `certs.der`:
```
[...]

Issue #777, opened 2023-12-09
requests: stream parameter is noop


tmplt:
`MicroPython v1.21.0; Raspberry Pi Pico W with RP2040`.

`requests.request` exposes a `stream=None` parameter:

https://github.com/micropython/micropython-lib/blob/d8e163bb5f3ef45e71e145c27bc4f207beaad70f/python-ecosys/requests/requests/__init__.py#L36-L46

but it is never functionally used, nor is some `iter_content` API exposed; the parameter is only forwarded to recursive `request` calls.

Workaround:
```python
r = requests.get(...)
# sidestep @property content, which just reads socket data into memory until EOF
s = r.raw
r.raw = None
buf = bytearray(1024)
while s.readinto(buf) > 0:
    # handle data in buf
```

if the above is implemented into some `iter_content`, is the stream parameter even needed? Am I missing something?

PR #1119, opened 2026-06-01
requests: HTTP/1.1 response bodies and max_body limit

### LINKED ISSUE #931: "requests" does not handle redirection with relative path correctly
In [python-ecosys/requests/requests/__init__.py:183](https://github.com/micropython/micropython-lib/blob/68e3e07bc7ab63931cead3854b2a114e9a084248/python-ecosys/requests/requests/__init__.py#L183),

The url `redirect` can be a relative path, which cause the missing of hostname and protocal at redirecting.

### LINKED ISSUE #1118: requests: HTTP/1.1 response bodies and max_body limit
## Summary

Extend the `requests` package with essential HTTP/1.1 client behaviour:

- Send requests as HTTP/1.1 (default), keep Connection: close
- Read response bodies with Content-Length and chunked Transfer-Encoding
- Case-insensitive response headers
- Fix relative Location in redirects (closes #931)
- Default max response size 32 KiB (`max_body` kwarg to override/disable)

## Out of scope (follow-up PRs)

- TLS certificate verification (#838)
- Real stream=True / incremental body read (#777)
- gzip Content-Encoding, keep-alive pool, cookies

## Testing

Unix port + ESP32 with fork mip index; extend test_requests.py socket mocks.

PR will follow from branch `requests-http11-essentials`.

pablogventura:
### Summary

I use `requests` on small boards fairly often, and the current package left me hanging on
chunked responses (`raw.read()` never finished) and still spoke HTTP/1.0 on the wire. This PR
is the first slice of the HTTP work I outlined in #1118: enough client behaviour for real
servers without pulling in TLS or streaming yet.

What I changed:

- Added `requests/_http.py` to parse status/headers and read bodies (Content-Length and
  chunked). The full body lands in `Response.content` so callers do not have to fight the
  socket after the headers.
- Requests now use **HTTP/1.1** with **`Connection: close`** (no keep-alive in this PR).
- `resp.headers` is a small `Headers` mapping with case-insensitive keys.
- Relative `Location` values on redirects are resolved against the original URL (fixes #931).
- New `max_body` argument, default **32 KiB**; pass `max_body=None` if you really want no cap.

Closes #1118.

### Testing

I started with the existing socket-mock tests and extended them for bodies, chunked encoding,
redirects, and `max_body`. On my machine:

```text
python3 -S python-ecosys/requests/test_requests.py
```

All **14** mock tests pass on CPython 3.12.

I also ran the same file on the **unix** port (`build-standard`) and added `test_requests_live.py`
(local TCP server in a thread) plus `run-unix-tests.sh` — **14 + 4** tests, all green.

On hardware I flashed **ESP32_GENERIC** (ESP32-D0WD, 4 MiB flash), copied this branch’s package
to `/lib/requests` (the board still ships an older frozen `requests`, so tests prefer `/lib`),
and ran `run-esp32-tests.sh` (mock + live). Same counts, all pass.

Finally I connected the board to my home WiFi and hit real HTTP endpoints (`example.com`,
`httpbin.org` GET + redirect) — status 200, bodies as expected. I have not added HTTPS here;
that stays for a follow-up (#838).

CI in this repo should pick up `python-ecosys/requests/test_requests.py` with the project’s
MicroPython build; the live/ESP32 scripts are mainly for local regression.

### Trade-offs and alternatives

- The entire body is buffered in RAM. That is deliberate for simplicity on microcontrollers; I
  documented it and added `max_body` so a runaway response does not silently exhaust memory.
- No keep-alive, gzip, or TLS certificate verification in this PR — I plan those separately
  (#838, #777).
- Outgoing header order still follows dict insertion order, same as before (I did not sort
  headers for diff stability).

### Generative AI

I used generative AI tools while working on this PR (including help polishing this
description). I reviewed all of the code and take responsibility for the changes and the
text above.

### DIFF
--- a/python-ecosys/requests/README.md
+++ b/python-ecosys/requests/README.md
@@ -4,7 +4,9 @@ This module provides a lightweight version of the Python
 [requests](https://requests.readthedocs.io/en/latest/) library.
 
 It includes support for all HTTP verbs, https, json decoding of responses,
-redirects, basic authentication.
+redirects (including relative Location URLs), basic authentication, HTTP/1.1
+requests, and reading response bodies with Content-Length or chunked
+Transfer-Encoding.
 
 ### Limitations
 
@@ -13,4 +15,12 @@ redirects, basic authentication.
   multipart-form encoding of post data (this can be done manually).
 * Compressed requests/responses are not currently supported.
 * File upload is not supported.
-* Chunked encoding in responses is not supported.
+* HTTP keep-alive connection reuse is not supported (Connection: close by default).
+* Response bodies are buffered in memory. By default at most 32 KiB are read
+  per response; pass ``max_body=None`` to disable the limit or set another
+  byte count with ``max_body=N``.
+
+### Follow-up work
+
+* TLS certificate verification (see micropython-lib issue #838).
+* ``stream=True`` incremental body reads (see issue #777).
--- a/python-ecosys/requests/manifest.py
+++ b/python-ecosys/requests/manifest.py
@@ -1,3 +1,3 @@
-metadata(version="0.10.2", pypi="requests")
+metadata(version="0.11.0", pypi="requests")
 
 package("requests")
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -63,7 +63,7 @@ def request(
         import binascii
 
         username, password = auth
-        formatted = (username + ":" + password).encode()
+        formatted = _to_bytes(username) + b":" + _to_bytes(password)
         formatted = str(binascii.b2a_base64(formatted)[:-1], "ascii")
         headers["Authorization"] = "Basic {}".format(formatted)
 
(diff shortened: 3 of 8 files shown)


dpgeorge:
Thanks for the contribution.

> The entire body is buffered in RAM. That is deliberate for simplicity on microcontrollers;

Unfortunately this change is a non-starter.  Microcontrollers need to have simple code -- yes that's true -- but more importantly they need to be able to run in limited RAM.

The current Response object has a `.raw` member which points to the underlying stream and allows the user to stream out the body as necessary.  Eg reading a small amount of data at a time into a pre-allocated buffer.  And then there's also `.content`, `.text` and `.json` properties/methods which allow convenient access to the body if you don't care to access `.raw`.  This is all compatible with the upstream CPython requests module and is microcontroller friendly (those two things are hard to achieve!).

This PR removes `.raw` (makes it a closed socket) which is a non-starter.

I suggest you start instead with a new PR and the simplest set of changes that implement HTTP/1.1.  That will be very useful.

pablogventura:
Closing in favor of #1124. Reworked to preserve `.raw` streaming per review feedback. This PR is intentionally minimal: HTTP/1.1 + Content-Length only; chunked and other features will follow in separate PRs.

PR #1121, opened 2026-06-10
requests: Add context manager for response object.

Gadgetoid:
### Summary

When accessing `response.raw` it's not always obvious you should call `response.close()` to close and free the socket. Since lwIP has a five socket limit by default this can very quickly result in cryptic ENOMEM errors.

A context manager is widely understood to handle freeing resources on exit and permits the following, Pythonic pattern without leaking open sockets:

```python
buffer = bytearray(1024)
with requests.get("some-resource") as response:
    while length := response.raw.readinto(buffer)
        output.write(memoryview(buffer)[:length])
```

While I came up with this idea and wrote this code independently , this PR transpired to be partially a dupe of https://github.com/micropython/micropython-lib/pull/278 but I have excluded the iteration features in favour of a clean, concise change in the hope it would be easier to consider.

### Testing

This is a minor change adding well understood magic methods, I did a cursory check to make sure it would work.

This is a very minor change for a lot of potential utility, though without associated documentation changes it's unlikely to receive adoption.

CPython's "requests" library has the concept of a `Session`, which includes a context manager at a lower level, and implements get/post etc using a transient Session. This is a potentially useful concept in MicroPython because the time cost of re-opening a socket when TLS is involved is prohibitive - 300-400ms. I'm currently using a completely custom HTTP library to avoid this.

This change also has some minor functional overlap with `urllib urequest`'s `urlopen` which is sometimes favoured for streaming requests.

### Generative AI

### DIFF
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -32,6 +32,12 @@ class Response:
 
         return json.loads(self.content)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
+
 
 def request(
     method,


python-ecosys/requests/requests/__init__.py
```diff
@@ -32,6 +32,12 @@ def json(self):
 
         return json.loads(self.content)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
agatti, line 38: A bit of a controversial take, but replacing the non-self arguments list with a single `*_` saves 28 bytes.

I'd say it's safe to assume that a context manager will always pass the correct arguments to `__exit__`, although that way will look quite uglier than what you initially wrote :)

Gadgetoid: > A bit of a controversial take, but replacing the non-self arguments list with a single *_ saves 28 bytes.

I did a quick search of the codebase and picked the most common pattern. I did see `*args` and my personal preference would probably be `_type, _value, _traceback` to both flag as unused *and* keep it unambiguous 😆

Might not hurt to get some consensus on this and raise a PR to make every `__exit__` in micropython-lib look the same.

dpgeorge: Using `*args` will lead to an allocation of a tuple at runtime whenever the function is called (to hold all the arguments).

So I suggest just keeping this as-is for now.

dpgeorge (approved):
This looks good to me, nice and simple, and matches the official requests (which has the same context manager on `Response` objects).

dpgeorge:
Merged in acf98ad2c0c9727cae06c6748d0c30da5d4b7682 (I also bumped the version in `manifest.py`).

PR #1122, opened 2026-06-11
unix-ffi/ffilib/ffilib.py: Extend libc versions range.

### LINKED ISSUE #962: Unix port: ffilib.py: libc.so.6 is no longer the latest version
While using `os.popen()` from unix-ffi libraries on FreeBSD 14.1-RELEASE-p5 amd64, I encountered this error:

```console
...
  File "os/__init__.py", line 32, in <module>
  File "ffilib.py", line 43, in libc
  File "ffilib.py", line 39, in open
  File "ffilib.py", line 33, in open
OSError: [Errno 2] ENOENT
```

On further investigation, I found that libc major version has been upgraded to 7, and ffilib.py is only looking for versions up to 6.

```console
$ ls /lib/libc.so*
/lib/libc.so.7
```

https://github.com/micropython/micropython-lib/blob/e4cf09527bce7569f5db742cf6ae9db68d50c6a9/unix-ffi/ffilib/ffilib.py#L42

Changing the 6 to 7 fixed the issue.

Only making a note here since the readme for unix-ffi says: "This directory is unmaintained." and I would like to have a reference come up in searches when someone encounters the error.

agatti:
### Summary

This PR extends the range of libc library versions that are looked up when required by user code.

FreeBSD 14.1-RELEASE and later versions require an update to the range since they bumped up their libc major version, from 6 to 7.  This has the side-effect of guaranteeing at least one failed lookup on modern Linux systems, where the libc is still at version 6.

With updated FreeBSD support in MicroPython core, now `sys.platform` properly returns `freebsd`.  This required a minor change in the code deciding whether to look for `.so`, `.dylib`, or `.dll` files, since `linux` is no longer the default fallback for generic unix systems.

(Hopefully) closes #962.

### Testing

Using an interpreter built from MicroPython git `master` with `require("ffilib")` added to the manifest,  `import ffilib; ffilib.libc()` was called making sure the function returned a valid `ffimod` object.

This was done manually both on a current Arch Linux/x64 system and on a FreeBSD 15.0-RELEASE x64 VM.

This comes with a 13 bytes increase, mostly to hold the `freebsd` string in the MPY file itself.
### Generative AI

### DIFF
--- a/unix-ffi/ffilib/ffilib.py
+++ b/unix-ffi/ffilib/ffilib.py
@@ -17,7 +17,7 @@ def open(name, maxver=10, extra=()):
         pass
 
     def libs():
-        if sys.platform == "linux":
+        if sys.platform in ("linux", "freebsd"):
             yield "%s.so" % name
             for i in range(maxver, -1, -1):
                 yield "%s.so.%u" % (name, i)
@@ -39,7 +39,7 @@ def open(name, maxver=10, extra=()):
 
 
 def libc():
-    return open("libc", 6)
+    return open("libc", 7)
 
 
 # Find out bitness of the platform, even if long ints are not supported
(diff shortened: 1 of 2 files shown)


dpgeorge (approved):
Looks good, thanks!

PR #1124, opened 2026-06-19
requests: HTTP/1.1 with Content-Length and streaming raw

pablogventura:
## Summary

- Send HTTP/1.1 requests (was HTTP/1.0).
- Parse response status/headers via a small `_http` module.
- Support `Content-Length` response bodies without buffering in `request()`.
- Preserve `.raw` as a live `BodyStream` wrapper; `.content` remains lazy.

Closes the approach in #1119 (which buffered the full body and closed the socket). Chunked responses, `max_body`, and relative redirects are intentionally left for follow-up PRs.

## Testing

- `micropython test_requests.py` (unix port) — 13 mock tests including incremental `.raw.read()`.
- CI: ruff, codespell, package tests.

## Trade-offs and Alternatives

- Chunked response bodies still raise `ValueError` (unchanged from master until a follow-up PR).
- `BodyStream` adds a small wrapper object (~two fields) instead of buffering the body in `request()`.
- `stream=True` is not implemented yet (issue #777).

## Generative AI

### DIFF
--- a/python-ecosys/requests/README.md
+++ b/python-ecosys/requests/README.md
@@ -4,7 +4,8 @@ This module provides a lightweight version of the Python
 [requests](https://requests.readthedocs.io/en/latest/) library.
 
 It includes support for all HTTP verbs, https, json decoding of responses,
-redirects, basic authentication.
+redirects, basic authentication, HTTP/1.1 requests, and reading response
+bodies with Content-Length via streaming ``.raw`` or lazy ``.content``.
 
 ### Limitations
 
@@ -14,3 +15,10 @@ redirects, basic authentication.
 * Compressed requests/responses are not currently supported.
 * File upload is not supported.
 * Chunked encoding in responses is not supported.
+* HTTP keep-alive connection reuse is not supported (Connection: close by default).
+
+### Follow-up work
+
+* Chunked response bodies.
+* TLS certificate verification (see micropython-lib issue #838).
+* ``stream=True`` incremental body reads (see issue #777).
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -1,6 +1,37 @@
 import socket
 
 
+class BodyStream:
+    def __init__(self, sock, remaining):
+        self._sock = sock
+        self._remaining = remaining
+
+    def read(self, n=-1):
+        if self._remaining == 0:
+            return b""
+        if n < 0 or n > self._remaining:
+            n = self._remaining
+        data = self._sock.read(n)
+        self._remaining -= len(data)
+        if not data:
+            raise ValueError("Connection closed before Content-Length satisfied")
+        return data
+
+    def readinto(self, buf):
+        if self._remaining == 0:
+            return 0
+        if len(buf) > self._remaining:
+            buf = memoryview(buf)[: self._remaining]
+        got = self._sock.readinto(buf)
+        self._remaining -= got
+        if not got:
+            raise ValueError("Connection closed before Content-Length satisfied")
+        return got
+
+    def close(self):
+        self._sock.close()
+
+
 class Response:
     def __init__(self, f):
         self.raw = f
@@ -104,7 +135,7 @@ def request(
             context = tls.SSLContext(tls.PROTOCOL_TLS_CLIENT)
             context.verify_mode = tls.CERT_NONE
             s = context.wrap_socket(s, server_hostname=host)
-        s.write(b"%s /%s HTTP/1.0\r\n" % (method, path))
+        s.write(b"%s /%s HTTP/1.1\r\n" % (method, path))
 
         if "Host" not in headers:
             headers["Host"] = host
@@ -161,6 +192,7 @@ def request(
         reason = ""
         if len(l) > 2:
             reason = l[2].rstrip()
+        remaining = None
         while True:
             l = s.readline()
             if not l or l == b"\r\n":
@@ -179,7 +211,10 @@ def request(
             elif parse_headers is True:
                 l = str(l, "utf-8")
                 k, v = l.split(":", 1)
-                resp_d[k] = v.strip()
+                v = v.strip()
+                resp_d[k] = v
+                if k.lower() == "content-length":
+                    remaining = int(v)
             else:
                 parse_headers(l, resp_d)
     except OSError:
@@ -195,7 +230,10 @@ def request(
         else:
             return request(method, redirect, data, json, headers, stream)
     else:
-        resp = Response(s)
+        if remaining is not None:
+            resp = Response(BodyStream(s, remaining))
+        else:
+            resp = Response(s)
         resp.status_code = status
         resp.reason = reason
         if resp_d is not None:
--- a/python-ecosys/requests/test_requests.py
+++ b/python-ecosys/requests/test_requests.py
@@ -3,9 +3,9 @@ import sys
 
 
 class Socket:
-    def __init__(self):
+    def __init__(self, read_data=b"HTTP/1.1 200 OK\r\n\r\n"):
         self._write_buffer = io.BytesIO()
-        self._read_buffer = io.BytesIO(b"HTTP/1.0 200 OK\r\n\r\n")
+        self._read_buffer = io.BytesIO(read_data)
 
     def connect(self, address):
         pass
@@ -15,6 +15,15 @@ class Socket:
 
     def readline(self):
         return self._read_buffer.readline()
+
+    def read(self, size=-1):
+        return self._read_buffer.read(size)
+
+    def readinto(self, buf):
+        return self._read_buffer.readinto(buf)
+
+    def close(self):
+        pass
 
 
 class socket:
@@ -43,7 +52,7 @@ def test_simple_get():
     response = requests.request("GET", "http://example.com")
 
     assert response.raw._write_buffer.getvalue() == (
-        b"GET / HTTP/1.0\r\n" + b"Connection: close\r\n" + b"Host: example.com\r\n\r\n"
+        b"GET / HTTP/1.1\r\n" + b"Connection: close\r\n" + b"Host: example.com\r\n\r\n"
     ), format_message(response)
 
 
@@ -53,7 +62,7 @@ def test_get_auth():
     )
 
     assert response.raw._write_buffer.getvalue() == (
-        b"GET / HTTP/1.0\r\n"
+        b"GET / HTTP/1.1\r\n"
         + b"Host: example.com\r\n"
         + b"Authorization: Basic dGVzdC11c2VybmFtZTp0ZXN0LXBhc3N3b3Jk\r\n"
         + b"Connection: close\r\n\r\n"
@@ -64,7 +73,7 @@ def test_get_custom_header():
     response = requests.request("GET", "http://example.com", headers={"User-Agent": "test-agent"})
 
     assert response.raw._write_buffer.getvalue() == (
-        b"GET / HTTP/1.0\r\n"
+        b"GET / HTTP/1.1\r\n"
         + b"User-Agent: test-agent\r\n"
         + b"Host: example.com\r\n"
         + b"Connection: close\r\n\r\n"
@@ -75,7 +84,7 @@ def test_post_json():
     response = requests.request("GET", "http://example.com", json="test")
 
     assert response.raw._write_buffer.getvalue() == (
-        b"GET / HTTP/1.0\r\n"
+        b"GET / HTTP/1.1\r\n"
         + b"Connection: close\r\n"
         + b"Content-Type: application/json\r\n"
         + b"Host: example.com\r\n"
@@ -91,7 +100,7 @@ def test_post_chunked_data():
     response = requests.request("GET", "http://example.com", data=chunks())
 
     assert response.raw._write_buffer.getvalue() == (
-        b"GET / HTTP/1.0\r\n"
+        b"GET / HTTP/1.1\r\n"
         + b"Transfer-Encoding: chunked\r\n"
         + b"Host: example.com\r\n"
         + b"Connection: close\r\n\r\n"
@@ -106,7 +115,7 @@ def test_overwrite_get_headers():
     )
 
     assert response.raw._write_buffer.getvalue() == (
-        b"GET / HTTP/1.0\r\n" + b"Connection: keep-alive\r\n" + b"Host: test.com\r\n\r\n"
+        b"GET / HTTP/1.1\r\n" + b"Connection: keep-alive\r\n" + b"Host: test.com\r\n\r\n"
     ), format_message(response)
 
 
@@ -119,7 +128,7 @@ def test_overwrite_post_json_headers():
     )
 
     assert response.raw._write_buffer.getvalue() == (
-        b"GET / HTTP/1.0\r\n"
+        b"GET / HTTP/1.1\r\n"
         + b"Connection: close\r\n"
         + b"Content-Length: 10\r\n"
         + b"Content-Type: text/plain\r\n"
@@ -137,7 +146,7 @@ def test_overwrite_post_chunked_data_headers():
     )
 
     assert response.raw._write_buffer.getvalue() == (
-        b"GET / HTTP/1.0\r\n"
+        b"GET / HTTP/1.1\r\n"
         + b"Host: example.com\r\n"
         + b"Content-Length: 4\r\n"
         + b"Connection: close\r\n\r\n"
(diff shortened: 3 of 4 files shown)


pablogventura:
## ESP32 hardware testing

Tested on device connected via `/dev/ttyACM0`:

| | |
|---|---|
| **Board** | ESP32 |
| **Firmware** | MicroPython v1.29.0-preview.337 (2026-06-01) |
| **WiFi** | LosMius (STA), device IP `192.168.1.151` |
| **Package deployed** | `/lib/requests/__init__.py` + `/lib/requests/_http.py` from this PR |

### Mock tests (`test_requests.py`) — 13/13 OK

All existing mock tests pass on-device (outbound HTTP/1.1, Content-Length inbound, incremental `.raw.read()`, chunked response still raises `ValueError`).

### Live tests (LAN HTTP server at `192.168.1.72:58080`) — 4/4 OK

| Test | Result |
|---|---|
| `GET /get` with Content-Length body | 200, 11 bytes |
| `GET /bytes/8` via `.content` | `abcdefgh`, socket closed after read |
| `GET /bytes/10` incremental `.raw.read(3)` + `.raw.read(3)` + `.raw.read()` | `3+3+4` bytes → `0123456789` |
| `GET /chunked` | `ValueError: Unsupported Transfer-Encoding: chunked` (expected for v1) |

Server log confirmed outbound requests use **HTTP/1.1**:

```
GET /get HTTP/1.1
GET /bytes/8 HTTP/1.1
GET /bytes/10 HTTP/1.1
GET /chunked HTTP/1.1
```

### Notes

- `httpbin.org` returned 503 from this network, so live body tests used a local test server instead.
- Mock tests replace `sys.modules["socket"]`; live tests reload `usocket` afterwards before hitting the network.

python-ecosys/requests/requests/__init__.py
```diff
@@ -1,5 +1,7 @@
 import socket
 
+from ._http import open_body, read_headers, read_status_line
```
dpgeorge: I don't think these functions need to be in a separate module.  MicroPython aims to be minimal, and we prefer minimal code over breaking things up into lots of modules.

So, please put everything from `_http.py` back in this file.  That will also make it easier to review the changes.

pablogventura:
Done — I moved everything from `_http.py` into `__init__.py` and dropped the extra module (commit 17f78a5).

Re-ran `test_requests.py` on the unix port locally (13/13). CI is green again on my side.

Let me know if anything else should change before you take another look.

pablogventura:
Also re-tested on the same ESP32 as in my earlier comment (ESP32-D0WD on `/dev/ttyACM0`, MicroPython v1.29 preview, WiFi at home). After inlining `_http.py` into `__init__.py`, mock tests still pass on-device (`test_requests.py`, 13/13) and live HTTP over the LAN worked as before.

python-ecosys/requests/requests/__init__.py
```diff
@@ -1,6 +1,81 @@
...
 
+def _header_get(headers, name):
+    name_lower = name.lower()
+    for k, v in headers.items():
+        if k.lower() == name_lower:
+            return v
+    return None
+
+
+def read_status_line(stream):
```
dpgeorge: This function is used only once so please inline it at its point of use.

python-ecosys/requests/requests/__init__.py
```diff
@@ -1,6 +1,81 @@
...
+        raise ValueError("HTTP error: empty status line")
+    parts = line.split(None, 2)
+    if len(parts) < 2:
+        raise ValueError("HTTP error: BadStatusLine:\n%s" % parts)
+    status = int(parts[1])
+    reason = parts[2].rstrip() if len(parts) > 2 else ""
+    return status, reason
+
+
+def read_headers(stream, parse_headers=True):
```
dpgeorge: This function is used only once so please inline it at its point of use.

python-ecosys/requests/requests/__init__.py
```diff
@@ -1,6 +1,81 @@
...
+            self._remaining -= len(data)
+            if self._remaining > 0 and not data:
+                raise ValueError("Connection closed before Content-Length satisfied")
+        return data
+
+    def close(self):
+        self._sock.close()
+
+
+def open_body(stream, headers):
```
dpgeorge: This function is used only once so please inline it at its point of use.

pablogventura:
Thanks for the review, @dpgeorge — much appreciated.

I've inlined `read_status_line`, `read_headers`, and `open_body` directly into `request()` (commit 03a85e3). `BodyStream` remains as the only helper class. `_header_get` was folded into the body-setup block as well.

CI should re-run shortly. Happy to adjust anything else.

pablogventura:
Re-tested after inlining the helpers and the ruff fix (commit d3c3628):

**Unix port** - `micropython test_requests.py`: 13/13 OK

**ESP32** (ESP32-D0WD, `/dev/ttyACM0`, MicroPython v1.29 preview, WiFi at home):
- Mock tests (`test_requests.py`): 13/13 OK
- Live HTTP over LAN (local test server): 4/4 OK (Content-Length, incremental `.raw.read()`, chunked -> `ValueError`)

CI re-running on d3c3628.

python-ecosys/requests/requests/__init__.py
```diff
@@ -145,37 +167,41 @@ def request(
...
-        # print(l)
-        l = l.split(None, 2)
-        if len(l) < 2:
-            # Invalid response
-            raise ValueError("HTTP error: BadStatusLine:\n%s" % l)
-        status = int(l[1])
-        reason = ""
-        if len(l) > 2:
-            reason = l[2].rstrip()
+        line = s.readline()
```
dpgeorge: Please keep original code unchanged if possible.  In this case it's just renaming `l` to `line`, and that doesn't really need to change.  It makes it much easier to review if the diff is minimal.

python-ecosys/requests/requests/__init__.py
```diff
@@ -145,37 +167,41 @@ def request(
...
+        if not line:
+            raise ValueError("HTTP error: empty status line")
```
dpgeorge: I don't think this check is needed.  If `line` is empty then the `len(parts) < 2` will catch that.

python-ecosys/requests/requests/__init__.py
```diff
@@ -145,37 +167,41 @@ def request(
...
+        parts = line.split(None, 2)
```
dpgeorge: Please keep this as `l = l.split(None, 2)` to keep the diff minimal (it's also more efficient bytecode to reuse the variable).

python-ecosys/requests/requests/__init__.py
```diff
@@ -145,37 +167,41 @@ def request(
...
+            raise ValueError("HTTP error: BadStatusLine:\n%s" % parts)
+        status = int(parts[1])
+        reason = parts[2].rstrip() if len(parts) > 2 else ""
+
+        resp_d = {}
         while True:
-            l = s.readline()
-            if not l or l == b"\r\n":
+            line = s.readline()
+            if not line or line == b"\r\n":
```
dpgeorge: Please keep these 2 lines as they were.

python-ecosys/requests/requests/__init__.py
```diff
@@ -145,37 +167,41 @@ def request(
...
+        line = s.readline()
+        if not line:
+            raise ValueError("HTTP error: empty status line")
+        parts = line.split(None, 2)
+        if len(parts) < 2:
+            raise ValueError("HTTP error: BadStatusLine:\n%s" % parts)
+        status = int(parts[1])
+        reason = parts[2].rstrip() if len(parts) > 2 else ""
+
+        resp_d = {}
```
dpgeorge: This can be set to `None` and save a dict alloc if `parse_headers` is `False`.  That's actually how the code was, so best to keep it as it was.

pablogventura:
Thanks again for the detailed review, @dpgeorge.

Addressed the latest round in the new commit:
- Restored `l`-based status/header parsing (no `line`/`parts` rename)
- Removed the extra empty status-line check
- Restored `reason = ""` / `if len(l) > 2:` as before
- Restored `resp_d = None` when `parse_headers is False`
- Restored in-loop `Transfer-Encoding` / `Location` handling from master
- Kept only the minimal additions: HTTP/1.1, `BodyStream`, Content-Length limit

Re-tested on unix port: `test_requests.py` 13/13 OK.

Happy to adjust further if needed.

python-ecosys/requests/requests/__init__.py
```diff
@@ -189,7 +215,13 @@ def request(
         else:
             return request(method, redirect, data, json, headers, stream)
     else:
-        resp = Response(s)
+        remaining = None
+        if resp_d is not None:
+            for k, v in resp_d.items():
+                if k.lower() == "content-length":
```
dpgeorge: To save code size, this check for `content-length` can probably go up in the code where `resp_d` is being populated.

python-ecosys/requests/requests/__init__.py
```diff
@@ -189,7 +215,13 @@ def request(
...
+                    remaining = int(v)
+                    break
+        resp = Response(BodyStream(s, remaining))
```
dpgeorge: If `remaining` is `None` I suggest not wrapping in `BodyStream`, it doesn't need it.  So that means the behaviour is unchanged from before if there's no `content-length`, and the `BodyStream` class can be simpler (because `remaining` is now non-None).

pablogventura:
Thanks for the follow-up comments, @dpgeorge.

Addressed in commit 8b20e5c:
- Detect `Content-Length` while populating `resp_d` in the header loop (removed the extra scan at the end)
- Use `Response(s)` when there is no Content-Length (unchanged behaviour from master)
- Wrap in `BodyStream` only when Content-Length is present; `BodyStream` now always takes a byte count

Re-tested on unix port: `test_requests.py` 13/13 OK. CI green on 8b20e5c.

python-ecosys/requests/requests/__init__.py
```diff
@@ -1,6 +1,26 @@
...
+
+    def read(self, n=-1):
+        if self._remaining == 0:
+            return b""
+        if n < 0 or n > self._remaining:
+            n = self._remaining
+        data = self._sock.read(n)
+        self._remaining -= len(data)
+        if self._remaining > 0 and not data:
+            raise ValueError("Connection closed before Content-Length satisfied")
```
dpgeorge, line 17: How does CPython's `requests` behave in this situation, does it raise an exception, or just return short data?

(Also, the check can be just `if not data:`.)

pablogventura:
Good question. I checked CPython 3.12 with requests 2.31 / urllib3 2.0.7 on a response with Content-Length: 10 but only 3 body bytes before close:

- `response.raw.read(n)` with partial data available: returns the short data (e.g. read(10) -> b"abc"), no exception on that call.
- A later read once the socket is exhausted but bytes are still expected per Content-Length: urllib3 raises IncompleteRead (requests wraps it as ProtocolError / ChunkedEncodingError).
- `.content` (full read): raises IncompleteRead.

So streaming behaviour is: partial reads are fine; exception when you try to read past EOF with Content-Length still outstanding. That matches what BodyStream does (ValueError instead of IncompleteRead, for simplicity).

Applied your suggestion in 6372c93: the check is now `if not data:`.
Re-tested on unix port: test_requests.py 13/13 OK.

dpgeorge:
Thanks for updating.

Looking at how `requests` is used by `mip`, `mip` will call `n= response.raw.readinto(buf)` and that will not work with the changes in this PR.  To fix that, `BodyStream` will need to implement `readinto(self, buf)`.  Could you add that?

Also regarding the tests: you can see in #1128 that I enhanced the mock socket so you can call `set_server_responses()` to set the response packet(s) from the server.  It might be a good idea to follow that approach, instead of patching `socket.socket` as done here.

(Eventually the test for `requests` should be rewritten as a unittest, but that's for a separate PR.)

pablogventura:
Thanks for the follow-up, @dpgeorge.

Added `BodyStream.readinto()` in commit baa3d77. It delegates to `_sock.readinto()` on a bounded `memoryview`, respects `Content-Length`, returns `0` when the body is exhausted, and raises the same `ValueError` as `read()` on premature EOF.

This covers the `mip` path: `_download_file()` passes `response.raw` to `_chunk()`, which loops on `readinto(buf)`.

Added `test_raw_readinto_content_length`, which mirrors that loop (small buffer, multiple reads until `n == 0`). Re-tested on unix port: `test_requests.py` 14/14 OK.

Left the `set_server_responses()` mock refactor for a follow-up, per your note on #1128.

python-ecosys/requests/requests/__init__.py
```diff
@@ -1,6 +1,36 @@
...
+        self._remaining -= len(data)
+        if not data:
+            raise ValueError("Connection closed before Content-Length satisfied")
+        return data
+
+    def readinto(self, buf):
+        if self._remaining == 0:
+            return 0
+        n = min(len(buf), self._remaining)
+        got = self._sock.readinto(memoryview(buf)[:n])
```
dpgeorge: So that this doesn't unnecessarily create memoryview objects, I suggest only slicing if needed, ie:
```py
if len(buf) > self._remaining:
    buf = memoryview(buf)[:self._remaining]
got = self._sock.readinto(buf)
```

python-ecosys/requests/README.md
```diff
@@ -14,3 +15,10 @@ redirects, basic authentication.
 * Compressed requests/responses are not currently supported.
 * File upload is not supported.
 * Chunked encoding in responses is not supported.
+* HTTP keep-alive connection reuse is not supported (Connection: close by default).
+
+### Follow-up work
+
+* Chunked response bodies (separate PR).
```
dpgeorge: Saying "separate PR" doesn't make sense in a README.

python-ecosys/requests/test_requests.py
```diff
@@ -16,6 +16,18 @@ def write(self, buf):
     def readline(self):
         return self._read_buffer.readline()
 
+    def read(self, size=-1):
+        return self._read_buffer.read(size)
+
+    def readinto(self, buf):
+        data = self._read_buffer.read(len(buf))
```
dpgeorge: Why not just `return self._read_buffer.readinto(buf)`?

python-ecosys/requests/test_requests.py
```diff
@@ -30,135 +42,236 @@ def socket(af=AF_INET, type=SOCK_STREAM, proto=IPPROTO_TCP):
         return Socket()
 
 
-sys.modules["socket"] = socket
+def install_mock_socket():
```
dpgeorge: Why is this a function?  Calling it multiple times doesn't seem like it will do anything, it only needs to be done once.

python-ecosys/requests/test_requests.py
```diff
@@ -30,135 +42,236 @@ def socket(af=AF_INET, type=SOCK_STREAM, proto=IPPROTO_TCP):
...
 
 def test_simple_get():
+    install_mock_socket()
     response = requests.request("GET", "http://example.com")
-
-    assert response.raw._write_buffer.getvalue() == (
-        b"GET / HTTP/1.0\r\n" + b"Connection: close\r\n" + b"Host: example.com\r\n\r\n"
-    ), format_message(response)
+    data = request_bytes(response)
+    assert_has(
```
dpgeorge: I understand that the headers may be printed out in a different order on different runs (due to dict ordering).  But this doesn't seem like the right way to handle it because it won't catch the case where `data` has extra unwanted lines.

Also, the first line must be "GET / HTTP ...", and that must be part of the check.

I suggest a more sophisticated that checks the first line, the headers, then the content.

python-ecosys/requests/test_requests.py
```diff
@@ -30,135 +42,236 @@ def socket(af=AF_INET, type=SOCK_STREAM, proto=IPPROTO_TCP):
...
+    test_overwrite_get_headers,
+    test_overwrite_post_json_headers,
+    test_overwrite_post_chunked_data_headers,
+    test_do_not_modify_headers_argument,
+    test_content_length_via_content,
+    test_chunked_response_raises,
+    test_raw_open_before_content,
+    test_raw_incremental_content_length,
+    test_raw_readinto_content_length,
+)
```
dpgeorge: Please just leave this as they were, being called directly.  Then a follow-up PR can rewrite it to use unittest (less churn that way).

pablogventura:
Thanks for the review, @dpgeorge.

Addressed the latest round in f23463f and cf1492d:

- `BodyStream.readinto()`: only slices into a `memoryview` when the buffer is larger than the remaining body, as suggested.
- Test mock: `Socket.readinto()` now just returns `self._read_buffer.readinto(buf)`.
- Tests: restored the original style with direct calls at the end and exact request-byte asserts. The existing tests only change `HTTP/1.0` to `HTTP/1.1`; the five new cases are appended after them.
- README: removed `(separate PR)` from the follow-up section.

Re-tested on the unix port: `test_requests.py` 14/14 OK.

dpgeorge (approved):
Thanks for updating.  This looks in good shape now.  I tested with `mip` and it seems to work well.

dpgeorge:
I squashed the commits into one.  And changed the version in the manifest to 1.0.0 because this is a pretty big change to use HTTP/1.1.

pablogventura:
Thanks a lot for the review and merge, @dpgeorge.

The squash and version bump to 1.0.0 both make sense to me. Thanks also for testing with `mip`.

PR #1125, opened 2026-06-25
unittest: optimise for code size and memory use

dpgeorge:
### Summary

Based on discussion and testing in https://github.com/micropython/micropython/pull/16311, this PR applies some optimisations to `unittest` to reduce its bytecode size, and reduce the amount of RAM it uses at runtime.

- reuse AssertRaisesContext for assertWarns
- remove support for running it with CPython
- optimise calling of _handle_test_exception
- only format `msg` if needed

The use of `raise AssertionError` instead of `assert` also means the module can now be compiled with -O3 and still function correctly (-O3 disables assert expressions!).

### Testing

Tested on ADAFRUIT_ITSYBITSY_M0_EXPRESS which has about 20k heap memory.

Prior to this change:
- `__init__.mpy` has size 5795 bytes
- `import unittest` costs 9648 bytes of heap RAM (measured via `g.collect(); micropython.mem_info()` before and after import)

With this change:
- `__init__.pym` has size 5516 bytes (-279 byte change)
- `import unittest` costs 9440 bytes (-208 byte change)

Also, the current version of `unittest` unconditionally does string formatting for the error message for each call to `self.assert<Condition>(...)`, which is really expensive in terms of RAM use and computation.  The changes here improve the situation so the string formatting of `msg` is only done if the assert fails.  That should free up a lot of RAM during the execution.

With these changes, ADAFRUIT_ITSYBITSY_M0_EXPRESS can successfully run the tests modified/added in https://github.com/micropython/micropython/pull/16311, they now all pass or skip correctly.

This module is no longer runnable under CPython due to removal of `traceback` module support.  But IMO it's too restrictive to get this working under CPython, we need to be able to optimise it for running under MicroPython.  The unittest test suite which runs under CI here anyway tests the behaviour, CPython is not needed for that.

### Generative AI

Not used.

### DIFF
--- a/python-stdlib/unittest/manifest.py
+++ b/python-stdlib/unittest/manifest.py
@@ -1,3 +1,3 @@
-metadata(version="0.10.5")
+metadata(version="0.11.0")
 
 package("unittest")
--- a/python-stdlib/unittest/unittest/__init__.py
+++ b/python-stdlib/unittest/unittest/__init__.py
@@ -1,11 +1,6 @@
 import io
 import sys
 
-try:
-    import traceback
-except ImportError:
-    traceback = None
-
 
 class SkipTest(Exception):
     pass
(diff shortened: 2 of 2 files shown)


dpgeorge:
Note that this PR is at odds with #1105 which adds a lot more features to `unittest`, making it a lot bigger.

projectgus (approved):
Changes look simple to me, that's great it reduces footprint significantly! Should churn the heap a lot less with the string formatting moved into the failure path, too.

dpgeorge:
Thanks for the quick review.

I'll let it sit a little in case @andrewleech or someone else has any comments.

andrewleech:
Thanks for the heads up, in agree these look like really good improvements! 

I haven't run the tests, but yes seeing as the module is used in CI here I'm comfortable with that covering most of not all the changes. 

Most of these changes are really clear and obvious in hindsight, and generally improve readability as well as optimising the size and runtime performance so it's a real win-win for new!

Gadgetoid:
```python
def _nop():
    pass
```

`lambda: None -> _nop()`

I believe it's used enough that the extra qstr is balanced out by not having four discrete anonymous function objects, but I'm a bit out of my depth and unsure if these optimisations generalise.

Another win is to combine the fail case into a helper:

```python
    def _fail(self, msg, fmt, *args):
        raise AssertionError(msg if msg else fmt % args)
```

Feels cleaner, too. (Edit: Okay this latter one might trade flash for RAM -200bytes, but ~+20bytes more RAM, not quite negating the lambda change.)

dpgeorge:
Thanks @Gadgetoid for the hints!

I've now made a *local* `nop` function which avoids even the global qstr for that.  That saves quite a bit not needing to create 4 separate lambads (that are all the same).

For the raise, I've changed them all to use `raise AssertionError(msg or fmt % args)` inline.  That gives the most savings.  Making that a function call only saves a byte or so, and the function costs around 36 bytes, so it's not worth it.

Altogether that saves a further 105 bytes of bytecode (size of the .mpy file).

Issue #1127, opened 2026-06-27
mip: can't resolve relative paths `../` and `requests` raises `Unsupported protocol`


o-murphy:
## Summary
mip package installer raises ValueError: Unsupported protocol when processing relative paths containing ../ segments. While the mip specification and common usage in micropython-lib imply support for relative URL resolution, the current implementation of requests (or the underlying URL parser) in the Unix port appears to reject URLs containing directory traversal segments.

## Environment
 * **MicroPython version:** v1.28.0
 * **Port:** Unix (GCC 15.2.0)
```
>>> import requests
>>> requests.__version__
'0.10.2'
>>> import mip
>>> mip.__version__
'0.4.2'
```

## Reproduction
When a package.json file uses a relative path to reference a file outside the current directory, the mip installer performs a concatenation that results in a URL containing ../.
**package.json:**
```json
{
    "version": "0.1.0",
    "urls": [
        ["file.py", "../subdir/file.py"]
    ]
}

```
**Execution:**
```python
import mip
mip.install("https://raw.githubusercontent.com/owner/repo/branch/subdir/package.json")

```
**Observed Result:**
```text
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "mip/__init__.py", line 1, in install
  ...
  File "requests/__init__.py", line 75, in request
ValueError: Unsupported protocol:

```

### Real-world example
```bash
./micropython -c "import requests; requests.get('https://raw.githubusercontent.com/ballistics-lab/micropython-bclibc/installer/natmod/../src/tiny_bclibc.py')"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "requests/__init__.py", line 205, in get
  File "requests/__init__.py", line 190, in request
  File "requests/__init__.py", line 75, in request
ValueError: Unsupported protocol: 
```

It still reachae with python3
```bash
python3 -c "import requests; p
rint(requests.get('https://raw.githubusercontent.c
om/ballistics-lab/micropython-bclibc/installer/nat
mod/../src/tiny_bclibc.py').ok)"
True
```

Urllib
```bash
python3 -c "import urllib.req
uest; print(urllib.request.urlopen('https://raw.gi
thubusercontent.com/ballistics-lab/micropython-bcl
ibc/installer/natmod/../src/tiny_bclibc.py').getco
de() == 200)"
True
```

[...]

PR #1128, opened 2026-06-29
requests: Add support for relative redirect URLs.

dpgeorge:
### Summary

A redirection URL in the "Location:" header may be relative and start with a "/".  Handle such cases.

### Testing

Added a unit test that runs under CI.

Also tested the case in #1127 and it now works.

### Generative AI

Not used.

### DIFF
--- a/python-ecosys/requests/manifest.py
+++ b/python-ecosys/requests/manifest.py
@@ -1,3 +1,3 @@
-metadata(version="1.0.0", pypi="requests")
+metadata(version="1.0.1", pypi="requests")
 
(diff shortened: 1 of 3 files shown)


o-murphy:
@dpgeorge 
Does it resolves both `/../` and `/./` subpaths?

dpgeorge:
> Does it resolves both `/../` and `/./` subpaths?

Yes.

It's not actually about the path, it's about the server responding with a redirect that normalises the path, and that redirect is relative instead of starting with `https://`.

PR #1130, opened 2026-07-01
requests: Support unicode in json and data payloads.

### LINKED ISSUE #251: urequests: miscalculation of Content-Length when dealling with unicode data

dpgeorge:
### Summary

`json.dumps()` returns a str object, and if the JSON data contains unicode then computing the length of this str -- which is used for "Content-Length" -- will return the number of characters which will be shorter than the JSON payload that is sent as bytes.  The server will then read a truncated body.

Fix that by converting any str payload into bytes.  This fix will also convert any str object passed in by the user as the `data` argument, and that actually seems to match the upstream requests behaviour (even if it's documented that `data` should be bytes).

Note that `bytes(str, "utf-8")` does not make a copy of the str data, it just makes a small bytes object that wraps the same underlying data.  So it's relatively efficient.  (Thanks to @projectgus for pointing that useful fact out!)

Fixes issue #251.

### Testing

Added tests that run under CI (also ran them locally).

Also tested sending a POST (of json unicode, and data that was str/bytes) to a local dummy server that prints out the payload, and compared with CPython requests to make sure it matched.

This shouldn't have any adverse impacts on existing users of requests:
- non-unicode json data should go through the same as before (it's now just bytes of the same data)
- unicode json data now works (length calculated correctly)
- `data` that was non-str will be unaffected
- `data` that was str will now be converted to bytes and will now have the correct length (fixing cases where non-ascii was supplied as a str to `data`)

### Generative AI

Not used in authoring the PR.

### DIFF
--- a/python-ecosys/requests/manifest.py
+++ b/python-ecosys/requests/manifest.py
@@ -1,3 +1,3 @@
-metadata(version="1.0.1", pypi="requests")
+metadata(version="1.0.2", pypi="requests")
 
 package("requests")
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -153,8 +153,11 @@ def request(
             if chunked_data:
                 if "Transfer-Encoding" not in headers and "Content-Length" not in headers:
                     headers["Transfer-Encoding"] = "chunked"
-            elif "Content-Length" not in headers:
-                headers["Content-Length"] = str(len(data))
+            else:
+                if isinstance(data, str):
+                    data = bytes(data, "utf-8")
+                if "Content-Length" not in headers:
+                    headers["Content-Length"] = str(len(data))
 
         if "Connection" not in headers:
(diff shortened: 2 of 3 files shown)


dpgeorge:
Note that without this fix, the only way to post json with unicode is to manually do it like this:
```py
requests.post(url, data=json.dumps(json_object).encode("utf-8"), headers={'Content-Type':'application/json'})
```
(as pointed out in https://github.com/micropython/micropython-lib/issues/251#issuecomment-453804501)

projectgus (approved):
Thanks @dpgeorge for picking this up, looks good to me

PR #1132, opened 2026-07-02
usb-device-keyboard: Add a simpler keyboard example, document releasing.

projectgus:
### Summary

The existing keyboard example doesn't make it obvious how to release a key, so rename that example to `keyboard_pin_example.py` and make a very simple `keyboard_hello_example.py` that presses and releases keys in sequence to type out "Hello".

Also add a note on the `send_keys()` API about how to release a key.

This is an alternative to adding a new key release API, as contributed in #892.

### Testing

* Ran the new example on an RPI_PICO2_W and an ESP32-S3 board.

* Adding a new API just for releasing keys (as implemented in #892 by @amogha1234) may be easier to use, but this way keeps the library as small as possible and it more closely maps to how USB HID works (i.e. even with a key release API, `send_keys()` will still release a key if it's not included in the argument). 

### Generative AI

### DIFF
--- a/micropython/usb/examples/device/keyboard_hello_example.py
+++ b/micropython/usb/examples/device/keyboard_hello_example.py
@@ -1,0 +1,71 @@
+# MicroPython USB Keyboard simple "Hello" example
+#
+# Simulates a keyboard pressing "H", "e", "l", "l", "o" in sequence, then exits.
+#
+# To run this example:
+#
+# 1. Make sure `usb-device-keyboard` is installed via: mpremote mip install usb-device-keyboard
+#
+# 2. Run the example via: mpremote run keyboard_hello_example.py
+#
+# 3. mpremote will exit with an error after the previous step, because when the
+#    example runs the existing USB device disconnects and then re-enumerates with
+#    the keyboard interface present. At this point, the example is running.
+#
+# 4. The example should stop running after USB has enumerated and it has typed Hello.
+#    At this point it will be possible to access the REPL again, but if you re-initialise
+#    USB then it will exit with an error again.
+#
+# For a more complex keyboard example that includes LED support, see keyboard_pin_example.py.
+#
+# MIT license; Copyright (c) 2026 Angus Gratton
+import usb.device
+from usb.device.keyboard import KeyboardInterface, KeyCode as KC
+import time
+
+
+def keyboard_example():
+    # Sequence to type.
+    #
+    # See 'class KeyCode' in usb-device-keyboard/usb/device/keyboard.py for
+    # the list of key codes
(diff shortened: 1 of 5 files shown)


micropython/usb/examples/device/keyboard_hello_example.py
```diff
@@ -0,0 +1,69 @@
+# MicroPython USB Keyboard simple "Hello" example
+#
+# Types "Hello" on the virtual keyboard once connected, then exits.
```
dpgeorge: Reading "virtual" here I initially thought it was referring to a virtual host on the PC with a simulated keyboard.  Maybe adjust the wording, eg "Simulates a keyboard pressing "H", "e", "l", "l", "o" in sequence, then exits."

micropython/usb/examples/device/keyboard_hello_example.py
```diff
@@ -0,0 +1,69 @@
...
+    # Register the keyboard interface and re-enumerate
+    k = KeyboardInterface()
+    usb.device.get().init(k, builtin_driver=True)
+
+    print("Waiting for keyboard to enumerate...")
+    while not k.is_open():
+        time.sleep_ms(100)
+
+    # Give the host a moment to be ready to receive keys
+    time.sleep_ms(250)
```
dpgeorge: Maybe increase to 1s?  Give the user a chance to see it happen after mpremote disconnects.

micropython/usb/examples/device/keyboard_hello_example.py
```diff
@@ -0,0 +1,69 @@
...
+            time.sleep_ms(10)
+
+        k.send_keys(keys)
+        time.sleep_ms(20)
+
+        # After setting a key code, need to un-set to release the key again.
+        # For this example we release all keys by sending an empty
+        # tuple of key codes
+        k.send_keys(())
+        time.sleep_ms(20)
```
dpgeorge: I feel like this last sleep should be separated because it's the delay between key strokes.

And maybe make it larger so the user can see each key coming in?  Eg:
```py
# ...
k.send_keys(())

# Delay between each key stroke
time.sleep_ms(500)
```

dpgeorge (approved):
This looks good to me, thanks!

In particular I think what you say with "it more closely maps to how USB HID works (i.e. even with a key release API, send_keys() will still release a key if it's not included in the argument)" is a good argument for doing it as per this PR.

Just a few minor (and maybe overly picky!) comments.

projectgus:
Thanks for the review, agree with your suggestions (and have updated.)

dpgeorge:
Great!

PR #1133, opened 2026-07-02
Extend Unittest with @expectedFailure

Josverl:
### Summary

I have found that in testing MicroPython it is sometimes hard to keep track of what is correct for Python, and what is correct for MicroPython.
We want the CI to remain green - to detect and avoid regressions, and MicroPython small , and close to CPython.

In MicroPython tests that means that we sometimes need to test for not CPython behaviour.
That can be done with `.exp` files for non-unittest files.

CPython's unittest has a [`@unittest.expectedFailure` decorator](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) that can be used to differentiate between: 
- unittests that pass because they are CPython compatible
- unittests that pass because they match MicroPython's minimalism 

> [!NOTE]
> This differentiation is not magically inserted , the test code still needs to be good.

I have used this approach when running matrix tests against the different typing implementations, one example where it was clear that a 100% CPython compatibility was a non-goal.
By decorating some tests with `@unittest.expectedFailure` a much cleared report can be built on the capabilities and correctness for both MicroPython and CPython.

## Implementation details

This is implemented as an optional add-on module in order to keep the `unitlest` module a lean as possible.
The base `unittest` package is intentionally small. Its `@unittest.expectedFailure` decorator swallows a failing test 
and raises `AssertionError` on an unexpected pass, but it does not track those two cases as
separate result categories the way CPython does.

This package layers that behaviour on top of the base package without modifying
it, in the same "overlay" style as `unittest-discover`: installing it adds a
single module, `unittest/expectedfailure.py`, next to the base
`unittest/__init__.py`, and importing that module patches `unittest` at runtime.

```python
import unittest
import unittest.expectedfailure  # activates the add-on (import for its side effect)

class MyTests(unittest.TestCase):
    @unittest.expectedFailure
    def test_known_difference(self):
        self.assertEqual(unsupported(), 42)  # currently fails; that's expected

if __name__ == "__main__":
    unittest.main()
```

I originally wrote this as a change to `unittest`, but with the recent optimizations refactored to an add-on module , modeled after unittest-discover. 

### Generative AI

### DIFF
--- a/python-stdlib/unittest-discover/manifest.py
+++ b/python-stdlib/unittest-discover/manifest.py
@@ -1,4 +1,4 @@
-metadata(version="0.1.3")
+metadata(version="0.1.4")
 
 require("argparse")
 require("fnmatch")
--- a/python-stdlib/unittest-discover/unittest/__main__.py
+++ b/python-stdlib/unittest-discover/unittest/__main__.py
@@ -146,8 +146,9 @@ def discover_main():
         # Ensure an appropriate output is printed if no tests are found.
         runner.run(TestSuite())
 
-    # Terminate with non zero return code in case of failures.
-    sys.exit(result.failuresNum + result.errorsNum)
+    # Terminate with non zero return code in case of failures. unexpectedSuccessesNum
+    # is only present when the unittest-expectedfailure add-on is installed.
+    sys.exit(result.failuresNum + result.errorsNum + getattr(result, "unexpectedSuccessesNum", 0))
 
 
 discover_main()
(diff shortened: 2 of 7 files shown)


dpgeorge:
> I originally wrote this as a change to `unittest`

Do you still have that changeset?  I feel like this is a relatively small addition and might be worth adding to the base `unittest` (but I'd have to see exactly what the diff is, and measure .mpy size diff).

Josverl:
> Do you still have that changeset?

Yes , but that was before your recent optimization. 
based on that I also did several additional simplifications to reduce .mpy size and memory needs
https://github.com/Josverl/micropython-lib/commit/9817ec5e885446d5daf4881a4299d598083417cb

PR #1134, opened 2026-07-03
CONTRIBUTING: Clarify some constraints for PR submission.

agatti:
### Summary

This PR updates the contribution guidelines to clarify some points related to submitting PRs in a way that wouldn't fail basic CI checks.

The contribution guidelines assumed that all PRs would come from a properly set-up environment, which it may not always be the case for small changes being done from within GitHub itself.  However these PRs would not undergo the checks provided by the pre-commit hooks that are set up in a properly set-up environment.

These changes aim to dissuade users who read the guidelines first to submit PRs from GitHub directly but to do so from a configured environment, and if that is not the case to at least make explicit the commit message validation criteria.  There isn't much to do for code formatting rules besides saying to use "ruff" first, unfortunately.

This PR was inspired by the brief discussion in #1077 - not to single the PR out - as the guidelines weren't clear on certain points and we end up with a PR that, whilst it fixed an existing issue, cannot be merged for reasons that may not be explicit enough.

### Generative AI

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,6 +37,13 @@ Similar to the main repository, a configuration is provided for the
 rules and apply linting checks. See the documentation for using pre-commit
 in [the code conventions document](https://github.com/micropython/micropython/blob/master/CODECONVENTIONS.md#automatic-pre-commit-hooks)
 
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
Thanks!

Issue #1126, opened 2026-06-25
aiohttp module case-insensitive headers


Tangerino:
Header field names are case-insensitive (RFC 9110 Sec. 5.1), so this code may break with some servers:

```python
                if line.startswith(b"Transfer-Encoding:"):
                    if b"chunked" in line:
                        chunked = True
                elif line.startswith(b"Location:"):
                    url = line.rstrip().split(None, 1)[1].decode()
```

agatti:
The relevant RFC passage: https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1 - the case-insensitivity requirement is present since RFC 2068 (§4.2 - https://www.rfc-editor.org/info/rfc2068/#section-4.2).

Technically this should also apply to `python-ecosys/requests/requests/__init__.py` as well, as there are references to capitalised headers there too.

agatti:
@Tangerino would something like this fix the issue for you? (if you have the environment set up to also check the `requests` changes it'd be great :))

```diff
diff --git i/python-ecosys/aiohttp/aiohttp/__init__.py w/python-ecosys/aiohttp/aiohttp/__init__.py
index 1e6d89d..362722a 100644
--- i/python-ecosys/aiohttp/aiohttp/__init__.py
+++ w/python-ecosys/aiohttp/aiohttp/__init__.py
@@ -125,10 +125,10 @@ class ClientSession:
                 if not line or line == b"\r\n":
                     break
                 _headers.append(line)
-                if line.startswith(b"Transfer-Encoding:"):
+                if line.lower().startswith(b"transfer-encoding:"):
                     if b"chunked" in line:
                         chunked = True
-                elif line.startswith(b"Location:"):
+                elif line.lower().startswith(b"location:"):
                     url = line.rstrip().split(None, 1)[1].decode()
 
             if 301 <= status <= 303:
diff --git i/python-ecosys/requests/requests/__init__.py w/python-ecosys/requests/requests/__init__.py
index deb262c..b2e8846 100644
--- i/python-ecosys/requests/requests/__init__.py
+++ w/python-ecosys/requests/requests/__init__.py
@@ -39,6 +39,37 @@ class Response:
         self.close()
 
 
+class _Headers:
+    def __init__(self, src = None):
+        self._d = {}
+        if src:
+            for k, v in src.items():
+                self[k] = v
+
+    def _k(self, key):
+        for current in self._d:
+            if current.lower() == key.lower():
+                return current
+        return None
+
[...]

PR #1135, opened 2026-07-04
requests: Add chunked response body support.

pablogventura:
### Summary

Follow-up to #1124, which added HTTP/1.1 with Content-Length streaming and
raised a ValueError on `Transfer-Encoding: chunked` responses. This PR adds
incremental decoding of chunked response bodies so `requests` can consume the
common case of servers that stream without a Content-Length.

A small `ChunkedStream` wrapper decodes the chunk framing on the fly. It keeps
the existing behaviour intact:

- `.raw.read(n)` and `.raw.readinto(buf)` stream the decoded body without
  buffering everything in `request()` (works for `mip`).
- `.content` stays lazy and materialises the full body on demand.
- `BodyStream` (Content-Length) is untouched; chunked responses select
  `ChunkedStream` instead.

Chunk extensions after the size are skipped and trailers are discarded.

### Testing

- Unix port: `micropython python-ecosys/requests/test_requests.py` passes,
  including a replaced test that decodes a chunked response via `.content` and
  a new test that consumes it via `.raw.readinto()`.
- Hardware: ESP32_GENERIC (MicroPython 1.29 preview) over WiFi against
  `http://httpbin.org/stream/2` (decoded via `.content`) and
  `http://httpbin.org/stream-bytes/64` (consumed via `.raw.readinto()` with a
  16-byte buffer). Both returned the expected bytes.

- Trailers are discarded; there is no API to expose them.
- Chunk extensions are ignored.
- Case-insensitive header matching is out of scope here (tracked in #1126).
- `.content` still materialises the whole body, consistent with the existing
  contract; `readinto()` is the incremental path.

### Generative AI

### DIFF
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -4,28 +4,41 @@ import socket
 class BodyStream:
     def __init__(self, sock, remaining):
         self._sock = sock
+        self._chunk = remaining < 0
         self._remaining = remaining
 
     def read(self, n=-1):
-        if self._remaining == 0:
-            return b""
-        if n < 0 or n > self._remaining:
-            n = self._remaining
-        data = self._sock.read(n)
-        self._remaining -= len(data)
-        if not data:
-            raise ValueError("Connection closed before Content-Length satisfied")
-        return data
+        buf = bytearray(n if n >= 0 else 256)
+        if n >= 0:
+            got = self.readinto(buf)
+            return buf[:got] if got else b""
+        result = b""
+        while True:
+            got = self.readinto(buf)
+            if not got:
+                return result
+            result += buf[:got]
 
     def readinto(self, buf):
-        if self._remaining == 0:
-            return 0
+        s = self._sock
+        if self._remaining <= 0:
+            if self._remaining == 0:
+                return 0
+            self._remaining = int(s.readline().split(b";")[0], 16)
+            if self._remaining == 0:
+                while True:
+                    l = s.readline()
+                    if not l or l == b"\r\n":
+                        return 0
         if len(buf) > self._remaining:
             buf = memoryview(buf)[: self._remaining]
-        got = self._sock.readinto(buf)
-        self._remaining -= got
+        got = s.readinto(buf)
         if not got:
-            raise ValueError("Connection closed before Content-Length satisfied")
+            raise ValueError("Connection closed before body complete")
+        self._remaining -= got
+        if self._remaining == 0 and self._chunk:
+            s.readline()
+            self._remaining = -1
         return got
 
     def close(self):
@@ -204,6 +217,7 @@ def request(
         if len(l) > 2:
             reason = l[2].rstrip()
         remaining = None
+        chunked = False
         while True:
             l = s.readline()
             if not l or l == b"\r\n":
@@ -211,7 +225,7 @@ def request(
             # print(l)
             if l.startswith(b"Transfer-Encoding:"):
                 if b"chunked" in l:
-                    raise ValueError("Unsupported " + str(l, "utf-8"))
+                    chunked = True
             elif l.startswith(b"Location:") and not 200 <= status <= 299:
                 if status in [301, 302, 303, 307, 308]:
                     redirect = str(l[10:-2], "utf-8")
@@ -243,7 +257,9 @@ def request(
         else:
             return request(method, redirect, data, json, headers, stream)
     else:
-        if remaining is not None:
+        if chunked:
+            resp = Response(BodyStream(s, -1))
+        elif remaining is not None:
             resp = Response(BodyStream(s, remaining))
         else:
             resp = Response(s)
--- a/python-ecosys/requests/README.md
+++ b/python-ecosys/requests/README.md
@@ -5,7 +5,8 @@ This module provides a lightweight version of the Python
 
 It includes support for all HTTP verbs, https, json decoding of responses,
 redirects, basic authentication, HTTP/1.1 requests, and reading response
-bodies with Content-Length via streaming ``.raw`` or lazy ``.content``.
+bodies with Content-Length or Transfer-Encoding: chunked via streaming
+``.raw`` or lazy ``.content``.
 
 ### Limitations
 
@@ -14,11 +15,9 @@ bodies with Content-Length via streaming ``.raw`` or lazy ``.content``.
   multipart-form encoding of post data (this can be done manually).
 * Compressed requests/responses are not currently supported.
 * File upload is not supported.
-* Chunked encoding in responses is not supported.
 * HTTP keep-alive connection reuse is not supported (Connection: close by default).
 
 ### Follow-up work
 
-* Chunked response bodies.
 * TLS certificate verification (see micropython-lib issue #838).
 * ``stream=True`` incremental body reads (see issue #777).
--- a/python-ecosys/requests/manifest.py
+++ b/python-ecosys/requests/manifest.py
@@ -1,3 +1,3 @@
-metadata(version="1.0.3", pypi="requests")
+metadata(version="1.1.0", pypi="requests")
 
 package("requests")
--- a/python-ecosys/requests/test_requests.py
+++ b/python-ecosys/requests/test_requests.py
@@ -233,19 +233,28 @@ def test_content_length_via_content():
     socket.socket = lambda *a, **k: Socket()
 
 
-def test_chunked_response_raises():
+def test_chunked_response_via_content():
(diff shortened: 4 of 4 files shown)


python-ecosys/requests/requests/__init__.py
```diff
@@ -32,6 +32,59 @@ def close(self):
         self._sock.close()
 
 
+class ChunkedStream:
```
dpgeorge: This is a lot of extra code.  Can you find a way to reduce the code size, maybe combine this class with `BodyStream`?

To test code size, run `mpy-cross __init__.py` and look at the size of the resulting `__init__.mpy` file.  Making that file smaller is the goal.

pablogventura: Thanks for the review. I merged chunked decoding into `BodyStream` with a `chunked` flag and inlined the chunk header parsing in `readinto()`, removing the separate `ChunkedStream` class (commit 4822856).

`__init__.mpy` sizes measured with `mpy-cross` against `upstream/master` (no chunked support):

| Version | `__init__.mpy` | vs master |
|---------|----------------|-----------|
| master (no chunked) | 2929 bytes | - |
| this PR, separate `ChunkedStream` | 3413 bytes | +484 (+16.5%) |
| this PR, merged into `BodyStream` | 3239 bytes | +310 (+10.6%) |

Merging the two classes saves 174 bytes (-5.1%) vs the separate-class version.

Unix tests pass. ESP32 (ESP32_GENERIC over WiFi) re-tested against httpbin chunked endpoints (`.content` on `/stream/2`, `readinto` on `/stream-bytes/64`).

Trailers are still discarded and chunk extensions are still ignored. The Content-Length code path is unchanged from master.

pablogventura: Follow-up in commit 4dafe6c: further reduced bytecode by using a `-1` sentinel for chunked mode, a single `read()` loop, and `-1` between chunks after CRLF.

`__init__.mpy` sizes measured with `mpy-cross` against `upstream/master`:

| Version | `__init__.mpy` | vs master |
|---------|----------------|-----------|
| master (no chunked) | 2922 bytes | - |
| merged `BodyStream` (4822856) | 3210 bytes | +288 (+9.9%) |
| current (4dafe6c) | 3178 bytes | +256 (+8.8%) |

Unix tests and ESP32 chunked tests against httpbin still pass.

pablogventura: One more small size tweak in 09f756e: unified the Content-Length and chunked EOF error string in `BodyStream.read()` / `readinto()`.

`__init__.mpy` (mpy-cross vs `upstream/master`):
- master: 2922 bytes
- 4dafe6c: 3177 bytes (+255)
- 09f756e: 3124 bytes (+202)

Unix and ESP32 chunked tests still pass. Chunk extensions still supported.

dpgeorge: Thanks for optimizing!  See if you can tweak it further.  Saving even 1 or 2 bytes is a win.

pablogventura: Thanks again for the review. I use MicroPython in production, so I am well acquainted with the tragedy of every spare byte - I still count them anyway, mostly for sport.

Follow-up commits since your size feedback:

| Commit | `__init__.mpy` | vs master | vs previous |
|--------|----------------|-----------|-------------|
| master (no chunked) | 2917 | - | - |
| 09f756e (unified EOF string) | 3122 | +205 | - |
| b98e90c (drop redundant `bytes()`, inline chunk size parse) | 3109 | +192 | -13 |
| latest (ruff format fix) | 3109 | +192 | 0 |

The memoryview slice spacing is ruff-required; mpy-cross measures the same 3109 bytes with or without that space, so the -13 byte win is from the other two tweaks only.

Unix tests pass. Chunk extensions still supported.

ESP32 (ESP32_GENERIC over WiFi) re-tested against httpbin:
- `.content` on `/stream/2` (372 bytes, 2 lines)
- `readinto` on `/stream-bytes/64` (64 bytes total)

All passed.

pablogventura:
Rebased onto current `master` and shrank `BodyStream` further in edbb676 by routing both Content-Length and chunked `read()` through `readinto`, plus a few micro-tweaks (`_remaining` passthrough, local sock binding, `split(b\";\")`).

`__init__.mpy` sizes with `mpy-cross` (compiled as `__init__.py`):

| Version | bytes | vs master | vs previous on this branch |
|---------|------:|----------:|---------------------------:|
| master (no chunked) | 2977 | - | - |
| this PR after rebase (pre-shrink) | 3169 | +192 | - |
| edbb676 unified `read` via `readinto` | 3094 | +117 | -75 |

Note: `read(n>=0)` may now return a `bytearray` slice (still compares equal to `bytes`; `read(-1)` / `.content` still build `bytes`). Short attribute names could save ~another 15-20 bytes but I left them out for readability.

Unix `test_requests.py` passes. ESP32_GENERIC over WiFi re-tested against httpbin: `.content` on `/stream/2` (372 bytes, 2 lines) and `readinto` on `/stream-bytes/64` (64 bytes).

pablogventura:
A few further size options on top of edbb676 (`__init__.mpy`: master 2977, current 3094 = +117):

| Option | bytes | vs master | vs current | Notes |
|--------|------:|----------:|-----------:|-------|
| A. Drop `remaining=0` default | 3089 | +112 | -5 | Safe; callers always pass an arg |
| B. Short attrs (`_s` / `_c` / `_r`) | 3077 | +100 | -17 | Smaller, less readable |
| C. B + drop default | 3072 | +95 | -22 | Same trade-off as B |
| D. Shorter EOF string | 3071 | +94 | -23 | Changes the error message |

I'd skip changing the `ValueError` text unless you want that. My lean is A for a tiny safe win, or C if you prefer maximizing the saving. Happy to apply whichever you prefer.

dpgeorge:
Thanks for continuing to opmitize.  It's going very well!

I suggest applying A (remove default).  And just shorten `_chunked` to `_chunk` (save 2 bytes, still readable).

Then that should be good enough to get this merged.  We are already down from an initial +484 to +112.

python-ecosys/requests/requests/__init__.py
```diff
@@ -2,30 +2,44 @@
...
-        self._remaining -= len(data)
-        if not data:
-            raise ValueError("Connection closed before Content-Length satisfied")
-        return data
+        buf = bytearray(n if n >= 0 else 256)
+        if n >= 0:
+            got = self.readinto(buf)
+            return buf[:got] if got else b""
+        result = b""
+        while 1:
```
dpgeorge: suggest using `while True:` for readability (should be same bytecode size)

python-ecosys/requests/requests/__init__.py
```diff
@@ -2,30 +2,44 @@
...
     def readinto(self, buf):
-        if self._remaining == 0:
-            return 0
+        s = self._sock
+        if self._remaining <= 0:
+            if self._remaining == 0:
+                return 0
+            self._remaining = int(s.readline().split(b";")[0], 16)
+            if self._remaining == 0:
+                while 1:
```
dpgeorge: suggest `while True:`

python-ecosys/requests/requests/__init__.py
```diff
@@ -2,30 +2,44 @@
...
+                    l = s.readline()
+                    if not l or l == b"\r\n":
+                        break
```
dpgeorge: can this be replaced with the final `return 0`?  might save 1 byte

pablogventura:
Thanks! Applied in fb72f58:

- A: drop `remaining=0` default
- rename `_chunked` -> `_chunk`
- `while True:` in both loops
- trailer drain: `return 0` instead of `break` + trailing `return 0`

`__init__.mpy` sizes (`mpy-cross`, compiled as `__init__.py`):

| master (no chunked) | 2977 | - |
| edbb676 (pre-tweak) | 3094 | +117 |
| fb72f58 (this commit) | 3084 | +107 |

Unix `test_requests.py` passes. ESP32_GENERIC: local chunked fixture `.content` + `readinto` OK.

pablogventura:
Follow-up ESP32_GENERIC live check over WiFi (chunked):

- `pie.dev/stream/2` via `.content` (200, 2 lines) and `pie.dev/stream-bytes/64` via `readinto` (64 bytes)
- `postman-echo.com/stream/2` via `.content` and `readinto` (388 bytes)

Also re-checked against a local chunked fixture (`.content` + `readinto`).

dpgeorge (approved):
Thanks for being diligent with this PR and following the values of MicroPython.

dpgeorge:
Squashed all the follow up / review commits and merged!

PR #504, opened 2022-07-06
WIP: Add uuid module

mattytrentini:
Basic functionality (notably, all of the `uuidX()` methods) of most of the [uuid](https://docs.python.org/3/library/uuid.html) module appears to work, or at least generate reasonable-looking results.

This started as a copy of the CPython code (see README) and then some functionality was trimmed or modified in order to execute under MicroPython.

Still to-do: 

- [ ] Update tests to pass
- [ ] Extend README.md - how to execute, which functionality is supported etc
- [ ] Remove large docstrings

PR #313, opened 2018-10-15
uuid: provide basic random uuid4() implementation

andrewleech:
This provides a random uuid4() function which returns a string, rather than a UUID class like the python builtin.

PR #1137, opened 2026-07-11
Add simple UUID module with uuid4()

dpgeorge:
### Summary

This PR pulls in @andrewleech 's minimal UUID implementation from https://github.com/pfalcon/pycopy-lib (originally submitted here as #312, #313 long ago).

Then I made a few updates to it:
- add `manifest.py`
- convert test to use `unittest` and run it under CI
- use `bytes.hex()` instead of `binascii.hexlify().decode()`
- run `ruff format`

### Testing

Test added that runs under CI.

This adds a minimal implementation of `uuid.uuid4()` which is arguably the more important of the uuidX functions.

Alternatives:
- #282 is similarly small, but didn't mask the correct bits for the UUID value
- #504 brings in the full CPython module which is large and probably we don't need all that functionality

### Generative AI

Not used.

### DIFF
--- a/python-stdlib/uuid/test_uuid.py
+++ b/python-stdlib/uuid/test_uuid.py
@@ -1,0 +1,15 @@
+import uuid
+
+u1 = uuid.uuid4()
+u2 = uuid.uuid4()
+
+assert str(u1) != str(u2), "Two uuid4 should not match"
+
+assert len(str(u1)) == len(str(u1)) == 36
+
+assert str(repr(u1)).startswith("<UUID")
+assert str(repr(u1)).endswith(">")
+
+assert len(u1.hex) == 32
+
+print("OK")
--- a/python-stdlib/uuid/test_uuid.py
+++ b/python-stdlib/uuid/test_uuid.py
@@ -15,9 +15,10 @@ class TestUUID(unittest.TestCase):
         self.assertEqual(len(str(u)), 36)
 
     def test_repr(self):
-        u = str(repr(uuid.uuid4()))
-        self.assertTrue(u.startswith("<UUID"))
-        self.assertTrue(u.endswith(">"))
+        u = repr(uuid.uuid4())
+        self.assertEqual(len(u), 44)
+        self.assertTrue(u.startswith("UUID('"))
+        self.assertTrue(u.endswith("')"))
 
     def test_constructor(self):
         u1 = uuid.uuid4()
--- a/python-stdlib/uuid/test_uuid.py
+++ b/python-stdlib/uuid/test_uuid.py
@@ -10,8 +10,9 @@ class TestUUID(unittest.TestCase):
 
     def test_len(self):
         u = uuid.uuid4()
-        self.assertEqual(len(str(u)), 36)
+        self.assertEqual(len(u.bytes), 16)
         self.assertEqual(len(u.hex), 32)
+        self.assertEqual(len(str(u)), 36)
 
     def test_repr(self):
         u = str(repr(uuid.uuid4()))
--- a/python-stdlib/uuid/test_uuid.py
+++ b/python-stdlib/uuid/test_uuid.py
@@ -1,15 +1,28 @@
+import unittest
 import uuid
 
-u1 = uuid.uuid4()
-u2 = uuid.uuid4()
 
-assert str(u1) != str(u2), "Two uuid4 should not match"
+class TestUUID(unittest.TestCase):
+    def test_unique(self):
+        n = 10
+        us = set(uuid.uuid4().bytes for _ in range(n))
+        self.assertEqual(len(us), n)
 
-assert len(str(u1)) == len(str(u1)) == 36
+    def test_len(self):
+        u = uuid.uuid4()
+        self.assertEqual(len(str(u)), 36)
+        self.assertEqual(len(u.hex), 32)
 
-assert str(repr(u1)).startswith("<UUID")
-assert str(repr(u1)).endswith(">")
+    def test_repr(self):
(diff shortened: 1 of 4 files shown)


python-stdlib/uuid/uuid.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+
+class UUID:
+    def __init__(self, bytes):
+        if len(bytes) != 16:
+            raise ValueError("bytes arg must be 16 bytes long")
```
andrewleech, line 7: I tried to test if cpython gave the save error here but it gave a much messier exception. Possibly not worth worrying about, this should be user friendly enough.

andrewleech (approved):
Looks great thanks.

python-stdlib/uuid/uuid.py
dpgeorge: Yeah, indeed CPython does give a confusing error.

dpgeorge:
Thanks for the review!

python-stdlib/uuid/test_uuid.py
```diff
@@ -0,0 +1,29 @@
+import unittest
+import uuid
+
+
+class TestUUID(unittest.TestCase):
+    def test_unique(self):
+        u1 = uuid.uuid4()
+        u2 = uuid.uuid4()
+        self.assertNotEqual(u1.hex, u2.hex)
+        self.assertNotEqual(str(u1), str(u2))
```
mattytrentini: Are there tests to ensure that, for a given port, `os.urandom` provides unique results? I could only find [tests/extmod/os_urandom.py](https://github.com/micropython/micropython/blob/master/tests/extmod/os_urandom.py). 

Perhaps we should then either test `urandom` more thoroughly or test uuid uniqueness here more thoroughly? For comparison, CPython generates 1000 UUIDs and ensures they're all unique (see [here](https://github.com/micropython/micropython-lib/pull/504/changes#diff-8bff41fb605816eece591df5aa0ab687c7bdd5d27acec32cfbdbe31b0c7dc025R626-R630)).

python-stdlib/uuid/uuid.py
```diff
@@ -0,0 +1,27 @@
...
+    @property
+    def hex(self):
+        return self._bytes.hex()
+
+    def __str__(self):
+        h = self.hex
+        return "-".join((h[0:8], h[8:12], h[12:16], h[16:20], h[20:32]))
+
+    def __repr__(self):
+        return "<UUID: %s>" % str(self)
```
mattytrentini: This repr output is different to CPython; would it be better to make it compatible?

```py
Python 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from uuid import uuid4
>>> uuid4()
UUID('84da63f4-127b-4c34-a7d9-b4321bf6f794')
```

python-stdlib/uuid/uuid.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+
+class UUID:
```
mattytrentini, line 4: I'm not sure if other methods/properties would make porting CPython code easier, it's hard to know which are in common use. `version` and `variant` seem like they would be high on the list - particularly if we decide to add support for UUIDv7 in the future - but it's probably best to hold off for now and keep it as minimal as possible.

python-stdlib/uuid/test_uuid.py
dpgeorge: Testing `urandom()` is out of scope of this PR.  A port needs to make sure that works as expected, and then this module should just be able to rely on that.

We could add a loop here to test a few more uniqueness, but IMO 1000 is excessive.

python-stdlib/uuid/uuid.py
dpgeorge: Yes, could make it the same as CPython, that would be easy.

python-stdlib/uuid/uuid.py
dpgeorge: I also don't really know what users are going to need...

dpgeorge:
@mattytrentini if you think this is not enough functionality then we can consider instead your #504 implementation.  That's fundamentally different in that it stores the UUID as an `int` and hence allows easy access of fields -- although I'm not sure how useful that will be?

mattytrentini:
> @mattytrentini if you think this is not enough functionality then we can consider instead your #504 implementation. That's fundamentally different in that it stores the UUID as an `int` and hence allows easy access of fields -- although I'm not sure how useful that will be?

No, I think this minimal approach is best, at least for now. I've tried to scour some popular applications that used uuid but of the few I looked at, they just used it without calling any other methods.

dpgeorge:
It's probably a good idea to eventually support more functions, like the namespace UUIDs.  And adding a `version` property would make sense because this `uuid4()` function creates RFC 4122 UUIDs.

python-stdlib/uuid/uuid.py
dpgeorge: Updated to match CPython.

python-stdlib/uuid/test_uuid.py
dpgeorge: Added a test that 10 UUIDs are unique.

dpgeorge:
I exposed the `.bytes` property to easily access the underlying bytes of the `UUID`.

mattytrentini:
I think this now strikes an excellent balance between functional and minimal. Looks good to me!

dpgeorge:
Thanks @mattytrentini !

PR #1138, opened 2026-07-15
usb-device-cdc: optmise CDCInterface constructor

dpgeorge:
### Summary

Don't set `CDCInterface` instance attributes that will just be reset by `self.init()`.

Saves 30 bytes of bytecode.

Also simplify the associated example.

### Testing

Built, and tested the example on RPI_PICO.  It still works.

This makes it less clear what exactly are the instance attributes, because their initialisation is done in `.init()`, not the constructor.  The added comment tries to work around that.

### Generative AI

Not used.

### DIFF
--- a/micropython/usb/examples/device/cdc_repl_example.py
+++ b/micropython/usb/examples/device/cdc_repl_example.py
@@ -25,8 +25,8 @@ import time
 import usb.device
 from usb.device.cdc import CDCInterface
 
-cdc = CDCInterface()
-cdc.init(timeout=0)  # zero timeout makes this non-blocking, suitable for os.dupterm()
+# Zero timeout makes this non-blocking, suitable for os.dupterm().
+cdc = CDCInterface(timeout=0)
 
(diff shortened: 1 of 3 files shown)


micropython/usb/examples/device/cdc_repl_example.py
```diff
@@ -25,8 +25,8 @@
 import usb.device
 from usb.device.cdc import CDCInterface
 
-cdc = CDCInterface()
-cdc.init(timeout=0)  # zero timeout makes this non-blocking, suitable for os.dupterm()
+# Zero timeout makes this non-blocking, suitable for os.dupterm().
+cdc = CDCInterface(timeout=0)
```
dpgeorge, line 29: We don't have to make this change, but I thought it was a little cleaner to set the timeout in the constructor.

dpgeorge:
@projectgus curious what you think about this change?  I can close the PR if you don't think it's a good idea.

projectgus (approved):
Looks like a nice optimisation to me!

PR #1139, opened 2026-07-15
Support @expectedFailure in unittest

Josverl:
### Summary

I have found that in testing MicroPython it is sometimes hard to keep track of what is correct for Python, and what is correct for MicroPython.
We want the CI to remain green - to detect and avoid regressiobns, and MicroPython small , and close to CPython.

In MicroPython tests that means that we sometimes need to test for not CPython behaviour.
That can be done with `.exp` files for non-unittest files.

CPython's unittest has a [`@unittest.expectedFailure` decorator](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) that can be used to differentiate between: 
- unittests that pass because they are CPython compatible
- unittests that pass because they match MicroPython's minimalism 

> [!NOTE]
> This differentiation is not magically inserted , the test code still needs to be good.

I have used this approach when running matrix tests against the different typing implementations, one example where it was clear that a 100% CPython compatibility was a non-goal.
By decorating some tests with `@unittest.expectedFailure` a much cleared report can be built on the capabilities and correctness for both MicroPython and CPython.

This PR implements this directly as part of the unittest module to keep usage as simple as possible.

```python
import unittest

class MyTests(unittest.TestCase):
    @unittest.expectedFailure
    def test_known_difference(self):
        self.assertEqual(unsupported(), 42)  # currently fails; that's expected

if __name__ == "__main__":
    unittest.main()
```
## impact on size 

Measured with mpy-cross -O3 (v1.27, mpy v6.3), using the same source filename for a fair comparison:

Build	| unittest.mpy (-O3)
-|-
master baseline|	4974 bytes
this PR|	5387 bytes
difference|	+413 bytes (+8.3%)

### Testing

Added test to test the unitest.

This is an alternative implementation to PR #1133

### Generative AI

### DIFF
--- a/python-stdlib/unittest-discover/unittest/__main__.py
+++ b/python-stdlib/unittest-discover/unittest/__main__.py
@@ -146,8 +146,8 @@ def discover_main():
         # Ensure an appropriate output is printed if no tests are found.
         runner.run(TestSuite())
 
-    # Terminate with non zero return code in case of failures.
-    sys.exit(result.failuresNum + result.errorsNum)
+    # Non-zero exit on failures; an unexpected success counts as a failure too.
+    sys.exit(result.failuresNum + result.errorsNum + result.unexpectedSuccessesNum)
 
 
 discover_main()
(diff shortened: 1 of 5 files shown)


Josverl:
Found a refactor to save 71 bytes; now at  +342 B (8.3% → 6.9%)

There are a few more options ; reduce the ClassNames names to be shorter but more cryptic would save ~24 bytes.

PR #1146, opened 2026-07-30
aiohttp: Match response header names case-insensitively.

### LINKED ISSUE #1126: aiohttp module case-insensitive headers

pablogventura:
### Summary

Fixes #1126. Response header field names are case-insensitive (RFC 9110
Sec. 5.1), but `ClientSession._request` matched `Transfer-Encoding` and
`Location` with fixed capitalization via `startswith`. Servers that send
`transfer-encoding:` or `location:` would skip chunked detection or
redirects.

Match those names with `line.lower().startswith(...)`, same idea as the
sketch on the issue and prior art in #523 (for old `urequests`, superseded
after the rename to `requests`). The `Location` value is still taken from
the original line so URL case is preserved.

### Testing

- `mpy-cross` on `aiohttp/__init__.py` (compiled as `__init__.py`):
  - master: 3787 bytes
  - this PR: 3796 bytes (+9)
- No automated aiohttp package test in CI yet; change is two comparisons in
  the header parse loop.

- aiohttp-only. The same wire-parse pattern exists in `requests` and can be
  a small follow-up; left out here to keep the diff minimal and avoid
  overlapping the open chunked-response work.
- Outgoing request header dict case-insensitivity (e.g. a `_Headers`
  helper) is a larger change and out of scope for this issue.

### Generative AI

### DIFF
--- a/python-ecosys/aiohttp/aiohttp/__init__.py
+++ b/python-ecosys/aiohttp/aiohttp/__init__.py
@@ -125,10 +125,11 @@ class ClientSession:
                 if not line or line == b"\r\n":
                     break
                 _headers.append(line)
-                if line.startswith(b"Transfer-Encoding:"):
+                lowerl = line.lower()
+                if lowerl.startswith(b"transfer-encoding:"):
                     if b"chunked" in line:
                         chunked = True
-                elif line.startswith(b"Location:"):
+                elif lowerl.startswith(b"location:"):
                     url = line.rstrip().split(None, 1)[1].decode()
 
             if 301 <= status <= 303:
--- a/python-ecosys/aiohttp/manifest.py
+++ b/python-ecosys/aiohttp/manifest.py
@@ -1,6 +1,6 @@
 metadata(
     description="HTTP client module for MicroPython asyncio module",
-    version="0.0.7",
+    version="0.0.8",
     pypi="aiohttp",
 )
 


dpgeorge:
see related #523

pablogventura:
Thanks. #523 looks like the same wire-parse issue in requests
(Transfer-Encoding / Location matched case-sensitively).

Would you rather I fold that fix into this PR, or keep a separate
small PR for requests? (If separate, I'd wait until #1135 merges so
it does not collide with the chunked work on the same header loop.)

dpgeorge:
I think we can just merge #523 (after resolving merge conflicts).  And then this PR is independent.

But, would be good to optimize this one to only call `.lower()` once, per #523.

pablogventura:
Thanks. Applied in 8678eea: one `lowerl = line.lower()` per header line, same pattern as #523.

`__init__.mpy` (`mpy-cross`, compiled as `__init__.py`):

| master | 3787 | - |
| two `.lower()` calls | 3796 | +9 |
| 8678eea (`lowerl` once) | 3803 | +16 |

I'll put the `requests` side in a separate PR (current `requests`, post-#1135) rather than rebasing #523, since that PR still targets the old `urequests` path.

dpgeorge (approved):
Thanks, looks good.

PR #1147, opened 2026-08-10
requests: Match response header names case-insensitively.

pablogventura:
## Summary
- Fixes #523 (supersedes the 2022 `urequests` PR; the module is now `requests`).
- Match `Transfer-Encoding` and `Location` on the wire case-insensitively (RFC 9110 Sec. 5.1), same `lowerl = l.lower()` once-per-line pattern as #523 and #1146.
- Covers the `requests` half of #1126. #1146 stays aiohttp-only.

## Test plan
- Unix `test_requests.py`, including lowercase `transfer-encoding` chunked and `location` redirect cases.
- `__init__.mpy` (`mpy-cross`, compiled as `__init__.py`): master 3160, this PR 3169 (+9).

### DIFF
--- a/python-ecosys/requests/requests/__init__.py
+++ b/python-ecosys/requests/requests/__init__.py
@@ -223,10 +223,11 @@ def request(
             if not l or l == b"\r\n":
                 break
             # print(l)
-            if l.startswith(b"Transfer-Encoding:"):
+            lowerl = l.lower()
+            if lowerl.startswith(b"transfer-encoding:"):
                 if b"chunked" in l:
                     chunked = True
-            elif l.startswith(b"Location:") and not 200 <= status <= 299:
+            elif lowerl.startswith(b"location:") and not 200 <= status <= 299:
                 if status in [301, 302, 303, 307, 308]:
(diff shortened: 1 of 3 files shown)


python-ecosys/requests/requests/__init__.py
dpgeorge: Could this line now be `if lowerl.startswith(b"content-length:")`?  That saves a call to `lower()` and the associated memory overhead.

pablogventura:
Thanks. Applied in e0d0462: `if lowerl.startswith(b"content-length:")` instead of `k.lower() == "content-length"`, plus a lowercase `content-length` test.

`__init__.mpy` (`mpy-cross`, compiled as `__init__.py`): master 3160, previous on this branch 3169 (+9), e0d0462 still 3169 (+9). Same bytecode size; avoids the extra `.lower()` at runtime.

Unix `test_requests.py` passes.

dpgeorge (approved):

dpgeorge:
Thanks!

pablogventura:
Thanks for merging!

PR #1148, opened 2026-08-15
aiorepl: Optimize sys.stdout.write calls.

ziesemer:
### Summary

- `aiorepl.mpy`: 3,152 -> 2,955 bytes (-197 bytes, -6.3%)
- Bump patch version to 0.2.4.

### Testing

Tested on RP2.

### Generative AI

### DIFF
--- a/micropython/aiorepl/aiorepl.py
+++ b/micropython/aiorepl/aiorepl.py
@@ -106,9 +106,10 @@ async def task(g=None, prompt="--> "):
         hist_n = 0  # Number of history entries.
         c = 0  # ord of most recent character.
         t = 0  # timestamp of most recent character.
+        write = sys.stdout.write
         while True:
             hist_b = 0  # How far back in the history are we currently.
-            sys.stdout.write(prompt)
+            write(prompt)
             cmd: str = ""
             paste = False
             curs = 0  # cursor offset from end of cmd buffer
@@ -125,7 +126,7 @@ async def task(g=None, prompt="--> "):
                         # LF or CR (handle both for raw terminal mode compatibility)
                         if paste:
                             # In paste mode, preserve the actual character
-                            sys.stdout.write(b)
+                            write(b)
                             cmd += b
                             continue
                         # Handle various newline sequences to avoid double-execution:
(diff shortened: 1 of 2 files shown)


agatti:
Thanks for the PR!

You can maybe pick a more readable name than `sw` and keep the same output size.

ziesemer:
@agatti - done.

I had tried to follow what was here:

* `s`: (StreamReader)
* `c`: "ord of most recent character"
* `pc`: "previous character"
* `t`: "timestamp of most recent character"
* `pt`: "previous time"

... with `sw` as `stdout.write`.  I now renamed this to `stdout_write`.

Regardless, confirmed - yes, the output size remains the same at 2,955 bytes.

agatti:
Thanks!

That naming scheme can work for global entities (eg. classes, functions, variables), arguments in function declarations, and class members.  Those names are retained in the MPY file once compiled, but local variable names get discarded when compiled.

ziesemer:
> That naming scheme can work for global entities (eg. classes, functions, variables), arguments in function declarations, and class members. Those names are retained in the MPY file once compiled, but local variable names get discarded when compiled.

Understood, but everything we're talking about here is all at the same scope.  :-)  So maybe those should be renamed as well...

agatti:
Point taken, then how about `write`? :)

Oh, and feel free to squash things into one single commit while you're at it.

Still, that's a nice change you've made there - I wonder where else in `micropython-lib` these modifications can be applied to.

ziesemer:
@agatti - agreed, updated, and squashed.

agatti:
Great, thank you!

agatti (approved):

micropython/aiorepl/aiorepl.py
```diff
@@ -201,55 +200,56 @@ async def task(g=None, prompt="--> "):
...
                         elif key == "[F":  # end
                             pcurs = curs
                             curs = 0
-                            sys.stdout.write("\x1b[{}C".format(pcurs))  # move cursor right
+                            write("\x1b[{}C".format(pcurs))  # move cursor right
                     else:
-                        # sys.stdout.write("\\x")
-                        # sys.stdout.write(hex(c))
+                        # sw("\\x")
+                        # sw(hex(c))
```
dpgeorge: these commented lines also need to be changed to `write`

ziesemer: @dpgeorge - Agreed, and fixed.  Sorry for missing this during the subsequent commits.

dpgeorge (approved):

micropython/aiorepl/aiorepl.py
dpgeorge: All good!

PR #1150, opened 2026-08-18
python-stdlib: Fix invalid backslash escapes in fnmatch and string.

### LINKED ISSUE #1143: Incorrect python syntax in stdlib /fnmatch
I found this while working to optimize the frozen files in the micropython stubs,
so I have not actully tried to use it , it is just that none of the Python linters or stubgens are able to condone or process this syntax.

"\Z" is not an allowed syntax, my assumption is that it should be "\\Z" 

https://github.com/micropython/micropython-lib/blob/0400c5683028417897a7e071df8350e2d4746b74/python-stdlib/fnmatch/fnmatch.py#L133-L134"

Also based on the referralsource : http://bugs.python.org/issue21464 which mentions : `'a\\ b\\Z(?ms)'`

Another similar error is in :

https://github.com/micropython/micropython-lib/blob/0400c5683028417897a7e071df8350e2d4746b74/python-stdlib/string/string/__init__.py#L9

pablogventura:
## Summary
- Fixes #1143: `"\Z"` in `fnmatch.translate` and `\]` in `string.punctuation` are unknown string escapes (linters/stubgens reject them).
- Write `"\\Z"` and `[\\]` instead. Runtime values are unchanged.

## Test plan
- `python-stdlib/fnmatch` unittest (`test_translate` included)
- `python-stdlib/string/test_translate.py`

### DIFF
--- a/python-stdlib/fnmatch/fnmatch.py
+++ b/python-stdlib/fnmatch/fnmatch.py
@@ -131,7 +131,7 @@ def translate(pat):
                 # Using ure rather than re-pcre
                 res = res + re_escape(c)
     # Original patterns is undefined, see http://bugs.python.org/issue21464
-    return "(?ms)" + res + "\Z"
+    return "(?ms)" + res + "\\Z"
 
 
 def re_escape(pattern):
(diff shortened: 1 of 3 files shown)


pablogventura:
`fnmatch.mpy` (`mpy-cross`, compiled as `fnmatch.py`):

| master | 962 | - |
| this commit | 962 | 0 |

`string/__init__.mpy` (`mpy-cross`, compiled as `__init__.py`):

| master | 506 | - |
| this commit | 506 | 0 |

Runtime strings are unchanged, so bytecode size is unchanged.

dpgeorge (approved):
Thank you.

Issue #604, opened 2023-02-01
ntptime: Does not handle KoD (Kiss of Death) resonse.


peterhinch:
A call to `.time()` can return a large negative number which results in a rather confusing exception in `.settime()`. This can occur if an application issues NTP requests too frequently: see [this post](https://github.com/micropython/micropython/discussions/10611#discussioncomment-4812100). I suspect there may also be other circumstances in which an NTP query returns zero.

Another problem reported in discussions is that `.time()` frequently times out. Clearly the timeout must be short, but this implies that on many networks timeouts will be common.

There are a number of simple fixes. One is for `.time()` to  return an error value in either of these cases which is tested by `.settime()`. The latter could return a `bool` indicating success or failure.

massimosala:
Hi Peter

I dislike the ntptime module, I think it is like "how to confuse a cat" !

Probably a lot of users get KoD because they don't realize that ntptime.time() isn't time.time() ...

I wrote my version, with more strict checking about the servers reponse.

Feel free to read and use it.
Ah, you have to remove the _compatibility_ module and the gc_collect()
The _compatibility_ is just to use this module also on windows / linux, defining _const_

Eventually I can donate it to mp ecosystem.

[ntptime2.py.txt](https://github.com/micropython/micropython-lib/files/11628081/ntptime2.py.txt)

massimosala:
I see now I use a variable MICRO in settime()

Just replace the code with:

```
	try :
		from machine import RTC
		RTC().datetime((tm[0], tm[1], tm[2], tm[6] + 1, tm[3], tm[4], tm[5], 0))
	except :
		print("Unsupported platform, no RTC")
```

and the printf() with print()

Ciao

PR #1151, opened 2026-08-18
ntptime: Reject short NTP replies.

### LINKED ISSUE #1136: ntptime spoof or bad response shape
# Current code
```python
def time():
    NTP_QUERY = bytearray(48)
    NTP_QUERY[0] = 0x1B
    addr = socket.getaddrinfo(host, 123)[0][-1]
    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    try:
        s.settimeout(timeout)
        s.sendto(NTP_QUERY, addr)
        msg = s.recv(48)
    finally:
        s.close()
    val = struct.unpack("!I", msg[40:44])[0]
```

# Proposal
```python
    msg, src = s.recvfrom(48)
    # Right host AND a full 48-byte reply. Skip spoofed/runt packets
    if src[0] == addr[0] and len(msg) >= 48:
        # got the response
    else:
        # not from the right host or packet too short
        # that may break on -> val = struct.unpack("!I", msg[40:44])[0]
```

pablogventura:
## Summary
- Fixes #1136: `time()` used `recv(48)` and unpacked `msg[40:44]` with no length check. A short datagram could raise a confusing error or yield a bad timestamp.
- Require `len(msg) >= 48`, otherwise `OSError(-1)` (same family as a socket timeout). Still uses `recv(48)`.
- Do not compare the UDP source address to the send target (can differ in complex network topologies).
- Kiss of Death / zero timestamps are unchanged (see #604).

## Test plan
- `micropython/net/ntptime/test_ntptime.py` (valid 48-byte reply, short payload), wired into package tests CI.

### DIFF
--- a/micropython/net/ntptime/manifest.py
+++ b/micropython/net/ntptime/manifest.py
@@ -1,3 +1,3 @@
-metadata(description="NTP client.", version="0.2.0")
+metadata(description="NTP client.", version="0.2.1")
 
 module("ntptime.py", opt=3)
--- a/micropython/net/ntptime/ntptime.py
+++ b/micropython/net/ntptime/ntptime.py
@@ -19,6 +19,8 @@ def time():
         msg = s.recv(48)
     finally:
         s.close()
+    if len(msg) < 48:
+        raise OSError(-1)
     val = struct.unpack("!I", msg[40:44])[0]
 
     # 2024-01-01 00:00:00 converted to an NTP timestamp
--- a/micropython/net/ntptime/test_ntptime.py
+++ b/micropython/net/ntptime/test_ntptime.py
@@ -1,0 +1,75 @@
+import struct
+import sys
+
+
+class UDPSock:
+    def __init__(self, reply):
+        self._reply = reply
+
+    def settimeout(self, timeout):
+        pass
+
+    def sendto(self, data, addr):
+        pass
+
+    def recv(self, n):
+        return self._reply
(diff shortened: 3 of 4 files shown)


pablogventura:
`ntptime.mpy` (`mpy-cross`, compiled as `ntptime.py`):

| master | 563 | - |
| this commit | 600 | +37 |

dpgeorge:
Thanks for the PR.

I'm not sure it's wise to check the address of the UDP sender via `recvfrom()`.  In complex network topologies it could be that the IP address differs between the sendto and recvfrom.

See eg this implementation which indeed does not check the address: https://github.com/vicharl/ntpclient/blob/master/ntpclient.c#L326-L333

Also might be worth seeing how this library handles that scenario: https://github.com/ntpsec/ntpsec

The other change to reject short replies seems fine.

pablogventura:
Thanks @dpgeorge - dropped the `recvfrom()` source-address check and kept rejecting short replies with `OSError(-1)`. Still uses `recv(48)`.

Addressed in `1aa3dfc`.

`ntptime.mpy` (`mpy-cross`, compiled as `ntptime.py`):

|---|---:|----:|
| master | 563 | - |
| previous on this branch | 600 | +37 |
| this commit | 584 | +21 |

dpgeorge (approved):
Thanks for updating.

PR #1152, opened 2026-08-18
ntptime: Reject Kiss-o-Death and zero timestamps.

### LINKED ISSUE #604: ntptime: Does not handle KoD (Kiss of Death) resonse.

pablogventura:
## Summary
- Fixes #604: a Kiss-o-Death reply (stratum 0) or a zero transmit timestamp was accepted. The Y2036 wrap then turned `val == 0` into a fake 2036 time, and `settime()` could apply it.
- After the length/source checks from #1151, reject `msg[1] == 0` or `val == 0` with `OSError(-1)`. Same error family as a timeout. `settime()` is unchanged.
- Stacked on #1151 (`ntptime-validate-response`). This PR also includes that commit until #1151 is merged; rebase onto `master` after that lands.

## Test plan
- `micropython/net/ntptime/test_ntptime.py`: valid reply now uses stratum 2; added KoD (stratum 0) and zero timestamp cases.

## Trade-offs and Alternatives
- Did not change `settime()` to return `bool` (would break existing callers).
- Did not change the default socket timeout.

`ntptime.mpy` (`mpy-cross`, compiled as `ntptime.py`):

| master | 563 | - |
| #1151 (previous on this branch) | 600 | +37 |
| this commit | 620 | +57 |

Delta vs #1151: +20.


pablogventura:
Follow-up: same `OSError(-1)` for stratum 0 or timestamp 0, written as `if not (msg[1] and val)`.

`ntptime.mpy` (`mpy-cross`, compiled as `ntptime.py`):

| master | 563 | - |
| previous on this branch | 620 | +57 |
| this commit | 616 | +53 |

pablogventura:
Rebased onto updated #1151 (dropped `recvfrom()` source check; short-reply validation still uses `recv(48)`). KoD / zero-timestamp rejection unchanged.

`ntptime.mpy` (`mpy-cross`, compiled as `ntptime.py`):

|---|---:|----:|
| master | 563 | - |
| previous on this branch | 616 | +53 |
| this tip (on #1151) | 600 | +37 |

dpgeorge:
Please rebase on latest master.

pablogventura:
Rebased onto latest `master` (with #1151 merged). Only the Kiss-o-Death / zero-timestamp commits remain on top.

`ntptime.mpy` (`mpy-cross`, compiled as `ntptime.py`):

|---|---:|----:|
| master | 584 | - |
| previous on this branch | 600 | +16 |
| this tip | 600 | +16 |

dpgeorge (approved):

PR #1153, opened 2026-08-31
unix-ffi/re: Fix the PCRE2 memory leaks.

klukonin:
**!!! This needs a build with `MICROPY_PY_WEAKREF` enabled !!!**

`unix-ffi/re` never frees anything it gets from PCRE2.  Every match leaks the
match data block, and every compiled pattern leaks as well, so a program or module (such as json) that
uses re module in a loop grows without bound.

Measured on a `ports/unix` build (1.30.0-preview) against libpcre2-8, resident
set size around 2x5000 calls:

| entry point | before | after |
| --- | --- | --- |
| `Pattern.search()`, match / no match | 4671 / 4383 B | 0 |
| `Pattern.match()` / `sub()` / `split()` / `findall()` | 4673 / 9025 / 9023 / 13728 B | 0 |
| `re.search()` / `re.match()` | 4848 / 4848 B | 0 |
| `re.sub()` / `re.split()` / `re.findall()` | 17824 / 17822 / 14629 B | 0 |
| `re.compile()`, same pattern / distinct patterns | 176 / 176 B | 0 |

70k operations grew the process from 4 MB to 367 MB before.
With the fixes it is flat now.

### DIFF
--- a/unix-ffi/re/re.py
+++ b/unix-ffi/re/re.py
@@ -185,10 +185,12 @@ class PCREPattern:
 
 
 def compile(pattern, flags=0):
-    errcode = bytes(4)
-    erroffset = bytes(4)
+    # These are output arguments and must be writable and of the size that
+    # pcre2_compile() writes: int for the error code, PCRE2_SIZE for the offset.
+    errcode = array.array("i", [0])
+    erroffset = array.array(PCRE2_SIZE_TYPE, [0])
     regex = pcre2_compile(pattern, PCRE2_ZERO_TERMINATED, flags, errcode, erroffset, None)
-    assert regex
+    assert regex, "error %d compiling regex at offset %d" % (errcode[0], erroffset[0])
     return PCREPattern(regex)
 
 
--- a/unix-ffi/re/re.py
+++ b/unix-ffi/re/re.py
@@ -91,12 +91,18 @@ class PCREMatch:
 class PCREPattern:
     def __init__(self, compiled_ptn):
         self.obj = compiled_ptn
+        self.key = None  # set while this pattern is held by the cache
 
     def _free(self):
         # MicroPython does not run __del__ on instances of Python classes, so
         # the compiled pattern cannot be released by the garbage collector and
         # has to be freed explicitly.
         if self.obj is not None:
+            if self.key is not None:
+                # Drop the pattern from the cache first, so that nothing hands
+                # out a pointer that is about to become invalid.
+                del _cache[self.key]
+                self.key = None
             pcre2_code_free(self.obj)
             self.obj = None
 
@@ -184,7 +190,7 @@ class PCREPattern:
             start = end
 
 
-def compile(pattern, flags=0):
+def _compile(pattern, flags):
     # These are output arguments and must be writable and of the size that
     # pcre2_compile() writes: int for the error code, PCRE2_SIZE for the offset.
     errcode = array.array("i", [0])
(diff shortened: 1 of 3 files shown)


unix-ffi/re/re.py
```diff
@@ -166,37 +190,92 @@ def findall(self, s):
...
+
+
 def search(pattern, string, flags=0):
-    r = compile(pattern, flags)
-    return r.search(string)
+    r, owned = _cached(pattern, flags)
+    try:
+        return r.search(string)
+    finally:
+        if owned:
```
agatti, line 241: If you put the `owned` check inside `_free` then it will probably be smaller once byte-compiled, so it won't have to be repeated multiple times.

unix-ffi/re/re.py
```diff
@@ -166,37 +190,92 @@ def findall(self, s):
             start = end
 
 
-def compile(pattern, flags=0):
-    errcode = bytes(4)
-    erroffset = bytes(4)
+def _compile(pattern, flags):
+    # These are output arguments and must be writable and of the size that
+    # pcre2_compile() writes: int for the error code, PCRE2_SIZE for the offset.
+    errcode = array.array("i", [0])
```
agatti, line 196: This is actually the same as doing `bytes(4)`.  Check `mp_binary_get_size` in `py/binary.c`.

TL;DR: maybe the smallest fix is to update `erroffset` size from 4 to 8 and use `int.from_bytes` to perform the bytes->int conversion.  Probably speed is the least of your worries for this function.

---

An array constructor takes 11 bytes (ignoring the extra entry in the string pool for the format string as that's already brought in elsewhere):

```
  11:02       LOAD_NAME array
  14:02       LOAD_METHOD array
  10:03       LOAD_CONST_STRING i
  80          LOAD_CONST_SMALL_INT 0 
  2b:01       BUILD_LIST 1
  36:02       CALL_METHOD 2
```

whilst creating a `bytes` object just 5:

```
  11:03       LOAD_NAME bytes
  84          LOAD_CONST_SMALL_INT 4 
  34:01       CALL_FUNCTION 1
```

`int.from_bytes` should probably suffice in this case to do the conversion, as this is only for printing (guaranteed?) positive numbers with a fixed upper size.

Shortening the message may also help here: how about "compile error %d at %d"?

unix-ffi/re/re.py
```diff
@@ -85,21 +91,39 @@ def span(self, n=0):
 class PCREPattern:
     def __init__(self, compiled_ptn):
         self.obj = compiled_ptn
+        self.key = None  # set while this pattern is held by the cache
+
+    def _free(self):
+        # MicroPython does not run __del__ on instances of Python classes, so
```
agatti, line 97: You may want to see if `weakref` can help here, added in 1.27.0.

dpgeorge: That would be great to use `weakref`... except it's not enabled on the unix port by default.

If using weakref would significantly simplify this PR then maybe we can consider enabling weakref by default on the unix port.

klukonin: Does it require too much movements or quite possible to enable for the next major unix-port version?

agatti: If I'm not mistaken it should be a matter of applying this diff, and once you've installed the packages mentioned in `docs/develop/gettingstarted.rst` [1], `make -C ports/unix` should yield an interpreter called `ports/unix/build-standard/micropython` with weakref enabled.

```diff
diff --git i/ports/unix/mpconfigport.h w/ports/unix/mpconfigport.h
index 9b66d8684..f8b4cda5b 100644
--- i/ports/unix/mpconfigport.h
+++ w/ports/unix/mpconfigport.h
@@ -125,6 +125,8 @@ typedef long mp_off_t;
 #define MICROPY_TRACKED_ALLOC       (1)
 #endif
 
+#define MICROPY_PY_WEAKREF          (1)
+
 // VFS stat functions should return time values relative to 1970/1/1
 #define MICROPY_EPOCH_IS_1970       (1)
``` 

[1] `build-essential`, `libffi-dev`, `git`, and `pkg-config` for Debian/Ubuntu.

dpgeorge: It's definitely possible to enable weakref for the next release.

klukonin: I pointed at the beginning of the PR

agatti: Well, this looks definitely neater.  Thanks for updating the code!

klukonin: @agatti Thank you very much for your suggestions =)

PR #1154, opened 2026-09-03
base64: Reduce module size and improve base32 decoding speed.

agatti:
### Summary

This PR performs a few minor updates to the `base64` module to reduce on its footprint (both on flash and in RAM) and improve Base32 decoding speed.

Base32 encoding/decoding support was updated to use a different method for symbols forward- and reverse-lookup.  Rather than using two dictionaries and a list, two `bytes` objects are used.  The forward lookup table was replaced with a single `bytes` object containing all symbols in sequence (since they were already mapped to 0..31, which incidentally is also the symbols' position in the data chunk), and the reverse lookup table was precomputed off-line to provide a 1:1 map between a byte value and its matching 0..31 index (if one is there).

That reduces the byte-compiled size by 137 bytes: one of the side-effects of that optimisation is the reduced number of objects being created to handle the Base32 encoding/decoding, reducing the overhead on the running interpreter and shrinking the objects table of the compiled version of the module.  The other being Base32 decoding should be faster since there's no need for a full dictionary lookup to figure out which 5-bits sequence a byte maps to (which was incidentally done twice in the original code).  The Base32 part of the test suite was also updated to feature the same non-ASCII buffer doing an encoding/decoding round-trip, as its Base64 counterpart.

As its CPython equivalent, `base64` is supposed to be able to run as a simple base64 encode/decode tool via the `-m base64` command line argument passed to the Python interpreter.  However, unlike the CPython implementation, this module also presents a `-t` argument that performs some basic encoding and decoding tests.

Given the existence of a test harness, the usefulness of this option is rather limited as test code is better placed in test scripts.  It is also rather unlikely that this module is executed by regular users as a standalone entity with the `-t` command line argument (this is even more relevant for embedded targets, as this functionality needs the `getopt` module being available to function).  Moving the test code from the module to the test harness and updating the help output reduced the byte-compiled size by 182 bytes.

`standard_b64{en,de}code` were defined as full-fledged function proxies for `b64{en,de}code`, now they're defined as aliases to those functions via assignment instead.  This reduced the compiled size by 31 bytes.

The combined size savings of the three proposed changes amount to 350 bytes.

### Testing

The updated `python-stdlib/base64/test_base64.py` script was run successfully on Linux/x64 using the current MicroPython `master`.  The test's output was also checked by running the test on CPython 3.14 once the existing `base64.py` file was renamed (to let CPython use its own `base64` module).

The module could be even shorter by replacing the reverse base32 lookup table with a call to `bytes.find` instead, however that's going to impact on the base32 decoding performance.  If that's not a concern then I will update the PR to do so.

The legacy interface constants could have been made `const()`, but I reckon that in theory the value for `MAXLINESIZE` could be changed globally and affect all usages of the interface.  `base64.decode` could also have been made shorter by looping through lines via `for line in input:` but then maybe this could have been used with non-`io.IOBase` derived classes (it's a legacy interface after all!).

### Generative AI


dpgeorge (approved):
Thanks, this all looks good to me.

Some pretty low hanging fruit  🎉

PR #1159, opened 2026-09-07
functools: Add an implementation for lru_cache.

agatti:
### Summary

This PR adds an implementation for both `lru_cache` and `cache` to the `functools` module.

If an unbounded cache is requested, then the function doesn't keep track of which entry was used last (so this becomes a generic `dict` wrapper).

Since dicts aren't hashable in MicroPython, keyword arguments aren't serialised internally as a dictionary to build part of the cached lookup key, but as a tuple of (key, value) tuples.  This takes up a bit more space in memory than a flattened tuple of interleaved keys and values, but that would take up more storage space when compiled.

The CI script was also modified to include the new cache-related tests (part of this commit) to the test suite.

CPython's implementation takes into account kwargs' order when computing function arguments keys, and MicroPython seems to pre-sort kwargs by key name.  This is something that may need some attention in case MicroPython's behaviour changes.

### Testing

A new test file has been added, `test_cache.py`, that should cover most use cases.  The test file on its own behaves the same also when run against CPython's `functools` module implementation (as in, copied somewhere else and executed from CPython).

The CI script has been modified to bring in `python-stdlib/functools/test_cache.py` as part of regular testing, so automated tests should pass too.

These changes increase the size of the `functools` module by 289 bytes once compiled.  The code footprint could be smaller if the memory optimisation in place for unbounded caches isn't there.

### Generative AI


Josverl:
Currently on mobile, but iirc recently CPython has added /will add `@functools.cached_method` as cached is a foot gun on methods.

https://discuss.python.org/t/adding-a-method-caching-decorator-to-functools/107164

Would be interesting to see how much additional bytes that might take.

agatti:
That's not present in 3.14 at least, I'll take a closer look, though.

Still, checking whether that was already there made me re-read the documentation and I spotted a small bug (keyword arguments order matters as far as keys are concerned) so thanks, I guess? :)

Josverl:
> That's not present in 3.14 at least

You are right,
Sorry for the distraction, glad you found something to improve still.

Issue #1162, opened 2026-09-16
usb-midi-device crashes esp32-s3 when OUT USB packets sent at line rate


projectgus:
Splitting this issue out from #1158 where it was independently discovered by @agatti while investigating a related issue.

The issue can be reproduced by setting up a CPython program that transmits MIDI data to the device as quickly as possible. For example:

```py
import rtmidi2
out = rtmidi2.MidiOut()
# Port 1 is 'Espressif Device:Espressif Device MIDI 1 24:0' in my case
out.open_port(1)
while True:
    out.send_pitchbend(0, 0)  # channel, pitch
    out.send_noteon(0, 48, 100)
    out.send_noteoff(0, 48)
    print(".", end="")
```

(Note that only the second two messages are currently processed by the MIDI device class, pitchbend messages are discarded.)

If running [midi_example.py](https://github.com/micropython/micropython-lib/blob/master/micropython/usb/examples/device/midi_example.py) on the ESP32-S3, the exhibited behaviour is a few messages may be processed normally and then the handler fails with nonsense errors such as:

```
Traceback (most recent call last):
  File "usb/device/midi.py", line 168, in _on_rx
  File "usb/device/core.py", line 859, in pend_read
AttributeError: 'Buffer' object has no attribute '_b'
Traceback (most recent call last):
  File "usb/device/core.py", line 342, in _xfer_cb
  File "usb/device/midi.py", line 156, in _rx_cb
  File "usb/device/core.py", line 827, in finish_write
AttributeError: 'Buffer' object has no attribute '_b'
```

and/or possibly a hard crash.

projectgus:
Some observations:

1. Suggest applying the fixes from #1163 before testing this.
1. This feels like memory corruption, as the missing attribute error makes little sense otherwise (and occasionally this outright crashes).
2. I originally suspected DMA alignment requirements, but confirmed the espressif__tinyusb managed component is being compiled in IRQ PIO mode so there shouldn't be any DMA involved.
3. The problem does seem to relate to partial buffers being submitted for the OUT transfer. For example, if I apply the patch below to avoid re-submitting any transfer until the buffer is entirely empty then the tests run much longer, although it still eventually crashes with a read from a corrupted address (with `EXCVADDR: 0x00308012` or similar, a register value which might be MIDI data...)

```py
diff --git i/micropython/usb/usb-device-midi/usb/device/midi.py w/micropython/usb/usb-device-midi/usb/device/midi.py
[...]

PR #1163, opened 2026-09-16
usb-midi-device: Fix handling of incoming MIDI messages under load.

### LINKED ISSUE #1158: USB-midi-device crashes when a lot of pitchbend messages are received [MicroPython v1.29.0]
As I wanted to test receiving MIDI pitchbend messages, I added the following lines to [micropython/usb/usb-device-midi/usb/device/midi.py](url):
```python
    def on_midi_event(self, cin, midi0, midi1, midi2):
        ch = midi0 & 0x0F
        if cin == _CIN_NOTE_ON:
            self.on_note_on(ch, midi1, midi2)
        elif cin == _CIN_NOTE_OFF:
            self.on_note_off(ch, midi1, midi2)
        elif cin == _CIN_CONTROL_CHANGE:
            self.on_control_change(ch, midi1, midi2)
        elif cin == _CIN_PITCH_BEND:              # <------------ Added these 4 lines
            # Pitch bend combines two 7-bit values into a 14-bit unsigned integer
            value = midi1 | (midi2 << 7)
            self.on_pitch_bend(ch, value)             
```
and
```python
    def on_pitch_bend(self, channel, value):
        pass 
```
To [micropython/usb/examples/device/midi_example.py](url) I added these lines:
```python
    def on_pitch_bend(self, ch, value):
        """Called when Pitch Bend is received. Value is an integer from 0 to 16383 (8192 is center)."""
        normalized = value - 8192
        print(f"PitchBend| Ch: {ch:<2} Raw: {value:<5} Relative to center: {normalized}")
```
Now when I move the pitchbend on my Arturia MiniLab MKII, it spits out a lot of pitchbend messages (every ~1ms). On a serial monitor, I see this:

Image

After this, the Raspberry Pi Pico 2 hangs and I have to hit Thonny's "Stop/Restart backend" button twice to return/reset the Pico to the REPL.
It seems the rapid messages overflow a queue?

Paul

projectgus:
### Summary

Fixes #1158 (RuntimeError raised in the USB transfer callback under high load of incoming MIDI messages), in three parts:

* Reduce the load on the `micropython.schedule()` queue by not scheduling redundant callbacks.
* If a call to `micropython.schedule()` does fail due to a full queue then the ongoing USB OUT endpoint transfer is still re-queued, to avoid stalling the device permanently.
* Re-submit the USB OUT transfer at the end of the scheduled callback, to handle the case where the buffer was full.

### Testing

* Ran the test case described in related issue #1162 against the MIDI example, but using an RPI_PICO2_W board.
* Confirmed messages are received correctly once these fixes are applied.

Note there is still an issue when USB-MIDI data is sent at max line rate to ESP32-S3. That issue is tracked in #1162.

### Generative AI


dpgeorge (approved):
Thanks, this looks like a good fix.

Maybe (separately) it's worth improving `micropython.schedule()` so you can ask it to only queue the given function at most once?  Or even better, preallocate the queue slot from Python so it's guaranteed to always be able to schedule.

psitech:
Tested this fix too on the Pico 2 [RP2350], and it seems fine. See [here](https://github.com/micropython/micropython-lib/issues/1158#issuecomment-5741619670).

projectgus:
Thanks @dpgeorge and @psitech!

> Maybe (separately) it's worth improving micropython.schedule() so you can ask it to only queue the given function at most once? Or even better, preallocate the queue slot from Python so it's guaranteed to always be able to schedule.

Yes, I think something like this would be useful to avoid this kind of pattern when the incoming event rate spikes up.

PR #1165, opened 2026-09-17
udnspkt: Make the module work again.

agatti:
### Summary

This PR updates the `udnspkt` module to make it work with current MicroPython versions.

The code depended on two functions called `readbin` and `writebin` that operate on byte buffers, which are not present anymore in neither micropython-lib or micropython repositories.

Those calls have been replaced with equivalent code that uses the `struct` module instead, or hardcoding a byte buffer with the static data being encoded.

The module API was tweaked slightly, in which `udnspkt.parse_resp` now raises an exception if the DNS server on the other end refuses the request.  The exception's payload is the raw response flags value from the server, in integer form.

The sample code was also updated to make it work with CPython, to make sure the behaviour is consistent across interpreters too.

And finally, the module is now ~33~ 34 bytes shorter when compiled :)

### Testing

This was tested on Linux/x64 on both MicroPython built from current git `master` and CPython 3.14.  The supplied sample code, once modified to make it work on CPython, behaves the same with both interpreters.

The DNS server was changed to 8.8.8.8 and 9.9.9.9 to solicit valid responses and rejections (9.9.9.9 doesn't like being queried for google.com..).

The main behaviour of the module wasn't changed, as in it will always return the first entry matching the required response type (IPv4 or IPv6), even if the server returned a list of possible IP addresses of a given type.

The module could be updated to return a generator instead and maybe still be smaller than the original once compiled.

### Generative AI


klukonin:
Nice improvement.
I have a little thing to add from my side.
On MicroPython 1.28 (unix port), on ath79/mips routers over cellular links, 
several asian carrier/CGNAT resolvers silently drop queries that carry an OPT RR.
Not FORMERR, not REFUSED — no response at all, so from the client's side it  
looks like packet loss or a dead uplink rather than a protocol problem, 
and  it reproduces only on the affected SIMs.
Dropping the OPT record fixed resolution on those networks.
Nothing else about the query has been changed.

Two consequences, if anything here grows an EDNS0 option later: 
- Appending it  unconditionally will look like a network fault on those mentioned carriers.
- Once  the request is back to plain 512-byte UDP, a TC→TCP fallback starts to matter again.

